PR #30, opened 2019-03-21
Improvements to README

hadley:
Based on my understanding of renv


README.md
```diff
@@ -4,26 +4,40 @@
 
 ## Overview
 
-The `renv` package is a new effort to bring private project-local R libraries
-to your projects. The goal is for `renv` to be a robust, stable replacement for
-the [Packrat](https://rstudio.github.io/packrat/) package, with fewer surprises
-and better default behaviors.
+The renv package creates **r**eproducible **env**ironments that increase the
```
kevinushey: I want to avoid over-promising on reproducibility. I'd prefer language like "`renv` helps you to produce reproducible research by <...>", rather than giving the impression that `renv` could be a panacea for all issues in reproducibility.

README.md
```diff
@@ -4,26 +4,40 @@
...
+reproducibility of your code, so it return the same results wherever and 
+whenever it is run. renv uses two tools to make this possible:
+
+* A __lockfile__, `renv.lock`, records the precise details of installed packages
+  so that anyone can reinstall exactly the same versions as you're using.
```
kevinushey: Wouldn't it be better to advertise the fact that `renv::restore()` will know how to restore these packages?

README.md
hadley: I thought this language was fairly mild - I just say "increase the chances"

README.md
hadley: My goal here was to establish the terminology to use when explaining the tools

kevinushey:
I've incorporated a modified version of this PR into the README. Thanks!

PR #83, opened 2019-06-07
Updated renv_snapshot_validate_dependencies() to ensure that a logical result is returned

khughitt:
Currently, `renv_snapshot_validate_dependencies()` can return either a logical, or a character array, if some problem is encountered.

The later results in a problem with [renv_snapshot_validate()](https://github.com/rstudio/renv/blob/master/R/snapshot.R#L133), however, when it attempts to evaluate the expression:

```r
  renv_snapshot_validate_dependencies(project, lockfile, library, confirm) &&
    renv_snapshot_validate_sources(project, lockfile, library, confirm)
```

With this change, any 'bad' packages detected will result in `renv_snapshot_validate_dependencies()` returning `FALSE`.

### DIFF
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -219,7 +219,7 @@ renv_snapshot_validate_dependencies <- function(project, lockfile, library, conf
   request  <- bad$Requested
 
   fmt <- "- %s requires %s, but %s will be snapshotted"
-  warning(sprintf(fmt, format(package), format(requires), format(request)))
+  warningf(fmt, format(package), format(requires), format(request))
 
   return(FALSE)
 }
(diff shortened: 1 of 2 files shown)


R/snapshot.R
```diff
@@ -219,8 +219,9 @@ renv_snapshot_validate_dependencies <- function(project, lockfile, library, conf
   request  <- bad$Requested
 
   fmt <- "- %s requires %s, but %s will be snapshotted"
-  sprintf(fmt, format(package), format(requires), format(request))
+  warning(sprintf(fmt, format(package), format(requires), format(request)))
```
kevinushey: You can also use `warningf()` (it's a shortcut for exactly this sort of call)

PR #98, opened 2019-07-05
Add pointers to environments

slopp:


slopp:
@kevinushey I assume pkgdown is built automatically? Or should I also commit the rendered html?

kevinushey:
Indeed, we use Travis to build the pkgdown website now: https://github.com/rstudio/renv/blob/b1a11f2b74f4ccf3889c48856f392d85358636ec/.travis.yml#L16-L19

PR looks good -- thanks!

Issue #100, opened 2019-07-09
Allow PAT environment variable to be specified for git2r repos


mstr3336:
Hi, 
The project I develop for is hosted on a GitHub Enterprise server, and unfortunately, as a result, installing packages from git requires git2r, and for the user to supply credentials.
For a feature enhancement, I was wondering if in the lockfile, for git2r repos that may not be public-read, if a field that specifies the env variable holding a user's Personal Access Token might be supplied?
Really, that probably just looks like a few lines of code and maybe an extra parameter to a git2r call, and if that sounds like something you'd be happy with integrating, I can implement it (I'd find it really handy since I presently ignore the enterprise packages and then have my own hacky script that's responsible for maintaining them).
Let me know what you think :)

mstr3336:
I implemented this in #101

kevinushey:
It looks like there are a few ways to accomplish this:

1. Specify e.g. a `GIT_USER` and `GIT_PASSWORD`, and use those to set up a repository URL (e.g. `https://<user>@github.com/repo/user`), and then set `GIT_ASKPASS` to just echo `GIT_PASSWORD` for authentication.

2. For certain hosts (e.g. GitHub) it looks like we can just use a PAT directly, e.g. `https://<token>:x-oauth-basic@github.com/repo/user`.

We'll have to be careful that whatever scheme is chosen we don't accidentally leak the token (e.g. it looks like cloning with a URL like that can save the token in `.git/config`, which is suboptimal)

mstr3336:
as per [my comment](https://github.com/rstudio/renv/pull/101#issuecomment-509844830)

> I like the sound of the
> ```r 
> options(renv.auth.git.MyPackage = function(record){})` 
> ```

Which could be called in .Rprofile, or in an auth script that .Rprofile could source  before sourcing `renv/activate.R`

> We'll have to be careful that whatever scheme is chosen we don't accidentally leak the token (e.g. it looks like cloning with a URL like that can save the token in `.git/config`, which is suboptimal)

I did want to check in re. this: 
When renv installs from git, how much cleanup does it do? Is the repository it clones deleted afterwards? 
If so, as long as the token is saved to the repos local config, how important is it that the token isn't saved before being deleted?

kevinushey:
> When renv installs from git, how much cleanup does it do? Is the repository it clones deleted afterwards?

It is indeed deleted afterwards. We actually just fetch the branch of interest and build from that:

[...]

PR #101, opened 2019-07-09
Allow users to set RemoteTokenEnvVar to an environment variable to manage private repos in lockfile

mstr3336:


mstr3336:
As per my issue in #100

kevinushey:
Thanks for the PR! Some thoughts:

Is there any prior precedent for naming this "RemoteTokenEnvVar"? Does that come from e.g. `git2r` or another package?

Currently, we use `GITHUB_PAT` for restores from GitHub as opposed to Git:

https://github.com/rstudio/renv/blob/21e149c20a91cb679cf2bf06e4cf8390e69ce401/R/download.R#L346-L354

Is there a way we could detect that the Git repository being used is associated with a GitHub host, and use that? Or even just something called `GIT_USERNAME` and `GIT_PASSWORD`? We do something like this for Bitbucket; e.g.

https://github.com/rstudio/renv/blob/21e149c20a91cb679cf2bf06e4cf8390e69ce401/R/download.R#L334-L344

It seems like if you used `remotes::install_github()` with `GITHUB_PAT` set appropriately then the existing machinery in `renv` would work out of the box.

mstr3336:
> Is there any prior precedent for naming this "RemoteTokenEnvVar"? Does that come from e.g. git2r or another package? 

Absolutely not, mostly chosen arbitrary

> Currently, we use GITHUB_PAT for restores from GitHub as opposed to Git:

My main problem with the hardcoded use of GITHUB_PAT is that, for my workplace's enterprise server, the system, and accounts are totally different from the public GitHub, and (I should re-investigate ) iirc, setting `GITHUB_PAT` to work on the enterprise domain seems to break things on the public github domain.

> It seems like if you used remotes::install_github() with GITHUB_PAT set appropriately then the existing machinery in renv would work out of the box.

For whatever reason, I have some problems getting things to work on the normal API, 

I seem to get the most consistent behaviour when installing from my work's remotes by using git2r, giving it a direct https git url `https://github.matts.work.edu.au/matts-team/repo.git` , and supplying a `git2r::credentials` object. 

I think allowing a general purpose auth-from-arbitrary-env-var use case for git repos not found on any specific host might be a flexible and useful feature for allowing users to maintain private repos in their lock files, given that not all repos across different/the same domains will necessarily have the same authorisations.

maybe I was a little shortsighted in implementing my use case, given that my use case really just takes a url resembling `https://github.matts.work.edu.au/matts-team/repo.git` and transforms it to `https://arbitrary_username:<password>@github.matts.work.edu.au/matts-team/repo.git` where `<password>` is just whatever is found @ `RemoteTokenEnvVar`.

It might make more sense to generalise the use case to instead allow users to specify something like "RemotePassword" and "RemoteUsername" in the lockfile record, which is easily extensible to my usecase.

Sorry if this was a bit wordy, I've had a long day, but I hope I've made myself clear. 

*tl;dr*

Does a general case of allowing users to specify per-record env vars for usernames/passwords for arbitrary git urls sound useful? If so I can implement that

kevinushey:
> Does a general case of allowing users to specify per-record env vars for usernames/passwords for arbitrary git urls sound useful? If so I can implement that

I agree 100% that something like this would be useful. First though, I'd like to think about how to make this consistent so that it could be useful across different remote sources + authentication methods.

Some questions worth answering in regards to the right implementation:

Should this be done with R options, or with environment variables? We have a bit more flexibility when using R options, but environment variables can be easier to use in other contexts.

For R options: suppose we have a package called `MyPackage`. Users could specify custom authentication with something like:

```
options(renv.auth.git.MyPackage = list(username = "<user>", password = "<password>")
```

Or, we could allow it to be a function; e.g.

```
options(renv.auth.git.MyPackage = function(record) {
  list(username = "<user>", password = "<password>")
})
```

Or, we could just have a single user-provided function for authentication in general (which would then do the right thing based on the package / record being queried)

```
options(renv.auth = function(package, record) {
    if (package == "MyPackage") { ... }
})
```

The advantage of a function here being that authentication could come from any source; e.g. environment variables, or some other encrypted storage mechanism on the system. Users could then set these options in an appropriate R startup file; e.g. in the project `.Rprofile` or something else. The function could even use different authentication methods depending on other fields in the record.

If we wanted to use environment variables, we could do something similar but it could get a bit more unwieldy:

```
RENV_AUTH_GIT_MYPACKAGE_USERNAME = <...>
RENV_AUTH_GIT_MYPACKAGE_PASSWORD = <...>
```

Overall, I think I would prefer to use R options. Let me know if you agree this would work in your use case as well.

kevinushey:
The other thing you could consider as a shortcut altogether is whether you could set up an SSH agent and copy your SSH key to GitHub. Since `renv` just invokes `git` from the command line, other command line tools should be able to help here as well.

mstr3336:
R options sound like a good mechanism, though I don't fully understand their mechanics, as I hardly touch them when developing packages (I'm aware that it's probably very simple and I can read an article) 

The option to allow users to define handlers for credential access sounds flexible and useful, & leaves the option of developers to also provide a default handler that does something easy, like access env vars or whatever makes sense to you.

I like the sound of the 
```r
options(renv.auth.git.MyPackage = function(record){})`
```

interface. 

What would the return of that interface? 
Perhaps [git2r::credential objects](https://docs.ropensci.org/git2r/reference/cred_env.html) or something resembling that?

Also, would that imply that on the lockfile record level, no changes would need to be reflected, but instead the behaviour would be handled purely in options/ .Rprofile?

Also, I'm aware that renv already relies on a simple `.Rprofile` for its setup infra, that just runs the activate script- would we just ask users to change 

```r
source("renv/activate.R")
```
to 
```r
source("my_setupt_script.R")
source("renv/activate.R")
```
or something similar?

> The other thing you could consider as a shortcut altogether is whether you could set up an SSH agent and copy your SSH key to GitHub.

For my use-case that's probably something worth investigating, though my goal is for other members of my team (Some of whom are ___academics___ to find the tools accessible & easy to pick up)

kevinushey:
I've implemented `renv.auth` here: https://rstudio.github.io/renv/articles/renv.html#authentication

However, this still doesn't answer the question as to how we should manage this for remotes using `git` directly. Let's continue the discussion in #100.

PR #119, opened 2019-07-29
return unique dependencies

schloerke:


kevinushey:
Sorry, I should have brought this up earlier -- it's actually intentional that we don't return unique dependencies here. In the future, this could be used to enumerate dependencies alongside the line numbers where dependencies are referenced; I didn't want to drop that information as well.

I think client code of `renv` can drop duplicates if so required anyhow?

schloerke:
Sounds fair. Feel free to close.

Thank you!

PR #163, opened 2019-08-23
Add support for falling back to the default user cache if the envars-…

m-l-1:
…defined cache is not writable by the user. 

In our organization we have set a central and global cache for renv which is only writable by admins. The cache is defined to be the default for all users using the variable _RENV_PATHS_CACHE_.

These changes allow the developers to use their local cache when a package version is not already installed in the global cache.

I haven't extensively tested it but it works with the main functions, install/purge/restore/...

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -126,8 +126,12 @@ renv_install_impl <- function(record, project) {
   # figure out whether we can use the cache during install
   library <- renv_global_get("install.library") %||% renv_libpaths_default()
   linkable <-
-    settings$use.cache(project = project) &&
-    identical(library, renv_paths_library(project = project))
+    settings$use.cache(project = project) && (
+      identical(library, renv_paths_library(project = project)) ||
+        # Support for packaged renv under inst
+        # Remove inst sub-folder from path
+      identical(sub("inst/","",library), renv_paths_library(project = project))
+    )
 
   linker <- if (linkable) renv_file_link else renv_file_copy
 
@@ -140,6 +144,11 @@ renv_install_impl <- function(record, project) {
 
     # check for cache entry and install if there
     cache <- renv_cache_package_path(record)
+    if (file.exists(cache))
+      return(renv_install_package_cache(record, cache, linker))
+
+    # check for cache entry in the default location and install if there
+    cache <- renv_cache_package_path(record, writable = TRUE)
     if (file.exists(cache))
       return(renv_install_package_cache(record, cache, linker))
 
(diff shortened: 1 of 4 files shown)


kevinushey:
Thanks for the pull request!

We require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

R/install.R
```diff
@@ -126,8 +126,12 @@ renv_install_impl <- function(record, project) {
   # figure out whether we can use the cache during install
   library <- renv_global_get("install.library") %||% renv_libpaths_default()
   linkable <-
-    settings$use.cache(project = project) &&
-    identical(library, renv_paths_library(project = project))
+    settings$use.cache(project = project) && (
+      identical(library, renv_paths_library(project = project)) ||
+        # Support for packaged renv under inst
+        # Remove inst sub-folder from path
+      identical(sub("inst/","",library), renv_paths_library(project = project))
```
kevinushey: Can you explain the rationale here?

kevinushey:
Thanks for the PR! I'll need to think about whether this is the approach I want to take here -- the idea of having potentially multiple active caches worries me a bit.

R/paths.R
```diff
@@ -14,8 +14,16 @@ renv_paths_common <- function(name, prefixes = NULL, ...) {
...
   # compute root path
   envvar <- paste("RENV_PATHS", toupper(name), sep = "_")
-  root <- Sys.getenv(envvar, unset = renv_paths_root(name))
-
+  # test if configured path is writable
+  root <- Sys.getenv(envvar)
+  if(root == ""){
+    root <- renv_paths_root(name)
+  }else{
+    if(writable && !assertthat::is.writeable(root)){
```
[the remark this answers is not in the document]
m-l-1: Done!

R/install.R
m-l-1: This change belongs to another branch, I have moved it.
see this PR https://github.com/rstudio/renv/pull/166

kevinushey:
I believe this is better handled now in `renv` -- if the cache is not writable by the user for some reason, then `renv` will not attempt to write to it. Users can install packages locally for their projects; they just won't get copied into the global cache.

Having separate global / local caches is beyond what I want to consider at this point.

PR #164, opened 2019-08-23
fix for cross device link error on linux os

m-l-1:
A quick fix using system command _mv_. Works fine!

### DIFF
--- a/R/files.R
+++ b/R/files.R
@@ -99,6 +99,14 @@ renv_file_move <- function(source, target, overwrite = FALSE) {
   if (renv_file_exists(target))
     return(TRUE)
 
+  # On Linux try another option
+  # Use the system command to move the links
+  if(tolower(Sys.info()[["sysname"]]) == "linux"){
+    sysmove <- catchall(system(paste("mv",source,target)))
+    if(renv_file_exists(target))
+      return(TRUE)
+  }
+
   # nocov start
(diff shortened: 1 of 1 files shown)


R/files.R
```diff
@@ -99,6 +99,14 @@ renv_file_move <- function(source, target, overwrite = FALSE) {
   if (renv_file_exists(target))
     return(TRUE)
 
+  # On Linux try another option
+  # Use the system command to move the links
+  if(tolower(Sys.info()[["sysname"]]) == "linux"){
```
kevinushey: I wonder if this should apply to all Unix-alikes? (could use `renv_platform_unix()` to check that)

kevinushey: Also: please use whitespace around `if`; and for e.g. `) {`

R/files.R
```diff
@@ -99,6 +99,14 @@ renv_file_move <- function(source, target, overwrite = FALSE) {
...
+    sysmove <- catchall(system(paste("mv",source,target)))
```
kevinushey: Need to use `shQuote()` on `source` and `target`

kevinushey:
Made a similar version of the changes here: https://github.com/rstudio/renv/commit/3c4103f75bcf9a2d1790497085774431938a499f

Thanks!

PR #165, opened 2019-08-23
In some situations the .getNamespaceInfo() function is returning a na…

m-l-1:
…med character vector and not a list as expected in the current code


kevinushey:
Thanks for the PR -- I'm going to take an alternate but similar approach here.

PR #181, opened 2019-09-07
Fix typo

uribo:
If do not close the parentheses, we will get the following error.

```
/bin/sh: 1: Syntax error: Unterminated quoted string
ERROR: Service 'rstudio' failed to build: The command '/bin/sh -c R -e 'install.packages("remotes", repos = c(CRAN = "https://cloud.r-project.org"))' returned a non-zero code: 2
```

Avoid errors by closing parentheses.


kevinushey:
Thanks for catching that!

Issue #191, opened 2019-09-20
Error on running `renv::snapshot()`


jimhester:
I setup renv using `renv::init()` and haven't done much else, but when I tried running `renv::snapshot()` I ran into an error.

```r
> renv::status()
The following package(s) are installed but not recorded in the lockfile:
       _
  renv   [0.7.0-88]

Use `renv::snapshot()` to add these packages to your lockfile.

The following package(s) are out of sync:

   Package   Lockfile Version   Library Version
      glue         1.3.1.9000             1.3.1

Use `renv::snapshot()` to save the state of your library to the lockfile.
Use `renv::restore()` to restore your library from the state of the lockfile.

*> renv::snapshot()
The following package(s) will be updated in the lockfile:

# CRAN ===============================
- glue   [1.3.1.9000 -> 1.3.1]

Error in vapply(x, f, ..., FUN.VALUE = character(1)) :
  values must be length 1,
 but FUN(X[[1]]) result is length 0
Traceback (most recent calls first):
8: renv::snapshot()
7: renv_snapshot_report_actions(actions, old, new)
6: renv_pretty_print_records_pair(lhs[names(lhs) %in% names(actions)],
       rhs[names(rhs) %in% names(actions)], "The following package(s) will be updated in the lockfile:")
5: renv_pretty_print_records_impl(old, new)
4: lapply(sort(unique(groups)), function(group) ...)
3: FUN(X[[i]], ...)
2: map_chr(nms, function(nm) {
       renv_record_format_pair(lhs[[nm]], rhs[[nm]])
   })
1: vapply(x, f, ..., FUN.VALUE = character(1))
> traceback()
8: vapply(x, f, ..., FUN.VALUE = character(1))
7: map_chr(nms, function(nm) {
       renv_record_format_pair(lhs[[nm]], rhs[[nm]])
   })
6: FUN(X[[i]], ...)
5: lapply(sort(unique(groups)), function(group) {
       lhs <- renv_records_select(old, groups, group)
       rhs <- renv_records_select(new, groups, group)
       nms <- union(names(lhs), names(rhs))
       text <- map_chr(nms, function(nm) {
           renv_record_format_pair(lhs[[nm]], rhs[[nm]])
       })
       if (group == "unknown")
           group <- "(Unknown Source)"
       vwritef(header(group))
       vwritef(paste("-", format(nms, width = n), " ", text))
       vwritef("")
   })
4: renv_pretty_print_records_impl(old, new)
[...]

PR #192, opened 2019-09-20
Collapse records before retrieving their names

jimhester:
This avoids an error for NULL results

Fixes #https://github.com/rstudio/renv/issues/191


kevinushey:
Thanks; I'll take this as a workaround for now and see if I can figure out what the underlying error is.

PR #216, opened 2019-10-17
Feature/discover import

swt30:
Been using renv and really liking it so far. I noticed that it doesn't have support for the `import` package (https://github.com/smbache/import) when discovering dependencies. I've put together a quick PR to enable that. Let me know if this is useful and I can write some tests too.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -817,7 +817,7 @@ renv_dependencies_discover_r_import <- function(node, envir) {
     return(FALSE)
 
   # match a call to any function with a '.from' argument
-  prototype <- function(.from = "", ...) {}
+  prototype <- function(.from, ...) {}
   matched <- catch(match.call(prototype, node, expand.dots = FALSE))
   if (inherits(matched, "error"))
     return(FALSE)
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -606,7 +606,8 @@ renv_dependencies_discover_r <- function(path = NULL, text = NULL) {
     renv_dependencies_discover_r_require_namespace,
     renv_dependencies_discover_r_colon,
     renv_dependencies_discover_r_pacman,
-    renv_dependencies_discover_r_modules
+    renv_dependencies_discover_r_modules,
+    renv_dependencies_discover_r_import
   )
 
   discoveries <- new.env(parent = emptyenv())
@@ -807,6 +808,30 @@ renv_dependencies_discover_r_modules <- function(node, envir) {
   envir[[as.character(package)]] <- TRUE
 
   TRUE
+}
+
+renv_dependencies_discover_r_import <- function(node, envir) {
+
+  node <- renv_call_expect(node, "import", c("from", "here", "into"))
+  if (is.null(node))
+    return(FALSE)
+
+  # match a call to any function with a '.from' argument
+  prototype <- function(.from = "", ...) {}
+  matched <- catch(match.call(prototype, node, expand.dots = FALSE))
+  if (inherits(matched, "error"))
+    return(FALSE)
+
+  # the '.from' argument is the package name, either a character vector of length one or a symbol
+  if ((is.character(matched$.from) && length(matched$.from) == 1)
+      || is.symbol(matched$.from))
+  {
+    envir[[as.character(matched$.from)]] <- TRUE
+    return(TRUE)
+  }
+
+  FALSE
+
 }
 
 renv_dependencies_list <- function(source,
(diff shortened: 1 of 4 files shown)


R/dependencies.R
```diff
@@ -716,6 +717,30 @@ renv_dependencies_discover_r_pacman <- function(node, envir) {
...
 }
 
+renv_dependencies_discover_r_import <- function(node, envir) {
+
+  node <- renv_call_expect(node, "import", c("from", "here", "into"))
+  if (is.null(node))
+    return(FALSE)
+
+  # match a call to any function with a '.from' argument
+  prototype <- function(.from, ...) {}
```
kevinushey: I think you'll need to use a different prototype depending on which function is being used. E.g.

```
> args(import::from)
function (.from, ..., .into = "imports", .library = .libPaths()[1L],  .character_only = FALSE) 
> args(import::into)
function (.into, ..., .from, .library = .libPaths()[1L], .character_only = FALSE) 
> args(import::here)
function (..., .from, .library = .libPaths()[1L], .character_only = FALSE) 
```

In other words, unless we're using `import::from`, I think the prototype has to be:

```
prototype <- function(..., .from)
```

kevinushey:
Thanks for the pull request! We require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept non-trivial pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

R/dependencies.R
swt30: I don't seem to be able to construct a valid call to import::from or import::here that would fail to match the given prototype. For example, any of the following match:
```
proto = function(.from, ...) {}

match.call(proto, expression(import::from(package, func)))
match.call(proto, expression(import::from(func, .from = package)))

match.call(proto, expression(import::here(func1, func2, .from = package)))
match.call(proto, expression(import::into(func1, func2, .into = "imports:pkg", .from = package)))
match.call(proto, expression(import::into("imports:pkg", func1, func2, .from = package)))
```
while constructions like the following that would otherwise fail to match aren't valid import calls (``Argument `.from` must be specified``)
```
import::here(func, package)
import::here(func1, func2, package)
import::into(func1, func2, package, .into = "imports:pkg")
```
So I think the prototype should cover all uses of this import mechanism, unless I'm missing something.

kevinushey: My main worry is that this code would cause `renv` to falsely infer a dependency:

```
> proto <- function(.from, ...) {}
> match.call(proto, quote(import::here("object")))
import::here(.from = "object")
```

This is of course an invalid use of `import::here()` by the user, but I think it's worth guarding against.

kevinushey: In other words: I think it's worth being resilient against certain inappropriate usages of these functions.

swt30:
I'm getting the form sorted out – once that's done I'll return to this PR and make your suggested change to deal more thoroughly with different forms of the import:: call 🙂

kevinushey:
@swt30 looks like we've received the CLA! Let me know if you have time to complete this PR or if you'd like to me try and take a look separately.

swt30:
Great, I'll take a look at this again then! We are starting to investigate the use of renv more widely at my workplace, and we use `import::`, so I think this will be a helpful addition.

swt30:
I came back to this and tried to continue working on it, but my progress has been stymied by an error that is preventing me from running renv at all on the latest master. I think it is a problem with the use of the utils::tar function `tar(..., tar = "internal")` which appears to be [incompatible with uids > 32767](https://stackoverflow.com/questions/30599326/warning-message-during-building-an-r-package-invalid-uid-value-replaced-by-that). My uid is really large, which sounds like a valid behaviour for a newer system but isn't handled by `tar`. Because renv sets options(warn=2) at some point, this produces a warning that is converted to an error. 

I've been unable to find a workaround; the workaround in that link is only for building/installing packages, but renv calls the tar function directly. I'll file an issue about this – maybe warn=2 isn't the best behaviour here, or maybe renv could use a system tar if possible instead of forcing the use of the internal function?

kevinushey:
FWIW we use R's internal tar explicitly because we've seen a variety of issues where attempts to invoke `tar` fail because the user has placed an esoteric / broken `tar` on the `PATH`. We could probably find some happy middle-ground but that's the intention at least.

kevinushey:
That said this isn't really a concern when running tests so I've dropped the explicit use of R's internal tar.

swt30:
Added tests, and used a different function signature for import::into and import::here to avoid matching incorrect usage like `import::here("object")`

kevinushey:
Thanks!

Issue #235, opened 2019-11-07
Module Dependencies


labriola:
This is a future-looking feature request. At the moment, dependencies expressed in patterns such as modules (https://cran.r-project.org/web/packages/modules/vignettes/modulesInR.html) are incorrectly identified. The consequences is that, unless one set snapshot type to simple, dependencies only used in these types of modules do not end up in the lock file (I am assuming during the dependency discovery phase it believes it's not needed).

Realize this is complex and renv::settings$snapshot.type("simple") is an easy workaround. Thanks

kevinushey:
Is there a common pattern in the way modules are used that `renv` could crawl for dependencies?

kevinushey:
IIUC we could try to crawl for `import()` calls within `module({})` definitions.

labriola:
That's what I was thinking. Let me see if I can prototype it up in a PR. I don't know your code base particularly well yet so no concerns if the effort is for information only

kevinushey:
For reference, we "register" a number of different methods for collecting dependencies from R scripts here:

https://github.com/rstudio/renv/blob/b434756f96cb1d0a701318a6d069955f532c93ef/R/dependencies.R#L537-L544

And this is a good example of how we parse packages from `xfun::pkg_attach()`:

https://github.com/rstudio/renv/blob/b434756f96cb1d0a701318a6d069955f532c93ef/R/dependencies.R#L572-L600

So a PR could add a similar method in this file.

labriola:
Perfect. I appreciate the info, let's see what I can get working

kevinushey:
Also worth saying up front -- if you put together a pull request, we do require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

labriola:
Understood and willing

kevinushey:
Implemented via #238. Thanks!

PR #237, opened 2019-11-08
Module matching per issue #235

labriola:
Let me know your thoughts on this approach. I didn't want to muck about too much in the code for a first PR but I also didn't want to copy the xfun implementation as its identical to the requirement for the module needs.

I am happy to revise so consider this preliminary

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: renv
 Type: Package
 Title: Project Environments
-Version: 0.8.2-23
+Version: 0.8.2-24
 Authors@R: c(
     person("Kevin", "Ushey", role = c("aut", "cre"), email = "kevin@rstudio.com"),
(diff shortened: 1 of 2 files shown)


labriola:
FYI, contributor agreement already submitted

labriola:
(Part of the failure seems related to: test-retrieve.R:77: error: we can retrieve packages from git
cannot retrieve package 'skeleton' from 'git://github.com/kevinushey/skeleton.git'

labriola:
Okay, I will take another stab and abandon this approach. I must be misinterpreting the action of the code regarding the ... and will look further. The general pattern of using import("foo") inside of a module is correctly captured correctly by this

kevinushey:
My guess is that it might fail if someone used e.g. `modules::import("rnorm", from = "stats")`?

labriola:
More than fair. I appreciate your time. I am on it

PR #238, opened 2019-11-09
Adding support for imports via modules::import()

labriola:
@kevinushey give me your thoughts on this. I tested against a real project and the combinations I could think of and all appears valid... however... I don't know what I don't know so feedback is very much appreciated. 

This is rebased on your latest commits. Addresses #235


kevinushey:
Looks perfect! Once I get confirmation that we've received the CLA, I'll merge this. Thanks for the PR!

labriola:
Let me know if you need anything else from me. Thanks!

kevinushey:
I've received confirmation that we've received the CLA. Thanks!

PR #298, opened 2019-12-11
correct full path example in documentation for what is appended to path

cderv:
Order is now renv version, R version and platform information
https://github.com/rstudio/renv/blob/38fbe4d1c67a719afdb8c0de9f2a6295720b62b4/R/paths.R#L73

and the doc still had an example with previous ordering


kevinushey:
Thank you!

Issue #124, opened 2019-08-05
Teach renv about RSPM binaries


slopp:
RSPM repositories can provide users linux package binaries if users set the repo option correctly.

For examples, see: http://cluster.rstudiopm.com/client/#/repos/1/overview

Packrat should remove the linux component of the URL to ensure the projects remain portable across distros.

kevinushey:
Should `renv` also provide some kind of mechanism to opt-in to the use of binaries from RSPM?

slopp:
Users would typically opt-in by setting the repo url, do you have other ideas for how this could work though?

e.g. I guess renv could try to see if a repository supports the __linux__ structure and point the user in that direction? IDK if renv would be able to figure out reliably what OS it is running on and which repo URL to select automatically

kevinushey:
> Users would typically opt-in by setting the repo url, do you have other ideas for how this could work though?

Right, but because only the unadorned repository URL will enter the lockfile (ie: stripped of `/__linux__/<platform>/`), binaries will effectively be "turned off" by default in `renv`. So, I think we need some mechanism for turning that back on. Does that make sense?

Is there a simple way to query an CRAN repository to ask if it's RSPM + binaries are available for a particular platform?

slopp:
@jonyoder  do you know if this this type of info is easily available? Maybe related to https://github.com/rstudio/package-manager/issues/1245

jonyoder:
We have some of this info in the status API:

`http://cluster.rstudiopm.com/__api__/status`:

```json
{
    "version": "1.0.11-1235",
    "metrics_enabled": true,
    "r_configured": true,
    "binaries_enabled": true,
    "binary_distros": [
        {
            "url": "centos6",
            "display": "CentOS/RHEL 6"
        },
        {
            "url": "centos7",
            "display": "CentOS/RHEL 7"
        },
        {
            "url": "opensuse42",
            "display": "OpenSUSE 42.3, SLES 12.3"
        },
        {
            "url": "opensuse15",
            "display": "OpenSUSE/SLES 15"
        },
        {
            "url": "xenial",
            "display": "Ubuntu 16.04 (Xenial)"
        },
        {
            "url": "bionic",
            "display": "Ubuntu 18.04 (Bionic)"
        }
    ]
}
```

[...]

PR #332, opened 2020-02-05
implement support for RSPM binary URLs

kevinushey:
This PR teaches `renv` how to use RSPM binary URLs by default. With this PR, `renv` will transform repository URLs of the form:

    https://rspm.company.org/cran/latest

into

    https://rspm.company.org/cran/__os__/platform/latest

as appropriate. We detect RSPM URLs in the following way:

1. The repository URL contains `/cran/` as a component;
2. The repository responds on the `__api__/status` endpoint;
3. That endpoint reports that binaries are supported for the current platform.

The OS + platform is detected by `renv` using `/etc/os-release`.

@slopp, can you sanity-check?

Closes https://github.com/rstudio/renv/issues/124.


colearendt:
A few potential pain points here:
- will it support "changing" this URL when on a different platform? Or is the URL encoded into the lock file? I.e. switching from Ubuntu to CentOS binaries? Ubuntu to Windows (once supported?)?
- RSPM allows for repos named other than "cran." For instance, our demo server https://demo.rstudiopm.com has `all`, `classroom`, `cran`, `curated`, `dev`, `internal`, `validated`. This generates URLs like `/all/latest` and `/classroom/latest`. The admin gets to decide what these are called (I am not sure off-hand about case sensitivity...). Any chance we can make `renv` more flexible to handle these? A few things that make this challenging:
    - We could just check `/__api__/status` (it looks like CRAN just gives a 404)
    - However, RSPM can also be hosted at non-root paths. I.e. the QuickStart uses `/packages`, which gives us `/packages/all/latest` (with repo named `all`) or `/packages/cran/latest`.

slopp:
Thanks Kevin, I am really excited for this addition!

@jonyoder do you mind taking a look at this as well?

Kevin, we currently support URLs of the form:

`__linux__/platform/id` where platform is one of: `xenial`, `bionic`, `opensuse15`, `opensuse42`, `centos7`

I _think_ if I am understanding the code correctly that you are clean up the results of `/etc/os-release` into these known fields - that's awesome!

A few questions:

- Why limit the search to repos named cran? I think that name can be w/e the admin has specified as the name of their repo in RSPM. 

- For Windows and Mac, we actually don't follow the `__os__/repo/id` pattern. 😢 Instead, we follow what CRAN expects (e.g. for Windows binaries to be served from `repo/bin/windows/x.y`, so for Windows or Mac no transform is needed. (We are currently working on a release that will include Windows support, we don't have Mac binaries yet. You can test Windows support at https://cluster.rstudiopm.com).

slopp:
Does this also address the cache ID concern Cole raised? I didn't grok that change in the code.

kevinushey:
> will it support "auto-changing" this URL when the project is loaded / activated on a different platform? Or is the URL encoded into the lock file? I.e. switching from Ubuntu to CentOS binaries? Ubuntu to Windows (once supported?)? Ubuntu binaries to source (when on Mac)? Thinking mostly about what is committed to VCS.

Yes, it will. The non-binary URL is always saved when the lockfile is generated, and the binary URL is generated from that URL when `renv` loads the project. The binary URL is only used if `renv` can confirm that the requested URL exists -- that is, the `__api__/status` endpoint confirms it is available.

> RSPM allows for repos named other than "cran."

Good point, and we should probably also be prepared for the fact that users might place proxies in front of RSPM and so don't have any obvious hint that a URL is associated with RSPM (or how we can find the status endpoint for a particular URL)

kevinushey:
> Why limit the search to repos named cran? I think that name can be w/e the admin has specified as the name of their repo in RSPM.

This was indeed an oversight; I'll clean this up.

jonyoder:
RSPM supports both CRAN and non-CRAN sources. For example, try visiting the following URLS:

* https://demo.rstudiopm.com/client/#/repos/11/packages/ggplot2 - Note that the UI indicates that the underlying "source" of this package is `cran`.
* https://demo.rstudiopm.com/client/#/repos/11/packages/tidymodels - Note that the UI indicates the "source" of this package as `classroom-packages`.

If we attempt to fetch binaries for non-CRAN packages, RSPM will respond correctly with a source package, but it will create additional load on RSPM since RSPM will go through all the steps to determine if a binary is available for the package.

It might be worth considering adding a way to determine if binaries _should_ be available for a given package based on its source. This would probably require a bit of work in RSPM, but shouldn't be too difficult.

aronatkins:
CC @jskress - should we consult locations other than `/etc/os-release`?

jskress:
Yes, `os-release` by itself is not sufficient.  I've sent what I've learned on to @kevinushey

jonyoder:
@jskress, can you send me the `os-release` info, as well? We do some OS detection in RSPM, too.

kevinushey:
This was merged but somehow not closed by the GitHub UI.

PR #354, opened 2020-02-28
Add missing type for `dependency.errors`

jpeacock29:
This missing info breaks the [table](https://rstudio.github.io/renv/reference/config.html#configuration) of the config docs.


kevinushey:
Thanks, but this change needs to be made in the roxygen documentation. I've done so here:

https://github.com/rstudio/renv/commit/d0c9202b6dafcd8bebdb71a727b4be4eecc283cc

Issue #359, opened 2020-03-03
Python packages used via reticulate are returned as dependencies


jdblischak:
## What I observed

When I run `dependencies()` on a project that imports Python packages with reticulate, the Python packages are returned.

## What I expected

I expected that `dependencies()` would only return R packages. And if it returned more than R packages, there would be a column distinguishing which packages belong to which language.

## Reproducible example

The example below reproduces the observed behavior. I ran it with the latest development version available on GitHub.

``` r
d <- tempfile()
dir.create(d)
f <- file.path(d, "script.R")
writeLines(c(
  "library(reticulate)",
  "pytime <- import(\"time\")",
  "pytime$time()"
  ), con = f)
packageVersion("renv")
#> [1] '0.9.3.43'
renv::dependencies(d)
#> Finding R package dependencies ... Done!
#>                                      Source    Package Require Version   Dev
#> 1 /tmp/RtmputHNtU/file260815a5233f/script.R reticulate                 FALSE
#> 2 /tmp/RtmputHNtU/file260815a5233f/script.R       time                 FALSE
```

<sup>Created on 2020-03-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

<details>

<summary>Session info</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 3.6.3 (2020-02-29)
#>  os       Ubuntu 18.04.4 LTS          
#>  system   x86_64, linux-gnu           
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       America/New_York            
#>  date     2020-03-03                  
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date       lib source                       
#>  assertthat    0.2.1      2019-03-21 [3] CRAN (R 3.5.3)               
#>  cli           2.0.2      2020-02-28 [3] CRAN (R 3.6.3)               
#>  crayon        1.3.4      2017-09-16 [3] CRAN (R 3.5.0)               
#>  digest        0.6.25     2020-02-23 [3] CRAN (R 3.6.3)               
#>  evaluate      0.14       2019-05-28 [3] CRAN (R 3.6.0)               
#>  fansi         0.4.1      2020-01-08 [3] CRAN (R 3.6.2)               
[...]

PR #406, opened 2020-04-28
allow modules::import outside module/amodule

arbelt:
Currently `modules::import` is only recognized as a dependency declaration within a `module::module` or `module::amodule` call. This misses dependencies within files used as modules, e.g., `m <- modules::use("myFileModule.R")`, where `myFileModule.R` contains top-level `modules::import` statements. 

I frequently do this b/c a file with top-level import statements can both be used as a module and be used as an input to `source()`, which make it more "portable" to those not using modules. 

I can't think of a case where `modules::import`would be called without the intention of declaring a dependency, so it seems to makes sense to remove the requirement that modules::import appear within a `module` expression...

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -869,17 +869,29 @@ renv_dependencies_discover_r_pacman <- function(node, stack, envir) {
 
 renv_dependencies_discover_r_modules <- function(node, stack, envir) {
 
+  # check for call of the form 'pkg::foo(a, b, c)'
+  colon <-
+    is.call(node[[1L]]) &&
+    is.name(node[[1L]][[1L]]) &&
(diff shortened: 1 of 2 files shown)


kevinushey:
This change was made due to:

https://github.com/rstudio/renv/issues/359

We might need some smarter way of disambiguating `reticulate::import()` from `modules::import()`.

arbelt:
Thanks for the info!

I've modified the PR to count as deps all _fully qualified_ calls to `modules::import`, whether inside or outside a module expression, while only counting `import(...)` calls when within a module expression. Does that work better?

I've also amended the test file to reflect this logic.

kevinushey:
This looks good to me. Thanks for the PR!

PR #422, opened 2020-05-11
export lockfile function

statnmap:
`lockfile()` function was not exported although available in the documentation.


JosiahParry:
I want to give my +1 to this. Utilizing it here: https://github.com/JosiahParry/manifest-to-renv/blob/main/README.md

kevinushey:
Sorry for the delay, but I'm still not ready to export this.

PR #571, opened 2020-11-17
Specify where latest version of renv is installed from

jdblischak:
I copied the text from `?renv::upgrade`


kevinushey:
Thanks! I made my own change on master here (this also needs to describe the behavior when one is using the development version of `renv`).

PR #596, opened 2020-12-11
Add vignette for using renv with RStudio Connect

blairj09:
This vignette attempts to address frequent questions we've seen from users who want to use `renv` but aren't sure how to do so in combination with RStudio Connect.


vignettes/rsconnect.Rmd
```diff
@@ -0,0 +1,59 @@
...
+## Publishing programatically
+
+When publishing content to RStudio Connect programatically, it is necessary to
+generate a manifest file describing the project environment. This can be done
+with the `writeManifest()` function from the
+[`rsconnect`](https://github.com/rstudio/rsconnect) package. When using `renv`,
+the only thing that needs to be considered is that `rsconnect` should be
+installed and executed from within the `renv` environment so that it recognizes
+the local project library when generating the manifest file. As long as
+`rsconnect` is run from within the `renv` created environment, it will capture
```
akgold, line 48: Worth being a little more explicit about what this means? 

i.e. `As long as you've got the project open in RStudio...`

vignettes/rsconnect.Rmd
```diff
@@ -0,0 +1,59 @@
...
+project dependencies from the local `renv` library.
+
+## A word about packrat
+
+RStudio Connect uses [`packrat`](https://rstudio.github.io/packrat/) to restore
+project environments on the RStudio Connect server. This should have no impact
```
akgold, line 54: Since you haven't mentioned packrat till now, worth connecting to above by mentioning that writeManifest does as well?

kevinushey: I think it's okay to leave this out since (hopefully) future releases of `rsconnect` will use `renv` rather than Packrat as well.

vignettes/rsconnect.Rmd
blairj09: @akgold I think I addressed this well enough in the latest iteration.

blairj09:
@slopp made a point about the fact that the local `renv` environment may be different, and in many cases it will be different, from the deployed environment created by RStudio Connect. I don't think we need to call this out here, since that's not necessarily a departure from normal RStudio Connect behavior. For example, when you publish anything to RStudio Connect, the RStudio Connect environment is not a full representation of your local environment. It's only a representation of **the environment needed to run your content**. That pattern holds true when using `renv`.

blairj09:
@kevinushey I think this is good to go.

vignettes/rsconnect.Rmd
akgold: 👍

kevinushey:
Thanks!

PR #613, opened 2021-01-18
Increase bootstrap reliability by not filtering out extra cran url

dbast:
Rename default extra cran url key to ensure it is not filtered out by the
repo URL deduplication code. It is thus always there as fallback in cases
where the cran url points to a server which is not available, outdated or
broken.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 * Fixed an issue where package installation could fail if the `configure.vars`
   option was set to be a named character, rather than a named list. (#609)
 
+* Increase bootstrap reliability by always trying an extra CRAN URL.
+
 # renv 0.12.5
 
 * Fixed an issue where `renv` would fail to bootstrap. (#608)
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -35,7 +35,7 @@ renv_bootstrap_repos <- function() {
   repos[repos == "@CRAN@"] <- "https://cloud.r-project.org"
 
   # add in renv.bootstrap.repos if set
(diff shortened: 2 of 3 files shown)


tests/testthat/test-bootstrap.R
```diff
@@ -58,6 +58,21 @@ test_that("bootstrap succeeds with empty repos", {
...
+test_that("bootstrap succeeds with fallback URL", {
+
+  skip_on_cran()
+  skip_on_os("windows")
+
+  renv_tests_scope()
+  renv_scope_options(repos = c(CRAN = "https://broken_url"))
+
+  library <- renv_libpaths_default()
+  bootstrap(version = "1.0.0", library = library)
```
dbast, line 70: Version="1.0.0" is a local version and should be changed to a cran available version like 0.12.3 to actually download from cran. But that seems to be not possible during the tests?

dbast: Any hints? Thanks

kevinushey: `renv` sets up its own local CRAN-like repository with "fake" versions of packages available, and those should be used during testing.

(Relying on CRAN itself means tests will fail intermittently if internet connectivity is spotty on the CRAN machines, which we do not want)

dbast: Ok, will then remove the test case, as the changed code is not testable that way.

Anything else to change/do? Any comment to the actual changed code? Thanks

kevinushey:
Thanks!

Issue #424, opened 2020-05-12
`renv::update()` fails for a ssh git package


vspinu:
Basically,

```
devtools::install_git("ssh::/..../some_package.git")
renv::snapshot()
renv::update("some_package")
## * Checking for updated packages ... Error in names(updates) <- extract_chr(updates, "Package") : 
##   attempt to set an attribute on NULL
## Traceback (most recent calls last):
## 2: renv::update("some_package")
## 1: renv_update_find(selected)
```
Sorry, don't have time for full repro ATM. Hopefully the above would be enough. 

This is `[renv 0.10.0] on R 4.0.0`.

kevinushey:
Currently, `renv::update()` only supported CRAN + GitHub packages, so I unfortunately wouldn't expect this to work. It definitely should fail with a better message, though.

vspinu:
I see. I would imagine this could be easely fixed as renv relies on system git command for ssh backend. 

BTW, the current implementation was a blessing for our team as our mac users were not able to install SSH support for git2r. They all rely on renv to fetch the internal packages inside the projects. I hope this behavior won't change soon :)

kevinushey:
I'm glad to hear that :-)

Indeed, I have no intention of switching to `git2r` -- `renv` will continue using `git` from the command-line by default, and if this were to ever change it would be opt-in in some way (but again, that change is unlikely).

mstr3336:
I'm also interested support being added for `update` of generic `git` repos - 
My current workflow for doing this tends to involve tweaking the version number (and removing the hash) for my repos in the lockfile, and then running `renv::restore()`. 

Occasionally, also, I swap between feature branches, main, and release tags while doing so. 

It's a fairly tedious regular chore, and I have to imagine that extending support to these repos is relatively trivial - Is it okay if I put in a PR to do so?

<sub>(Also since we're talking about it, I too love the use of system level git. It just works. Thanks Kevin.)</sub>

kevinushey:
A PR would definitely be appreciated! I think you would need to add a function `renv_update_find_git()`, and use it here:

https://github.com/rstudio/renv/blob/735e17f696e17873895891f4e0a84c578344e0f6/R/update.R#L85-L92

Ultimately, the function should be able to take a package record (which is more-or-less the package name, version, and any remote fields) and use that to produce a new package record mapping to the latest-available version of that package.

matthewstrasiotto:
[...]

PR #618, opened 2021-01-27
Fix issue #424

natbprice:
I added a function `renv_update_find_git()` to check for updates for Git repos. It is not a great solution because I fetch the repo just to see if it has been updated.

### DIFF
--- a/R/retrieve.R
+++ b/R/retrieve.R
@@ -259,6 +259,14 @@ renv_retrieve_git <- function(record) {
 
   renv_git_preflight()
 
+  path <- renv_retrtieve_git_impl(record)
+
+  renv_retrieve_successful(record, path)
+
+}
+
+renv_retrtieve_git_impl <- function(record) {
+
   path <- tempfile("renv-git-")
   ensure_directory(path)
 
@@ -293,8 +301,7 @@ renv_retrieve_git <- function(record) {
     stopf(fmt, record$Package, record$RemoteUrl, status)
   }
 
-  renv_retrieve_successful(record, path)
-
+  return(path)
 }
 
 renv_retrieve_local_find <- function(record, project = NULL) {
--- a/R/update.R
+++ b/R/update.R
@@ -35,6 +35,12 @@ renv_update_find_github <- function(records) {
 
 }
 
+renv_update_find_git <- function(records) {
+
+  renv_parallel_exec(records, renv_update_find_git_impl)
+
+}
+
 renv_update_find_github_impl <- function(record) {
 
   # validate we have a ref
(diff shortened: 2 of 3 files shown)


R/update.R
```diff
@@ -77,6 +83,62 @@ renv_update_find_github_impl <- function(record) {
 
 }
 
+renv_update_find_git_impl <- function(record) {
+
+  # validate we have a ref
+  if (!renv_record_validate(record))
+    return(NULL)
+
+  renv_git_preflight()
```
kevinushey, line 92: Looks like a lot of this was copied from here:

https://github.com/rstudio/renv/blob/25ed22481bdeb89b5dde5f9ab5bc971597b14c4c/R/retrieve.R#L258-L298

Could we refactor that part into a function `renv_retrieve_git_impl(record)`, and then use that both here and in `renv_retrieve_git()`?

kevinushey (commented):
Overall LGTM; thanks for the PR! My main comment is to avoid code duplication; I think that piece of code could be refactored to be shared.

In addition, we require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

R/update.R
natbprice: I finally came back to resolve this, but now I am confused about the intended functionality. 

If an R package is installed from a feature branch, then it seems logical that the update process will only check the feature branch for updates. Similarly, if a package is installed form main branch, then only the main branch will be checked for updates.

If an R package is installed based on a commit ID, then it seems logical that no updates will ever occur because a commit ID will not change. However, the user may intend to install based on a commit ID and then receive updates based on the corresponding branch. 

I am not sure how to code for the different possibilities of what parameters were used to install and whether or not it is possible/logical to perform updates.

kevinushey: Right now, `renv::restore()` always chooses to restore the exact commit associated with whatever reference was supplied. We might change this in the future (likely via an option / opt-in mechanism) but that is currently intentional for both cases.

natbprice:
Can this be merged or are changes needed?

matthewstrasiotto:
hi @natbprice 

I'm putting in a very similar PR, with a more general scope #762  . 
In general, our PRs are compatible - however, our approaches appear to have differed.

Bear with me while I figure out which of @kevinushey's comments have been resolved on your work.

#618 appears to have duplicated some of the git retrieval logic, while #762 avoids this, unless I'm confused. 

I think your naming approach is more conventional than mine, however.

matthewstrasiotto:
I'm going to modify #762 as per this PR, and list you as a co-author if that's okay @natbprice

R/update.R
```diff
@@ -77,6 +83,55 @@ renv_update_find_github_impl <- function(record) {
...
+
+  # if sha is empty, return NULL
+  # probably because recored$RemoteRef is SHA, not branch
+  if (length(sha) == 0) {
+    return(NULL)
+  }
+
+  # check for changed sha
+  if (sha == record$RemoteSha)
+    return(NULL)
```
matthewstrasiotto, line 114: My feedback here:

Redundant logic - renv_retrieve_git_impl will handle this

R/update.R
```diff
@@ -77,6 +83,55 @@ renv_update_find_github_impl <- function(record) {
...
+  current$Version <- desc$Version
+  current$RemoteSha <- sha
+
+  # check that the version has actually updated
+  updated <-
+    current$RemoteSha != record$RemoteSha &&
+    numeric_version(current$Version) >= numeric_version(record$Version)
+
+  if (updated)
+    return(current)
```
matthewstrasiotto, line 131: This can be refactored into `renv_resolve_git_description`

R/retrieve.R
```diff
@@ -259,6 +259,14 @@ renv_retrieve_git <- function(record) {
 
   renv_git_preflight()
 
+  path <- renv_retrtieve_git_impl(record)
+
+  renv_retrieve_successful(record, path)
+
+}
+
+renv_retrtieve_git_impl <- function(record) {
```
matthewstrasiotto, line 268: type in retrieve :)

R/update.R
matthewstrasiotto: I'll refactor this into it's own function so its purpose (short-circuit cloning the repo if possible) is more obvious

natbprice:
Thanks @matthewstrasiotto @kevinushey! It sounds like this is no longer needed because it was incorporated into #762. If so, maybe we can just close this without merging?

matthewstrasiotto:
Your assessment is correct, close it.
Thanks for your work on this PR, it was inminently helpful in mine

PR #633, opened 2021-02-04
Unignored check for file size in case of zip/tar archive during pkg installation

### LINKED ISSUE #504: Add a way to "recover" from the "wrong" repository
I have had it happen several times where I:

- am restoring an environment and want to be sure to use binary packages from RSPM
- start `renv::restore()`
- see packages downloading from `cran.rstudio.com` or some other repo with no binaries
- CTRL+C to abort
- fix the repository URL
- restart `renv::restore()`
- `renv` _still_ installs the downloaded packages from the first pass (i.e. source bundles) and does not go back to retrieve them

It would be great if there were a way to gracefully "abort" and "re-restore" from a fresh state.

donarus:
resolves #504 

Once this PR is merged, one should be able to reinstall a package in case the package size changed in the remote repository.

This scenario can happen for example when a user uses a version identifier like 99.99.99.9000 as a snapshot placeholder, where snapshot means a development version that should reinstalled always when renv::install is called

Another reason for this fix is a scenario where a downloaded package is invalid from somehow reason (network failure, packet loss) and a user needs to reinstall it.

### DIFF
--- a/R/download.R
+++ b/R/download.R
@@ -43,15 +43,6 @@ download <- function(url, destfile, type = NULL, quiet = FALSE, headers = NULL) 
   # on Windows, try using our local curl binary if available
   renv_scope_downloader()
 
-  # if this file is a zipfile or tarball, rather than attempting
-  # to download headers etc. to validate the file is okay, just
-  # check that the archive appears not to be damaged
-  status <- renv_download_check_archive(destfile)
-  if (identical(status, TRUE)) {
-    vwritef("\tOK [file is up to date]")
-    return(destfile)
-  }
-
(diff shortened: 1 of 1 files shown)


kevinushey:
Thanks for the PR! In fact, we've already made this change on master:

https://github.com/rstudio/renv/commit/9ecc2c71c01a7f30bddab9cb870c6b2c3c9b580a#diff-b2404771101f46531301f957fc7e3ebb5179e08f985bb9195b13503b527b43b6

I'm curious why GitHub doesn't seem to see that...

Issue #628, opened 2021-02-01
Feature request: Multiple cache directories


vandenman:
Would it be possible to support multiple cache directories?

For example, `Sys.setenv("RENV_PATHS_CACHE" = "~/.local/share/renv/cache1;~/.local/share/renv/cache2")`, could indicate that both `~/.local/share/renv/cache1` and `~/.local/share/renv/cache2` are directories where renv looks for cached packages. When adding a new package to the cache, renv would use the first writeable path.

My use case is the following: our software consists of multiple modules (basically R packages with a user interface). Each of these modules has its own dependencies, and we supply a default set of R packages through an installer. Right now, we are making these modules more modular so that we can update them independently. During an update, the package dependencies may change. To efficiently reuse available packages, I'd like to use renv with its excellent caching feature.
Now here's the problem. The R packages shipped by default end up in the installation folder of a user. We do not have write access there after installing (at least on macOS and on Linux with flatpak). In our own local folder (inside Appdata, ~/.local/share, etc.) we do have write access. My goal is to create a second cache directory in our local folder. That way, we can 1) efficiently reuse the R packages in the installation folder and 2) reuse R packages installed while updating modules.

The changes required to make this work are rather minimal, at least in the small example that I tested locally. I'd be more than happy to open a PR if you think this could be useful.

kevinushey:
I'd be willing to review a PR but my gut instinct is that there's a fair amount of code internally in `renv` that assumes there is a single cache path and so getting this to work may be challenging. (but I'd be happy to be proven wrong!)

kevinushey:
Also worth saying up-front: we require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories.

PR #653, opened 2021-02-16
Support multiple caches

vandenman:
Fixes https://github.com/rstudio/renv/issues/628 and adds a unit test. I've emailed the contributor agreement.

An example that uses multiple caches:
```r
mkdir <- function(path) {
  if (!dir.exists(path))
    dir.create(path, recursive = TRUE)
  else TRUE
}

chmod <- function(path, mode = c("read", "read+write")) {
  path <- normalizePath(path)
  mode <- match.arg(mode)
  stopifnot(file.exists(path))
  # I think '5' (read and execute) is the minimum that will work.
  # With '4' (read only) things like dir() don't work anymore.
  numbers <- if (mode == "read") "577" else "777"
  system(sprintf("chmod %s %s", numbers, path))
}

cacheDir1 <- "~/.local/share/renvTest/cache1"
cacheDir2 <- "~/.local/share/renvTest/cache2"

# use a ; to distinguish between multiple caches
Sys.setenv("RENV_PATHS_CACHE" = paste(cacheDir1, cacheDir2, sep = ";"))
Sys.getenv("RENV_PATHS_CACHE")

# two paths
renv::paths$cache()

sapply(renv::paths$cache(), unlink, recursive = TRUE)
sapply(renv::paths$cache(), mkdir)

# installed into first path
renv::install("crayon")

sapply(renv::paths$cache(), dir)

chmod(renv::paths$cache()[1], "read")

# installed into second path
renv::install("R.methodsS3")

sapply(renv::paths$cache(), dir)

# remove symlink to crayon and R.methodsS3
file.remove(file.path(renv::paths$library(), c("crayon", "R.methodsS3")))

sapply(renv::paths$cache(), chmod, "read")

# both caches are read-only but linking still works
renv::install("crayon")
renv::install("R.methodsS3")

# new package is installed in the local folder
renv::install("R.oo")

# R.oo is not in the cache but in the local folder
sapply(renv::paths$cache(), dir)
"R.oo" %in% sapply(renv::paths$library(), dir)
```
Note that there is already quite some stuff in place that just 'works' with multiple caches because the code accepts vectorized input. This goes for the changes in `paths.R` but also when copying from the cache. I'm happy to look into more edge cases (cache 1 contains version A of a package while cache 2 contains version B) but I thought this is an okay start.

Comments and feedback are welcome!

### DIFF
--- a/R/paths.R
+++ b/R/paths.R
@@ -17,8 +17,10 @@ renv_paths_common <- function(name, prefixes = NULL, ...) {
     Sys.getenv(envvar, unset = NA) %NA%
     renv_paths_root(name)
 
-  if (identical(name, "cache") && grepl(";", root, fixed = TRUE))
-    root <- strsplit(root, ";", fixed = TRUE)[[1]]
+  # check if the cache consists of multiple paths, if yes then split the paths
+  # this allows mixing read-only and read+write cache directories. See also https://github.com/rstudio/renv/issues/628
+  if (identical(name, "cache") && grepl("[:;]", root))
+    root <- strsplit(root, "[:;]")[[1]]
 
   # form rest of path
   prefixed <- if (length(prefixes))
--- a/R/paths.R
+++ b/R/paths.R
@@ -17,6 +17,9 @@ renv_paths_common <- function(name, prefixes = NULL, ...) {
     Sys.getenv(envvar, unset = NA) %NA%
     renv_paths_root(name)
 
+  if (identical(name, "cache") && grepl(";", root, fixed = TRUE))
+    root <- strsplit(root, ";", fixed = TRUE)[[1]]
+
   # form rest of path
   prefixed <- if (length(prefixes))
     file.path(root, paste(prefixes, collapse = "/"))
--- a/R/cache.R
+++ b/R/cache.R
@@ -114,6 +114,19 @@ renv_cache_synchronize <- function(record, linkable = FALSE) {
 
   # construct cache entry
   cache <- renv_cache_find(record)
+
+  copied <- FALSE
+  for (cachePath in cache) {
+    copied <- renv_cache_synchronize_inner(cachePath, record, linkable, path)
+    if (copied)
+      return(TRUE)
+  }
+  return(FALSE)
+
+}
+
+renv_cache_synchronize_inner <- function(cache, record, linkable, path) {
+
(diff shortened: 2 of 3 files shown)


R/paths.R
```diff
@@ -17,6 +17,9 @@ renv_paths_common <- function(name, prefixes = NULL, ...) {
     Sys.getenv(envvar, unset = NA) %NA%
     renv_paths_root(name)
 
+  if (identical(name, "cache") && grepl(";", root, fixed = TRUE))
```
kevinushey, line 20: Could we add a brief comment here outlining the motivation? Also, I wonder if this should be `:` rather than `;`, or `[:;]`, or `.Platform$path.sep`.

kevinushey (commented):
LGTM; thanks for the PR! The change ended up being much more straightforward than I initially thought.

R/paths.R
vandenman: I added a comment, let me know if it needs to be more or less detailed. Also, I opted for `[:;]` so that code that creates the environment variables can be written in a platform-agnostic manner.

kevinushey:
Thanks; LGTM! (test failures look unrelated and I'm still trying to figure out what's going on with those)

kevinushey:
Sorry; I realize now why we want to avoid `:` on Windows: path components can contain a `:`; e.g. as in `C:/path/to/file`. So we should use `;` on Windows and `:` otherwise (or just `;` uniformly).

vandenman:
Ah yes, that's windows paths starting with `C:/` is an oversight on my side. Either way thanks for merging this!

shug0131:
I'm trying to use multiple caches on windows.  The documentation here https://rstudio.github.io/renv/articles/renv.html#multiple-caches  doesn't mention the semi-colon option for setting the RENV_PATHS_CACHE value. 

But more importantly,  I can successfully change the file/folder permissions  Sys.chmod(list.dirs(), "555")),  but then when I install a new package, it still copies to the cache location that I want to be turned off (global one) so that it resorts to the secondary one (project specific).   So in effect it only ever uses the first cache location and always ignores any others.  What am I doing wrong here ?

kevinushey:
@shug0131 I've started a separate thread here for discussion: https://github.com/rstudio/renv/issues/1094

PR #699, opened 2021-03-26
fix a typo in the local-sources vignette

wenjie2wang:
Thanks for this nice package.  This PR fixes a typo in the `vignette/local-sources.Rmd`.


kevinushey:
Thank you for the kind words and the PR!

Issue #696, opened 2021-03-24
How to permanently remove renv/local from .gitignore?


fdetsch:
I want to ship custom source packages located in `renv/local` along with my Git projects. It seems that as of **renv-0.13.x**, the folder gets added to `renv/.gitignore` automatically upon start-up even if I deleted the entry manually during the last session. 

Is there an option similar to `vcs.ignore.library` that would allow me to permanently NOT ignore this folder? My current workaround is to put the below code chunk in my project local `.Rprofile`:

```r
lns = readLines("renv/.gitignore")
lcl = grepl("local(/)?", lns)
if (any(lcl)) {
  writeLines(
    lns[!lcl]
    , "renv/.gitignore"
  )
}
```

kevinushey:
Thanks for the feature request -- I've added `settings$vcs.ignore.local()` as a way of controlling whether the `local` folder enters the `.gitignore`.

That said, are you sure you want this folder to be included in your Git repository? It could cause your repository to balloon in size over time if you need to add / remove packages from that folder while it is in use.

fdetsch:
Thanks. I agree that this wouldn't make sense for an excessive use of custom pkgs. In my case, it's only a handful of pkgs in the range of a few 100 KBs, which I figure shouldn't be too problematic.

kevinushey:
Makes sense. Thanks for following up!

PR #709, opened 2021-04-07
added missing documentation entry for 'vcs.ignore.local'

fdetsch:
Added missing documentation in `?renv::settings` (see #696).


kevinushey:
Thanks!

PR #719, opened 2021-04-13
Keep only library path subfolders that are valid R package names

HenrikBengtsson:
I noticed that **renv** can complaints if you have non-package subfolders in your library path, e.g.

```r
Project '~/test' loaded. [renv 0.13.2]
The following package(s) are missing their DESCRIPTION files:

    _foobar [/home/hb/test/renv/library/R-4.0/x86_64-pc-linux-gnu/_foobar]

These may be left over from a prior, failed installation attempt.
Consider removing or re-installing these packages.
```

This PR fixes that by ignoring folders that cannot be packages. This also removes the need for hardcoded filters, such as the one used to drop subfolders named `_cache`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: renv
 Type: Package
 Title: Project Environments
-Version: 0.13.2-12
+Version: 0.13.2-14
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


kevinushey:
LGTM; thank you for the PR!

PR #728, opened 2021-04-21
Removes the path->shortpath normalizePath hack when R version >= 4.0.0

JorisGoosen:
I sadly enough cannot give an easily reproducible example of the problems we ran into.

But using this hack that is apparently only necessary on R<4 causes our application (with embedded R) to choke windows somehow. And if the hack isn't necessary anyhow than we prefer not using it at all.


kevinushey:
Thanks!

JorisGoosen:
You're welcome ^^

Issue #733, opened 2021-04-22
messages are suppressed unexpectedly for embedded applications


randy3k:
I understand that the purpose of `renv_session_quiet` is to suppress renv messages when a user runs `R --quiet`. However, it does introduce some unexpected behavior for applications that embeds R.  For example, in [radian](https://github.com/randy3k/radian), we have
```
r$> commandArgs(trailingOnly = FALSE)
[1] "radian"               "--quiet"              "--no-restore-history"
[4] "--no-readline"        "--no-save"            "--no-restore-data"
```
And the `--queit` flag has forced `renv_session_quiet` and hence `renv_load_quiet` to return `TRUE`. 

FYI, reading for `R --help`
```
  -q, --quiet           Don't print startup message
```
It does seem that renv is being a bit aggressive to suppress the messages for `R --quiet`.

PR #734, opened 2021-04-22
stop detecting --quiet and -q

randy3k:
Because they just mean to suppress the startup message
```
  -q, --quiet           Don't print startup message
  --silent              Same as --quiet
```
Not sure if it is the best way to handle #733, but this is a solution.


kevinushey:
Thanks!

PR #742, opened 2021-05-04
Fix box::use() declarations with trailing comma

klmr:
As of ‘box’ v1.0.1, the `box::use()` declaration allows trailing commas. This PR adds support for this syntax to the ‘renv’ dependency discovery. Without this PR, presence of such a trailing comma causes a crash during dependency discovery.

The PR also adds a test case which runs successfully with the change. However, some of the other tests (in `test-actions.R` and `test-migrate.R`) reproducibly fail on my system (regardless of these code changes), probably due to mismatching configuration.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # renv 0.14.0 (UNRELEASED)
 
+* Fixed crash during dependency discovery when encountering `box::use()`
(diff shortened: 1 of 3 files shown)


klmr:
Found something else, and it makes sense to merge the two issues into one PR.

PR #743, opened 2021-05-04
Update handling of box::use() declarations

klmr:
Handling of `box::use()` declarations requires two slight adjustments:

* as of ‘box’ v1.0.1, `box::use()` permits trailing commas
* `box::use()` supports relative module references via `.` and `..`, which are currently incorrectly parsed as package dependencies

This PR adds support for this syntax to the ‘renv’ dependency discovery. Without this PR, presence of a trailing comma causes a crash during dependency discovery, and `.` and `..` are incorrectly treated as package dependencies.

The PR also adds a test case which runs successfully with the change. However, some of the other tests (in `test-actions.R` and `test-migrate.R`) reproducibly fail on my system (regardless of these code changes), probably due to mismatching configuration.


kevinushey (approved):
LGTM, thanks!

PR #757, opened 2021-05-31
Check for .Rprofile in home directory

jolars:
The following line does not work when `R_PROFILE_USER` is set to `""`, which it apparently is on my installation (Fedora 34).

https://github.com/rstudio/renv/blob/bfc791ba80ea7f8e5669538a9ba0552008ad2e14/R/load.R#L263

This means that renv doesn't find my profile (in `~/.Rprofile`) when I set `RENV_CONFIG_USER_PROFILE=TRUE`.

With this pull request, the function now explicitly checks the home folder for the file. I removed the input to `unset`, but maybe it could've stayed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: renv
 Type: Package
 Title: Project Environments
-Version: 0.13.2-59
+Version: 0.13.2-60
 Authors@R: c(
     person("Kevin", "Ushey", role = c("aut", "cre"), email = "kevin@rstudio.com"),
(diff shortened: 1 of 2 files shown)


kevinushey:
I'm not entirely convinced this is the right fix -- shouldn't the change be made in your R distribution? Where and why is `R_PROFILE_USER` being set to the empty string?

jolars:
You're right. In fact I just realized that this behavior comes from the VS Code R extension. Sorry for the inconvenience!

kevinushey:
No worries, thanks!

PR #760, opened 2021-06-01
Bug subsubdir support

### LINKED ISSUE #759: `renv` installation / remotes syntax conflict with remotes/devtools for deeper nested packages
I have come across an issue (similar to #497) where I am unable to install a package using `remotes` / `devtools` syntax for GitHub based packages nested more deeply than at the top level / 1 folder down in a repository.

This is an issue for me since we use a monorepo at work, and I am sometimes not able to keep a package high up enough in the repo structure to keep `renv` happy.

I have created a [repo and instructions](https://github.com/gtm19/subsub.renv) which demonstrates that _either_ `remotes`/`devtools` OR `renv` is happy, but in short, installation like follows results in an error using `renv`, despite `remotes::install_*` being happy:

```r
desc::desc_set_remotes("github::gtm19/skeleton/subdir/subsubdir")
renv::install()
# Error: failed to resolve remote 'github::gtm19/skeleton/subdir/subsubdir' -- failed to retrieve 'https://api.github.com/repos/gtm19/skeleton/subdir' [error code 22]
# In addition: Warning message:
# curl: (22) The requested URL returned error: 404
```

Whereas making `renv` happy then makes `remotes` / `devtools` unhappy:

```r
desc::desc_set_remotes("github::gtm19/skeleton:subdir/subsubdir")
devtools::build()
devtools::install_local("../subsub.renv_0.0.0.9000.tar.gz")

# Error: Failed to install 'subsub.renv' from local:
#  HTTP error 404.
#  Not Found

#  Did you spell the repo owner (`gtm19`) and repo name (`skeleton:subdir`) correctly?
#  - If spelling is correct, check that you have the required permissions to access the repo.
# Execution halted
```

I believe I have a fix for this which passes all existing `renv` tests. I will PR shortly.

gtm19:
Fixes #759

* Harmlessly (to tests, at least) modifies the regex in `R/remotes.R` to allow installation from sub-subdirectories of repositories
* Adds tests for parsing and installing such sub-subdirectory packages
* Bumps version
* Info added to `NEWS.md`


kevinushey:
Thanks for the PR! My main concern is for e.g. Gitlab, where projects can be nested. Unfortunately I didn't previously have a test for this in `renv`, but have just added one here:

https://github.com/rstudio/renv/blob/1511f184f0825849bd877724e22cdcd992302fe4/tests/testthat/test-remotes.R#L123-L148

Would your PR still pass with that?

Note that this is part of the reason why `renv` uses the `:` syntax for sub-directories -- unfortunately the `remotes` syntax is ambiguous as to whether `/` should refer to a sub-group or to a sub-directory within the associated project; using `:` as a delimiter between the repository-proejct identifier and the sub-directory within the project makes this more clear.

For what it's worth, I _think_ the right fix here would be to tweak the parser to handle `foo/bar/baz` specially for GitHub remotes; likely to be done here.

https://github.com/rstudio/renv/blob/1511f184f0825849bd877724e22cdcd992302fe4/R/remotes.R#L148-L163

(In fact, I think the original intention of that block of code was to handle the `foo/bar/baz` syntax in a remotes-compatible way...)

kevinushey:
Looking a little closer, I believe the bug actually is in https://github.com/rstudio/renv/blob/1511f184f0825849bd877724e22cdcd992302fe4/R/remotes.R#L148-L163. We're only taking the last path component, but we should be taking all but the first instead. I'll make a PR for that (+ test).

kevinushey:
I think this is fixed on master with https://github.com/rstudio/renv/commit/e5a7319ca5f711a64a8affabca9f2c995919e3ad -- would you be able to test?

gtm19:
@kevinushey - it passes the tests I added to this PR 👍🏻 

Thanks 😄

kevinushey:
Should be resolved on master now -- thanks for the PR!

PR #762, opened 2021-06-03
Allow parsing of git spec (as per remotes)

### LINKED ISSUE #424: `renv::update()` fails for a ssh git package

matthewstrasiotto:
Resolves #424


matthewstrasiotto:
@kevinushey  interested to hear if you think that adding another function to `methods` here sounds like a good approach for resolving recursive dependencies

https://github.com/matthewstrasiotto/renv/blob/6ac9f1f9d957038e1c1fd6ad682d2e655e4258b0/R/available-packages.R#L318-L333

kevinushey:
Shouldn't recursive git dependencies be handled via those packages declaring so in a suitably-specified `Remotes:` field?

kevinushey:
In addition, require a [Contributor License Agreement](https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf) before we can accept pull requests on RStudio repositories. Would you be willing to fill one out, and send it to the e-mail indicated in the form? (Please also reference this PR in your e-mail.)

R/remotes.R
```diff
@@ -145,6 +188,19 @@ renv_remotes_parse_finalize <- function(parsed) {
 
 }
 
+renv_remotes_is_git <- function(parsed) {
+  return(
+    "git" %in% c(
```
kevinushey, line 193: Can you clarify why this is necessary?

matthewstrasiotto: I'll leave a comment, but essentially I wanted to be able to parse a git url that wasn't given explicitly with a type, eg:

git@github.com:user/repo.git

The hints are the git@, and the .git

matthewstrasiotto:
Happy to fill a CLA out, though I have done this in the past for other rstudio projects- would that still apply

matthewstrasiotto:
> Shouldn't recursive git dependencies be handled via those packages declaring so in a suitably-specified `Remotes:` field?

They should, I'll give a proper reprex but essentially, at the moment if we have a package:

```dcf
Package: pkg1
Remotes:
   git::git@host:user/pkg2.git
```

Say we install pkg2 directly:

```r
renv::install("git::git@host:user/pkg2.git")
```

This just tends to work

But say we install the dependent package:

```r
renv::install("git::git@host:user/pkg1.git")
```

This will fail when installing dependant packages atm.

Previously `renv` would simply skip `git::` remotes specs, (couldn't parse, gave up) which was mostly fine, 

But with my changes, it seems to parse them just fine, but doesn't appear to have a way to resolve them when querying available packages (I think this is what's happening)

I get the impression that this is the last step before this PR is complete, (except for addressing review comments, and tests)

matthewstrasiotto:
In terms of testing, there are a few obvious unit tests I can write, is there any "right" way to write end-to-end type tests that covers installing, for example, a package with a git dependency

R/remotes.R
matthewstrasiotto: I'm at a computer so I can explain further- I want the following to work: `git@github.com:user/repo.git`

matthewstrasiotto:
Okay, so, I've made a sort of minimal reprex of packages that include other git packages, and wasn't able to reproduce the bug.

See my packages: [renv.git1](https://github.com/matthewstrasiotto/renv.git1) &
[renv.git2](https://github.com/matthewstrasiotto/renv.git2)

If you'd like to try to install them, (personally, I just did this directly from my own `renv` fork) - you can have a crack like so:

```r
# any of the following should work
url_formats <- c(
  "git::git@github.com:matthewstrasiotto/renv.git1.git",
  "git@github.com:matthewstrasiotto/renv.git1.git",
  "git::https://github.com/matthewstrasiotto/renv.git1.git",
  "https://github.com/matthewstrasiotto/renv.git1.git"
)
renv::install(url_formats[[1]])
```
Outputs:
```
Installing renv.git2 [0.0.0.9000] ...
	OK [built from source]
Copying renv.git2 [0.0.0.9000] into the cache ...
	OK [copied to cache in 17 milliseconds]
Installing renv.git1 [0.0.0.9000] ...
	OK [built from source]
Copying renv.git1 [0.0.0.9000] into the cache ...
	OK [copied to cache in 15 milliseconds]
```

R/remotes.R
matthewstrasiotto: May I ask - Are you asking me to add a comment, or were you asking me right now what the change is for?

R/remotes.R
```diff
@@ -176,6 +232,12 @@ renv_remotes_parse <- function(entry) {
     return(parsed)
   }
 
+  parsed <- catch(renv_remotes_parse_git(entry))
+  if (!inherits(parsed, "error")) {
+    parsed$type <- parsed$type %||% "git"
+    return(parsed)
+  }
+
```
matthewstrasiotto, line 240: Upon reflection - https://github.com/rstudio/renv/pull/762/files#r644995771 seems redundant given this block

matthewstrasiotto: Wrong - `renv_remotes_parse_finalize` will default anything to `"github"` before `parsed$type <- parsed$type %||% "git"`  can default it to `"git"`

R/remotes.R
matthewstrasiotto: Maybe it's not necessary to infer type without the `git::` prefix - Also - given that `https://` git links don't work correctly without the `git` prefix, and it's probably be a fair bit more work to resolve that, perhaps this inconsistency implies the extra complexity is not worth it.

Interested in your thoughts @kevinushey

matthewstrasiotto:
Outstanding problems:

- [ ] Git type inference fails for `https://` urls
  - [ ] Potentially remove `"git"` type inference, and expect `"git"` remotes to be explicitly typed.
- [ ] When resolving `git` packages, package name & package version are not currently resolved, but instead just "guessed"

R/remotes.R
matthewstrasiotto: @kevinushey would you like me to remove this type inference?

kevinushey: I think so. I think any of the inference needs to happen up-front when `renv_remotes_parse_git()` is called (assuming it detects what appears to be a compatible `git` reference). Once that type has been inferred, it can be saved and reused.

In other words, if the code reaches here:

https://github.com/rstudio/renv/pull/762/files#diff-c30ff3320c29175c880b55adb3cb4229d4b0840e9a6f9df162ec8760216a063dR174

then I think you could just "force" the setting of the type to be "git", since the git parser successfully detected it. Does that make sense?

kevinushey: We could also take the view that this parser only considers entries that end with `.git`, since it would be unlikely that any other entry type would use that suffix. That might help simplify some pieces and remove potential ambiguities with the other parsers.

matthewstrasiotto: > We could also take the view that this parser only considers entries that end with `.git`, since it would be unlikely that any other entry type would use that suffix.

To clarify - do you mean that this parser should reject an entry that does not end with `.git`, for example:

- Works:
  - `git::git@github.com:user/repo.git`
- Fails:
  - `git::git@github.com:user/repo`

If so, I disagree with your intepretation, as `git@github.com:user/repo` is a valid route on most git servers.

Or do you mean - the parser should only use the explicit `git::` prefix, or `.git` extension to infer `git` type?

Unfortunately, without more extensive changes, the `.git` extension wouldn't force `git` type inference reliably, as in:

- This will fail unless I change more stuff
  - `https://github.com/user/repo.git` 
  - `https` is handled specially, earlier.

> That might help simplify some pieces and remove potential ambiguities with the other parsers.

matthewstrasiotto:
Migrating to #776

Issue #775, opened 2021-06-10
Can't resolve `git::` entries in remotes


matthewstrasiotto:
Adding this to formally document something discussed in #424 (related issues)

PR #776, opened 2021-06-10
Resolve git dependencies

matthewstrasiotto:
Resolves: #424 
Resolves: #775

Combination of #762 & #618 

These PR's both use logic that's fairly tightly coupled, their features are highly related, & to cleanly implement one without the other would involve de-duplicating logic across the two.

Co-authored-by: Nat Price <nathaniel.price@icf.com>


matthewstrasiotto:
This is failing TCs on "remotes" - Not sure why exactly, is this a GITHUB_PAT thing or something?

R/remotes.R
```diff
@@ -145,6 +190,36 @@ renv_remotes_parse_finalize <- function(parsed) {
...
+      parsed$type,
+      parsed$login,
+      parsed$ext,
+      parsed$protocol
+    )  ||
+      parsed$protocol == "ssh"
+  )
+}
+
+
```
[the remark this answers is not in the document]
matthewstrasiotto: I've reduced this scope & moved the logic to within `renv_remotes_parse_git` as requested

matthewstrasiotto:
To implement tests, I probably need some kind of git remotes. 
I doubt the maintainers are comfortable with me linking to my personal `git` repos for these tests - 

Would the maintainers be okay with hosting test repos, similar to [kevinushey/skeleton](https://github.com/kevinushey/skeleton)?

R/remotes.R
```diff
@@ -394,6 +473,110 @@ renv_remotes_resolve_github <- function(entry) {
...
+  record$Version <- desc$Version
+
+  record
+}
+
+
+renv_remotes_resolve_git_sha_ref <- function(record) {
+  renv_git_preflight()
+  template <- c(
+    "git ls-remote \"${ORIGIN}\" \"${REF}\" | head -1 | sed 's/\t.*//'"
```
kevinushey, line 523: This may not work on Windows (where `head` and `sed` are not necessarily available / on the PATH).

Could we just use the result of `git ls-remote \"${ORIGIN}\" \"${REF}\"` and then filter or transform the output in R?

R/remotes.R
```diff
@@ -132,8 +133,52 @@ renv_remotes_parse_remote <- function(entry) {
...
+      "pull",
+      "ref"
+    )
+  parsed <- renv_remotes_parse_impl(entry, pattern, fields, perl = TRUE)
+
+  if (!nzchar(parsed$repo))
+    stopf("'%s' is not a valid remote", entry)
+
+  # If type has not been found & repo looks like a git repo, set it as git
+  if (renv_remotes_is_git(parsed))
```
kevinushey, line 175: I'm still wondering why this check is needed -- if the above regular expression test succeeds, doesn't that imply that this is a `git` remote? Or are there cases where this regular expression will capture some URL / specification that is not actually to be used by `git`? (If so, an example would be helpful)

kevinushey: (FWIW I think the implementation may be good as-is, I just want to double-check)

R/remotes.R
```diff
@@ -394,6 +473,110 @@ renv_remotes_resolve_github <- function(entry) {
...
+
+  return(sha)
+}
+
+
+renv_remotes_resolve_git_description <- function(record) {
+  path <- tempfile("renv-git-")
+  ensure_directory(path)
+
+  renv_retrieve_git_impl(record, path)
```
kevinushey, line 557: Do you know if it is possible to check out only a single file from a git repository? It would be nice to be able to avoid a whole clone just to get the DESCRIPTION.

kevinushey (commented):
Overall LGTM; some comments around tying up potential loose ends.

R/remotes.R
matthewstrasiotto: there are cases where the regex would capture non-git URLs - it's able to capture a lot of URLS, for example:

```r
renv:::renv_remotes_parse_git("http://cran.com/non-git")
```

would yield:

```
$entry
[1] "http://cran.com/non-git"

$type
NULL

$url
[1] "http://cran.com/non-git"

$protocol
[1] "http"

$login
NULL

$host
[1] "cran.com"

$user
[1] "non-git"

$repo
NULL

$ext
NULL

$subdir
NULL

$pull
NULL

$ref
NULL
```

R/remotes.R
matthewstrasiotto: >  It would be nice to be able to avoid a whole clone just to get the DESCRIPTION.

It sure would- unfortunately, from googling around on the subject, there doesn't seem to be many robust, portable options for this.

- Some of the git mechanisms require the following not-so-portable conditions:
  - the remote git server has specific options enabled, which some (eg github) don't
  - git client version >= 2.25ish (that's fairly modern & not likely on many distros, though `renv` does already require a minimum git version that's somewhat modern)
- OR, they require the refspec meet restrictions that don't work for us, such as being an actual formal ref, as opposed to a commit SHA, etc.

Future work might be to write up an implementation that can try out some of these options to reduce overhead, however, the script would probably be non-trivial, and involve stuff like:

- try out mechanism 1, check the return code
- try out mechanism 2, check it was a valid ref that could be fetched
- etc

Because the `git` protocol should be as general as possible to any `git` repo, unfortunately I think this is still the best bet, and we can try to improve upon this down the line.

R/remotes.R
matthewstrasiotto: > > Would the maintainers be okay with hosting test repos, similar to [kevinushey/skeleton](https://github.com/kevinushey/skeleton)?
> FWIW I've now forked those two repositories to my own account, if you want to use those versions in tests.

Great, I'll update the entries in `renv.git1`'s `Remotes:` field to match your urls & put in a PR or however it works when you own the upstream

R/remotes.R
matthewstrasiotto: I had the same thought while adapting it! (& ignored it for some reason) 

I agree it makes more sense to rely on R to manipulate the output string

R/remotes.R
matthewstrasiotto: I'll leave this unresolved so you can have a read of it, but for now, I think we should accept the "clone whole repo" implementation as "good enough"

matthewstrasiotto:
> > Would the maintainers be okay with hosting test repos, similar to [kevinushey/skeleton](https://github.com/kevinushey/skeleton)?
> FWIW I've now forked those two repositories to my own account, if you want to use those versions in tests.

Great, I'll update the entries in `renv.git1`'s `Remotes:` field to match your urls & put in a PR or however it works when you own the upstream

R/remotes.R
kevinushey: Thanks! Given your investigation I agree. (Perhaps in the future we could implement something like "if git.version >= 2.25, then use the clever implementation; otherwise, fall back to the portable one")

kevinushey:
Awesome work! This LGTM now -- is there anything else you think needs to be addressed before merge?

matthewstrasiotto:
Oh! Maybe i need to add `skip_on_cran()` (idk about the R version requirement) to my test cases?

matthewstrasiotto:
I think it's ready, @kevinushey :)

kevinushey:
Awesome -- thank you for the PR!

Issue #801, opened 2021-07-08
renv::dependencies() skips R utility script files without file extension


klmr:
I frequently use R to write command line utilities, especially to do common build tasks inside R projects. By common POSIX convention these scripts don’t have a filename extension, carry the executable flag, and contain a suitable shebang which execs Rscript (`#!/usr/bin/env Rscript`). They’re commonly located in a subdirectory called `scripts` or `build` or similar.

Unfortunately `renv::dependencies` doesn’t scan such files. Thus all dependencies used in these files are missed.

I’d be happy to submit a PR for this. As far as I can see this would entail adding a fallback to `renv_dependencies_callback` — something along these lines:

```diff
diff --git a/R/dependencies.R b/R/dependencies.R
index f883a7bf..48096cde 100644
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -242,8 +242,14 @@ renv_dependencies_callback <- function(path) {
     ".rnw"         = function(path) renv_dependencies_discover_multimode(path, "rnw")
   )
 
-  cbname[[basename(path)]] %||% cbext[[tolower(fileext(path))]]
+  cbname[[basename(path)]] %||% cbext[[tolower(fileext(path))]] %||% {
+    if (is_r_executable(path)) renv_dependencies_discover_r(path)
+  }
+
+}
 
+is_r_executable <- function(path) {
+  grepl("^#!.*\\b(R|r|RScript)\\b", readLines(path, n = 1L))
 }
 
 renv_dependencies_find_extra <- function(root) {
```

Untested; of course a PR would include unit tests. Potentially this should also test the executable flag on systems which support this. — Maybe via `(file.info(path)$mode & S_IXUSR) != 0L` (with `S_IXUSR <- 64L`), but that code is probably platform-dependent (at the very least I’d have to check how Windows permissions map to POSIX file modes).

kevinushey:
I'd be open to a PR. My main concern is we should do our best to avoid reading files that we won't need to read; e.g. perhaps we should only do this check for:

- Files that are executable,
- files that don't have an extension (?),
- Files that live in a particular directory (?)

Do we have a strategy for avoiding binary files (since binary executables might be included in an `renv` project)? I wonder if we should first read the file as binary (e.g. the first 512 bytes), discard anything containing a nul byte, and then try to parse the shebang line from that.

The counter-argument, I think, is that a well-constructed `.renvignore` would also be a fix for over-aggressive dependency scanning.

[...]

PR #803, opened 2021-07-10
Add dependency discovery support for utility scripts

klmr:
Implements #801 

If a file has no file extension, this PR adds code that opens the file in binary mode and reads the first two bytes; if these bytes are equal to `#!`, it continues to read the first line and checks whether it contains the word `R`, `r` or `RScript`. If so, this file is treated as an R script file.

For now this PR does *not* check whether the file is executable, since file executability is not a cross-platform property, which would thus lead to different results of `renv::dependencies()` across platforms (in particular, `file.access` does not consider file mode flags on Windows; instead, it tests for specific file extensions).

I’m unsure whether the helper function location (in `R/dependencies.R`) is correct within the project’s scope: maybe they should instead go into `R/utils.R`?

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -249,25 +249,13 @@ renv_dependencies_callback <- function(path) {
 }
 
 is_r_executable <- function(path) {
+  # don't check executability via `file.access` since this is implemented
+  # differently on Windows, and would thus lead to different dependencies being
+  # discovered across platforms
   if (nzchar(fileext(path)))
     return(FALSE)
 
-  grepl("\\b(R|r|Rscript)\\b", get_shebang_command(path))
-}
-
-get_shebang_command <- function(path) {
-  con <- file(path, open = "rb")
-  on.exit(close(con), add = TRUE)
-
-  signature <- charToRaw("#!")
-  first_two_bytes <- readBin(con, what = "raw", n = length(signature))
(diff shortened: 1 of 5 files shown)


R/dependencies.R
```diff
@@ -242,8 +242,32 @@ renv_dependencies_callback <- function(path) {
...
+    if (is_r_executable(path)) function(path) renv_dependencies_discover_r(path)
+  }
+
+}
 
+is_r_executable <- function(path) {
+  if (nzchar(fileext(path)))
+    return(FALSE)
+
+  grepl("\\b(R|r|Rscript)\\b", get_shebang_command(path))
```
kevinushey, line 255: Could we add a comment here explaining why we don't check if the file is executable (ie, for Windows)?

R/dependencies.R
```diff
@@ -242,8 +242,32 @@ renv_dependencies_callback <- function(path) {
...
+}
+
+get_shebang_command <- function(path) {
```
kevinushey, line 258: What about putting this in `files.R` and calling it `renv_file_shebang()`?

kevinushey (commented):
LGTM; I provided one suggestion for renaming the shebang utility function.

R/dependencies.R
klmr: How about now?

kevinushey:
Thanks, LGTM!

PR #824, opened 2021-09-02
Using a singularity container is implicit consent

kiwiroy:
In similarity to using a docker container, this pull request adds a test for the `/.singularity.d` directory in a singularity container, to provide implicit consent.

Thanks for considering this.

### DIFF
--- a/R/consent.R
+++ b/R/consent.R
@@ -79,6 +79,7 @@ renv_consent_check <- function() {
     !is.na(Sys.getenv("CI", unset = NA)) ||
     !is.na(Sys.getenv("GITHUB_ACTION", unset = NA)) ||
     !is.na(Sys.getenv("RENV_PATHS_ROOT", unset = NA)) ||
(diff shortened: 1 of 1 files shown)


R/consent.R
```diff
@@ -79,6 +79,7 @@ renv_consent_check <- function() {
     !is.na(Sys.getenv("CI", unset = NA)) ||
     !is.na(Sys.getenv("GITHUB_ACTION", unset = NA)) ||
     !is.na(Sys.getenv("RENV_PATHS_ROOT", unset = NA)) ||
+    dir.exists("/.singularity.d") ||
```
kevinushey, line 82: Could we use `file.exists()` here instead? (Just because `dir.exists()` was added in R 3.2.0 and, at least in theory, some might be using `renv` with an older verison of R.)

kiwiroy: Sure, good catch.

Issue #863, opened 2021-11-12
option for ignore rules


aronatkins:
renv should allow callers to indicate project-level ignore rules using an in-memory option. This would let us map the packrat option ignored.directories to an renv setting, for example.

We are considering a function like the following to create renv rules from the packrat setting:

```r
# Return renv ignore patterns based on the packrat ignored.directories option.
# Each directory is returned as a rooted pattern for renv, meaning that it
# should only apply at the root directory of the project.
#
# Note: The "/data/" and "/inst/" directories are ignored by default.
ignoredDirsForRenv <- function(ignoredDirs) {
  if (length(ignoredDirs) > 0) {
    # Make sure all the directories end with a slash.
    ignoredDirs <- ifelse(
        substr(ignoredDirs, nchar(ignoredDirs), nchar(ignoredDirs)) != "/",
        paste0(ignoredDirs, "/"),
        ignoredDirs
    )
    # Make sure all the directories begin with a slash.
    ignoredDirs <- ifelse(
        substr(ignoredDirs, 1, 1) != "/",
        paste0("/", ignoredDirs),
        ignoredDirs
    )
    ignoredDirs
  }
}
```

That code would then be called like:

```r
do_some_thing <- function() {
    old_renv_ignore <- options("renv.settings.ignore.matching" = ignoredDirsForRenv(packrat::opts$ignored.directories())
    on.exit(do.call(options, old_renv_ignore), add = TRUE)
    deps <- renv::dependencies(".")
    # ...
}
```

The `renv.settings.ignore.matching` setting name is an example of what renv could use; it does not need to use this name.

kevinushey:
Some questions:

1. Should these rules override the default `renv` ignore rules, or supplement them?

2. What is the "right" behavior for a person who has opted into using `renv`? (For example, suppose they're deploying a project that already has `renv.lock`, or they're already using an `.renvignore`). This might be a concern for Packrat, not renv, though.

aronatkins:
I don't have great answers to either of those questions.

1. Maybe the best example to follow is the `$HOME/.gitignore` that folks use. These are ignore rules that are applied to all projects. They augment the project-specific ignore rules. I vote supplement.

2. You're right - this is more of a question for packrat / other callers who need to decide how to combine renv with the choices they've already made. I was hoping that the `renv` rules (given `.renvignore` files) combined with the `ignored.directories` constraint from `packrat` would be sufficient.

PR #866, opened 2021-11-12
allow renv ignores to be set via option

kevinushey:
Closes https://github.com/rstudio/renv/issues/863.

This PR allows one to set `.renvignore`-style ignores for `renv` via the option `renv.renvignore.exclude`. These can either be `.gitignore`-style ignores, or (if the `asis` attribute is set to `TRUE`) PCREs which match against the full file path. (Directories will have a trailing `/` appended to the path.)

The option is primarily intended for internal consumption, so I haven't documented any explicit interface here.

I think the usage for Packrat here would be something like:

```
# get ignores
ignores <- packrat::opts$ignored.directories()

# post-process ignore in some appropriate way

# tell renv to use the pattern as-is
attr(ignores, "asis", TRUE) <- TRUE

# set the option
options(renv.renvignore.exclude = ignores)
```

Let me know if this seems sane.

### DIFF
--- a/R/renvignore.R
+++ b/R/renvignore.R
@@ -45,6 +45,11 @@ renv_renvignore_pattern <- function(path = getwd(), root = path) {
   include <- unlist(extract(patterns, "include"))
   exclude <- unlist(extract(patterns, "exclude"))
 
+  # allow for inclusion / exclusion via option
+  # (primarily intended for internal use with packrat)
+  include <- c(include, renv_renvignore_pattern_extra("include", root))
+  exclude <- c(exclude, renv_renvignore_pattern_extra("exclude", root))
+
   # ignore hidden directories by default
   exclude <- c("/[.][^/]*/$", exclude)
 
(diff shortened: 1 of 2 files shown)


aronatkins (approved):
This change appears to work on my end.

Code like the following converts from the `ignored.directories` setting to an renv ignore pattern:

```r
ignoresForRenv <- function(ignoredDirectories) {
  ignores <- NULL
  if (length(ignoredDirectories) > 0) {
    ignores <- ignoredDirectories
    # Make sure all the directories end with a slash.
    ignores <- ifelse(
        substr(ignores, nchar(ignores), nchar(ignores)) != "/",
        paste0(ignores, "/"),
        ignores
    )
    # Make sure all the directories begin with a slash.
    ignores <- ifelse(
        substr(ignores, 1, 1) != "/",
        paste0("/", ignores),
        ignores
    )
    attr(ignores, "asis") <- TRUE
    cat(paste0("renv ignore: ", ignores), sep = "\n")
  }
  ignores
}
```

That then feeds into an option-set:

```r
  old_renv_exclude <- options("renv.renvignore.exclude" = ignoresForRenv(opts$ignored.directories()))
  on.exit(do.call(options, old_renv_exclude), add = TRUE)
```

One thing I had not done previously was forcing the same `path` and `root` for `renv::dependencies`; most of my previous experimentation left `root = NULL` (but I had been careful to `setwd`). With this set of changes, I have all tests passing.

PR #867, opened 2021-11-15
Don't write to global cache when cache is disabled

### LINKED ISSUE #837: {renv} tries to write to global package cache although it is disabled
{renv} is used in pre-commit.com, `language: r` and recently, support for pre-commit.ci was [added](https://github.com/pre-commit-ci/runner-image/pull/80). However, there are strange errors occurring sometimes when multiple processes try to activate a renv (governed by `require_serial:` in the hook specs file `.pre-commit-config.yaml`). The error is:

```
failed to create directory at path '/tmp/renv'
Traceback (most recent calls last):
19: suppressWarnings({
        old <- setwd("/pc/clone/xQ5UXFOMRsuyt4tMlwQDwQ/renv-default")
        source("renv/activate.R")
        setwd(old)
        renv::load("/pc/clone/xQ5UXFOMRsuyt4tMlwQDwQ/renv-default")
    })
18: withCallingHandlers(expr, warning = function(w) if (inherits(w, 
        classes)) tryInvokeRestart("muffleWarning"))
17: source("renv/activate.R")
16: withVisible(eval(ei, envir))
15: eval(ei, envir)
14: eval(ei, envir)
13: local(...)
12: eval.parent(substitute(eval(quote(expr), envir)))
11: eval(expr, p)
10: eval(expr, p)
 9: eval(quote(...), new.env())
 8: eval(quote(...), new.env())
 7: renv_bootstrap_load(project, libpath, version)
 6: renv::load(project)
 5: renv_load_project(project)
 4: ensure_parent_directory(projects)
 3: ensure_directory(unique(dirname(path)))
 2: stopf("failed to create directory at path '%s'", rownames(info))
 1: stop(sprintf(fmt, ...), call. = call.)
Execution halted
```
E.g. in this [log](https://results.pre-commit.ci/run/github/85580810/1633372635.nQEJrJf1SgiamgC21Ssl_A). One should note that {renv} does not have the permission to write to this directory anyways in the container where the above log comes from. But the error does not occur unless parallel processing is enabled.

Not sure what is going on. We configured {renv} to not even use the global cache as it is deactivated using various env variables in https://github.com/pre-commit-ci/runner-image/blob/master/Dockerfile#L191-L195, i.e.:
```
RENV_CONFIG_CACHE_ENABLED=false 
RENV_CONFIG_CACHE_SYMLINKS=false 
RENV_PATHS_ROOT=/tmp/renv
``` 
`/tmp/renv` does not exist in the container. Creating it beforehand leads to the same error. I am not sure it would resolve the problem but `renv_load_project`, I think we should not try to create the parent directory of `projects` and write files if the global cache is disabled,  i.e.   `if (project %in% projlist)` should be extended to return `TRUE` if `RENV_CONFIG_CACHE_ENABLED` (and the corresponding option/config/R option) is set to `False`, 0, etc.

```r
renv_load_project <- function(project) {

  # record the active project in this session
  project <- renv_path_normalize(project, winslash = "/")
  Sys.setenv(RENV_PROJECT = project)

  # read project list
  projects <- renv_paths_root("projects")
  projlist <- character()
  if (file.exists(projects))
    projlist <- readLines(projects, warn = FALSE, encoding = "UTF-8")

  # if the project is already recorded, nothing to do
  if (project %in% projlist) 
    return(TRUE)

  # otherwise, update the project list
  renv_scope_locale("LC_COLLATE", "C")
  projlist <- sort(c(projlist, project))
  ensure_parent_directory(projects)
  catchall(writeLines(enc2utf8(projlist), projects, useBytes = TRUE))

  TRUE

}
```

lorenzwalthert:
Closes #837. In particular it solves weird concurrency problems with pre-commit.ci

### DIFF
--- a/R/load.R
+++ b/R/load.R
@@ -253,7 +253,7 @@ renv_load_project <- function(project) {
     projlist <- readLines(projects, warn = FALSE, encoding = "UTF-8")
 
   # if the project is already recorded or cache disabled, nothing to do
-  if (project %in% projlist || !renv::config$cache.enabled())
+  if (project %in% projlist && renv::config$cache.enabled())
     return(TRUE)
 
   # otherwise, update the project list
--- a/R/load.R
+++ b/R/load.R
@@ -252,8 +252,8 @@ renv_load_project <- function(project) {
   if (file.exists(projects))
     projlist <- readLines(projects, warn = FALSE, encoding = "UTF-8")
 
-  # if the project is already recorded, nothing to do
-  if (project %in% projlist)
+  # if the project is already recorded or cache disabled, nothing to do
+  if (project %in% projlist || !renv::config$cache.enabled())
     return(TRUE)
 
   # otherwise, update the project list
--- a/R/load.R
+++ b/R/load.R
@@ -253,7 +253,7 @@ renv_load_project <- function(project) {
     projlist <- readLines(projects, warn = FALSE, encoding = "UTF-8")
 
   # if the project is already recorded or cache disabled, nothing to do
-  if (project %in% projlist && renv::config$cache.enabled())
+  if (project %in% projlist || !renv::config$cache.enabled())
     return(TRUE)
 
   # otherwise, update the project list


kevinushey:
Thanks -- I made a similar change in https://github.com/rstudio/renv/commit/36846af68980bcd18a847b61e470361f7617ba27.

lorenzwalthert:
Thanks Kevin. I am a bit confused about the two ways of cache usage:

```
settings (project-local, specified via R option) -> use.cache
config (R option, fallback to env variable) -> cache.enabled
```

Why are there two ways to disable the cache and both are required for disabling? It was not clear for me from the documentation (help file, search in GitHub issues) how settings and configs are different, and why one can be specified as environment variable and the other not, maybe the help file for `settings` could be a bit more verbose on the high-level concept.

With my change, I was able to specify the behaviour purely over env variables *on the fly*, your change requires also R options (which means I have to write to `.Rprofile`). I can go into details why that is impractical but that's probably off-topic.

kevinushey:
The project setting is used to control whether you want a particular project to use the cache (and only that project), in a way that is "sticky" and preserved in the project. By setting it as a project setting, that value will also apply to anyone else working in that project as well.

The configuration value is supposed to allow you to override the project decision, if (for example) you want to work with a project but need to disable the cache for some reason, but only for yourself.

The global cache will only be enabled if both the project setting + the user config values are on, as per e.g.

https://github.com/rstudio/renv/blob/3f0f23e833999c895e6a1573035374f77153def6/R/cache.R#L518-L520

Setting either of those options should suffice; you shouldn't need to set both.

> With my change, I was able to specify the behaviour purely over env variables on the fly, your change requires also R options (which means I have to write to .Rprofile). I can go into details why that is impractical but that's probably off-topic.

You can set `RENV_CONFIG_CACHE_ENABLED = FALSE` as an environment variable as well.

lorenzwalthert:
Thanks for the explanation. Of course you are right, disabling one option is sufficient, the problem is that `RENV_CONFIG_CACHE_ENABLED` is set in the GitHub master branch of [pre-commit.ci/runner-image](https://github.com/pre-commit-ci/runner-image/commit/595669ca9bdc75685c254ebfa94e91975954088b), but that's not the deployed version, which is why I see the error.

PR #897, opened 2021-12-17
Bugfix/find bioc installer package

deeenes:
Fixed: `renv_lockfile_create` run with a warning if an old `BiocInstaller` installation happened to present in any of the libraries, because `renv_snapshot_filter` added `BiocInstaller` to the lockfile with its `Version` being `NULL`

### DIFF
--- a/R/bioconductor.R
+++ b/R/bioconductor.R
@@ -100,7 +100,7 @@ renv_bioconductor_installer_package <- function(project = NULL) {
 
   bioc_version <- renv_bioconductor_version(project)
 
-  if(is.null(bioc_version)){
+  if (is.null(bioc_version)) {
     old <- utils::compareVersion(getRversion(), "3.5.0") < 0L
   } else {
     old <- utils::compareVersion(bioc_version, "3.8") < 0L
--- a/R/bioconductor.R
+++ b/R/bioconductor.R
@@ -110,7 +110,7 @@ renv_bioconductor_installer_package <- function(project = NULL) {
 
 }
 
-renv_bioconductor_packages <- function(packages, records, project = NULL){
+renv_bioconductor_packages <- function(packages, records, project = NULL) {
   # add in bioconductor infrastructure packages
   # if any other bioconductor packages detected
   sources <- extract_chr(packages, "Source")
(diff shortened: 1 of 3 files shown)


R/bioconductor.R
```diff
@@ -95,3 +95,17 @@ renv_bioconductor_repos_biocinstaller <- function(version) {
   version <- version %||% BiocInstaller$biocVersion()
   BiocInstaller$biocinstallRepos(version = version)
 }
+
+renv_bioconductor_installer_package <- function(project = NULL){
```
kevinushey, line 99:
```suggestion
renv_bioconductor_installer_package <- function(project = NULL) {
```

R/bioconductor.R
```diff
@@ -95,3 +95,17 @@ renv_bioconductor_repos_biocinstaller <- function(version) {
...
+
+  bioc_version <- renv_bioconductor_version(project)
+
+  if(is.null(bioc_version)){
+    old <- utils::compareVersion(getRversion(), "3.5.0") < 0L
+  } else {
+    old <- utils::compareVersion(bioc_version, "3.8") < 0L
+  }
+
+  if(old) "BiocInstaller" else "BiocManager"
```
kevinushey, line 109:
```suggestion
  if (old) "BiocInstaller" else "BiocManager"
```

R/snapshot.R
```diff
@@ -890,17 +882,26 @@ renv_snapshot_filter_packages <- function(project, records, packages) {
   paths <- renv_package_dependencies(packages, project = project)
   all <- as.character(names(paths))
   kept <- keep(records, all)
+  kept <- renv_add_bioc_packages(kept, records, project)
 
+  kept
+
+}
+
+renv_add_bioc_packages <- function(packages, records, project = NULL){
```
kevinushey, line 891: Can you please rename this to e.g. `renv_bioconductor_packages` or something similar? (Anything relevant to Bioconductor functionality should normally be prefixed with `renv_bioconductor_`)

kevinushey (changes requested):

R/snapshot.R
kevinushey: (Or, if you're keeping this in `snapshot.R` then `renv_snapshot_bioconductor_packages`)

kevinushey:
I've made a number of Bioconductor changes that I believe may have obviated this PR. If you're still seeing this issue, can you please provide a reproducible example so I can better understand what's going on?

PR #977, opened 2022-04-08
document `RENV_PATHS_LOCKFILE`

salim-b:
`RENV_PATHS_LOCKFILE` was missing for some reason in the corresponding table of env vars.


kevinushey:
Thanks!

PR #1003, opened 2022-05-04
Fix handling of Rtools 4.2

### LINKED ISSUE #1002: Renv fails to install packages under R 4.2.0 with Rtools42 on Windows
Renv fails to identify the Rtools42 folder when Rtools 4.2 is installed on Windows. If RENV_PATHS_RTOOLS is unset, R BUILD fails to find g++. If RENV_PATHS_RTOOLS is set, another bug prevents this from working since the `renv_paths_rtools` function throws an error.

NikNakk:
Fixes #1002

- Identify Rtools 4.2 when installed
- Update compatibility to indicate that Rtools 4.2 should be used for R >= 4.2
- Set `BINPREF` to `""` when using Rtools 4.2
- Handle RENV_PATHS_RTOOLS environment variable correctly

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: renv
 Type: Package
 Title: Project Environments
-Version: 0.15.4-28
+Version: 0.15.4-29
 Authors@R: c(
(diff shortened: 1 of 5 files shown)


kevinushey (approved):
LGTM; thanks for the PR!

Issue #1025, opened 2022-06-10
renv and ACL's when using ACL'ed cache 


michaelmayer2:
`renv` is doing a great job of retaining ACL's for all file/directory copy and move operations. 

In a situation however where there exists an ACL both on the project folder but also in the global renv cache, any package copied to the cache will inherit the ACL from the project folder but not adopt the one from the cache. 

This then creates issues when different users install the same package. `renv` will then create backup files of previous package installations and reinstall and re-add to the cache again. 

It is not clear what the rationale is to have `renv` being so restrictive with ACL's.

kevinushey:
`renv` can either 'copy' or 'link' packages into / from the cache, depending on whether it's safe for the current library paths to hold symlinks. (We choose not to use symlinks if someone's using `renv` without a project library.)

https://github.com/rstudio/renv/blob/main/R/cache.R#L183-L188

See also:

https://github.com/rstudio/renv/blob/main/R/files.R#L191-L198

Presumedly, `mv` here is also preserving ACLs, so a complete fix here would need to avoid using `mv` if we need to worry about ACLs.

kevinushey:
I'm going to take a closer look at this today.

kevinushey:
I've now made some changes in `renv` in the same line as your PR. In particular, we no longer try to move packages into the cache via `mv`; instead, we always attempt to copy packages (via `cp -R`) so that ACLs are not preserved after copy. Hopefully this is enough to address the customer's concern!

kevinushey:
I've spent some time looking further into this, and I think there's two issues...

1. Packages can be moved into a cache either via `mv` or `cp`. This implies that ACLs (or, in general, permissions) will be preserved when that package is placed into the cache.

2. Even if a package is copied or moved into a directory via `cp` or `mv`, the default ACLs on the cache itself is ignored.

To wit, `renv` now tries to solve this by copying the ACLs set on the cache directory via:

```
getfacl <cache> | setfacl -R --set-file=- <package>
```

Essentially, we try to forcefully reapply the ACLs from the parent directory recursively after the package has been copied into the cache.

This is going to require some testing + feedback to confirm I've got everything right, but hopefully this is the right direction.

kevinushey:
> Happy to help/do any testing as needed.

Are you working with any customers who are bumping into this sort of issue? 
It might be helpful if we could ask them to trial an install of the development version of `renv` to test.

[...]

PR #1026, opened 2022-06-10
lessen restrictive copy

michaelmayer2:
This is a proposal to fix https://github.com/rstudio/renv/issues/1025.


kevinushey:
Rather than change these functions here directly, we should have functions like `renv_cache_move` and `renv_cache_copy` that propagate the appropriate flags down to `renv_file_move` and `renv_file_copy`.

PR #1051, opened 2022-07-25
reference settings directly without using a package reference

aronatkins:


kevinushey:
Thanks! I ended up making this change as part of https://github.com/rstudio/renv/commit/ff15636745873e69840503ce40a38940fedd4de1.

PR #1065, opened 2022-08-05
Support package names that differ from the repository name in GitHub

### LINKED ISSUE #1064: Renv fails when package name differs from repo name
Unlike [remotes](https://remotes.r-lib.org/reference/parse-git-repo.html), renv fails when trying to install a GitHub-hosted package using the `pkgname=user/repo` syntax:

``` r
renv::install("testpkg=tomshafer/example-r-package")
#> Warning: curl: (22) The requested URL returned error: 404
#> Error: failed to resolve remote 'testpkg=tomshafer/example-r-package' -- failed to retrieve 'https://api.github.com/repos/testpkg=tomshafer/example-r-package' [error code 22]
```

Cf. remotes:

``` r
remotes::install_github("testpkg=tomshafer/example-r-package")
#> Using github PAT from envvar GITHUB_PAT
#> Downloading GitHub repo tomshafer/example-r-package@HEAD
#> Running `R CMD build`...
#> * checking for file ‘/private/var/folders/jc/97flnncx6rl3rq_xv_31xwd1d808dq/T/Rtmp5OUxaR/remotese8255bf3c58b/tomshafer-example-r-package-1ad1d48/DESCRIPTION’ ... OK
#> * preparing ‘testpkg’:
#> * checking DESCRIPTION meta-information ... OK
#> * checking for LF line-endings in source and make files and shell scripts
#> * checking for empty or unneeded directories
#> * building ‘testpkg_1.0.0.tar.gz’
#> Installing package into '/Users/tom.shafer/Work/CSC/renv reprex/reprex/renv/library/R-4.2/x86_64-apple-darwin17.0'
#> (as 'lib' is unspecified)
```
<sup>Created on 2022-08-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

The problem compounds when using renv for package development. When my development package Imports a GitHub-hosted package like above via `Remotes:`, renv cannot install _any_ packages. Presumably it looks at my local `DESCRIPTION` each time it tries to do an installation and I receive similar errors -- even though testpkg isn't involved in my install request in any way.

tomshafer:
PR to fix #1064. I can't get the test suite to run on my laptop, so I haven't been able to fully test.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: renv
 Type: Package
 Title: Project Environments
-Version: 0.15.5-49
+Version: 0.15.5-50
 Authors@R: c(
     person("Kevin", "Ushey", role = c("aut", "cre"), email = "kevin@rstudio.com"),
     person("RStudio, PBC", role = c("cph"))
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: renv
 Type: Package
 Title: Project Environments
-Version: 0.15.5-48
+Version: 0.15.5-49
 Authors@R: c(
     person("Kevin", "Ushey", role = c("aut", "cre"), email = "kevin@rstudio.com"),
     person("RStudio, PBC", role = c("cph"))
--- a/R/remotes.R
+++ b/R/remotes.R
@@ -118,8 +118,8 @@ renv_remotes_parse_remote <- function(spec) {
 
   pattern <- paste0(
     "^",
-    "(?:[a-zA-Z.]+=)?",                # optional package name + "="
-    "(?:([^@:]+)(?:@([^:]+))?::)?",    # optional prefix, providing type + host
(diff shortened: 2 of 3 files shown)


kevinushey:
I think this looks correct -- effectively, this tells `renv` to ignore the package name when trying to resolve a remote here (since we don't really need it to find the remote repository, and we can infer the proper package name from the DESCRIPTION file anyhow).

Do you know why this is necessary for `remotes`? It seems like `renv` can work happily by just ignoring the field.

tomshafer:
I don't know for sure, but I expect that what's happening is that `remotes` looks at a package's Imports, etc. and then tries to go and find those packages; the `name=repo` format for the Remotes section makes taht linking possible with custom repo names.

In my case, though, an Import named `example` actually lives at `someuser/example-r-package`, so `remotes` couldn't resolve the dependency until I found this `name=repo` thing.

kevinushey:
Looks like this was implemented in https://github.com/rstudio/renv/commit/e7c288f83927fba23aa33fbefd858aa462d19ac1.

Issue #1156, opened 2023-03-06
renv 0.17.0 issues with provisioning non-CRAN packages


zkamvar:
update: I have determined that the problem lies with how {renv} handles non-CRAN repositories mixed with CRAN repositories. Details in https://github.com/rstudio/renv/issues/1156#issuecomment-1458677629

----

## Problem

If I try to use `renv::install()` with a bioconductor package (e.g. `renv::install("bioc::SummarizedExperiment")` (related to https://github.com/rstudio/renv/issues/1110), the process fails:

Note: I am using the most recent GitHub version of {renv}: https://github.com/rstudio/renv/tree/0.17.0-16

<details>
<summary>(to avoid having to copy/paste/delete, here is the reprex code to generate the output)</summary>

```r
reprex::reprex({ 
    tmp <- tempfile() 
    dir.create(tmp) 
    writeLines("library(SummarizedExperiment)", file.path(tmp, "test.R")) 
    setwd(tmp) 
    renv::init() 
    renv::install("bioc::SummarizedExperiment") 
  }, std_out_err = TRUE, html_preview = FALSE, si = TRUE)
```

</details>

``` r
tmp <- tempfile()
dir.create(tmp)
writeLines("library(SummarizedExperiment)", file.path(tmp, "test.R"))
setwd(tmp)
renv::init()
#> * Discovering package dependencies ... Done!
#> * Resolving missing dependencies  ...
#> Warning in value[[3L]](cond): restarting interrupted promise evaluation
#> The version of R recorded in the lockfile will be updated:
#> - R      [* -> 4.2.2]
#> 
#> * Lockfile written to '/tmp/RtmpZBW61e/file1222c1c0355c7/renv.lock'.
renv::install("bioc::SummarizedExperiment")
#> Warning in value[[3L]](cond): restarting interrupted promise evaluation
#> Error in object[[i]]: object of type 'promise' is not subsettable
```

<sup>Created on 2023-03-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Standard output and standard error
</summary>

``` sh
The following package(s) were not installed successfully:

	[SummarizedExperiment]: object of type 'promise' is not subsettable

You may need to manually download and install these packages.

The following package(s) will be updated in the lockfile:

# GitHub =============================
- renv   [* -> rstudio/renv@HEAD]

Warning message:
In in_dir(input_dir(), expr) :
[...]

PR #1157, opened 2023-03-07
fix typos in stopf and warningf

zkamvar:
stopf and warningf had the patterns of `call. = call.`, which was likely a typo. This will address #1156


zkamvar:
Sorry, that was a mistake. My bad -_-

Issue #1161, opened 2023-03-08
{renv} 0.17.0 cannot find rmarkdown on hydrate with bare project on CI


zkamvar:
I'm having a problem with {sandpaper}, which uses {renv} to provision a non-invasive dependencies for lessons. 

I'm finding an odd situation: on Continuous integration, and ONLY on R > 4.2 and < 4.0, [{renv} can not detect {rmarkdown} during bootstrap.](https://github.com/carpentries/sandpaper/actions/runs/4367134933/jobs/7638060723#step:13:1708) 

During bootstrap, I do the following steps:

1. [initialise the project containing R Markdown files with a bare renv with a named profile](https://github.com/carpentries/sandpaper/blob/3bb22eff927a1fbe1017000f3ba8c2d91244b89a/R/utils-renv.R#L119-L142)
2. [run hydrate to capture all packages in the project](https://github.com/carpentries/sandpaper/blob/3bb22eff927a1fbe1017000f3ba8c2d91244b89a/R/utils-renv.R#L257-L258)
3. snapshot

However, with the latest version of {renv}, I've noticed problems with the bootstrap hydration phase:

```
  * Discovering package dependencies ... Done!
  The following packages are used in this project, but not available locally:
  
  	rmarkdown
  
  renv will attempt to download and install these packages.
  
  * Hydrated 0 packages in 0.005 milliseconds.
```

But when I run `renv::diagnostics()`, {rmarkdown} is available locally:

<details>
<summary>diagnostics report</summary>

````
Diagnostics Report [renv 0.17.0-27]
  ===================================
  
  # Session Info =======================
  R version 4.2.2 (2022-10-31)
  Platform: x86_64-pc-linux-gnu (64-bit)
  Running under: Ubuntu 22.04.2 LTS
  
  Matrix products: default
  BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
  LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.20.so
  
  locale:
   [1] LC_CTYPE=C.UTF-8       LC_NUMERIC=C           LC_TIME=C.UTF-8       
   [4] LC_COLLATE=C           LC_MONETARY=C.UTF-8    LC_MESSAGES=C.UTF-8   
   [7] LC_PAPER=C.UTF-8       LC_NAME=C              LC_ADDRESS=C          
[...]

PR #1162, opened 2023-03-09
prevent "NULL" user libpaths

zkamvar:
In R CMD check for R >= 4.2, the environment variable R_LIBS_USER somehow gets set to the character NULL.

This is a bandaid for this. See https://github.com/rstudio/renv/issues/1161#issuecomment-1461092077


kevinushey:
Thank you!

PR #1207, opened 2023-04-25
Tweaks to CI vignette

hadley:
* Make it more clear why you'd want to use renv on CI, and why it's not a good idea for CRAN packages
* Flip order of renv approaches (since IMO r-lib/setup-renv is easiest for most people)
* Remove travis, since few people in the R community are using it any more.


kevinushey (approved):
Thanks!

Issue #413, opened 2020-05-01
A way to "check out repository versions"


colearendt:
The use case here is that (even using `renv`), the project library can be an unexpected mashup of random GitHub versions, CRAN versions, etc. of packages.

I would like a way to `renv::checkout_repo()` or something where all of my packages are either (1) have the latest version retrieved from my repository or (2) stay the way they are (i.e. if not found on CRAN) and are notified about.

This is especially useful when using frozen repositories (i.e. RSPM) and checking out a point in time, or when trying to emulate the build environment for CI (which typically installs the latest from a repository into a clean environment). `renv::update()` seems to get close here, but it reaches out often to git repositories.

One noteworthy point here is that I want to be careful to (1) use the `renv` cache if I have the version already, (2) populate the `renv` cache if I don't.

kevinushey:
Mini-sketch for this function would basically be like:

```
checkout <- function(repos) {
    renv_scope_options(repos = repos)
    deps <- renv::dependencies()
    pkgs <- unique(deps$Package)
    renv::install(pkgs)
}
```

hadley:
Note interface discussion in https://github.com/rstudio/renv/issues/507#issuecomment-1527933691

PR #1219, opened 2023-05-02
implement renv::checkout()

kevinushey:
Closes https://github.com/rstudio/renv/issues/413.

### DIFF
--- a/R/checkout.R
+++ b/R/checkout.R
@@ -11,6 +11,14 @@
 #' packages used in a particular `renv` project to the package versions
 #' provided by a particular snapshot.
 #'
+#' If your library contains packages installed from other remote sources (e.g.
+#' GitHub), but a version of a package of the same name is provided by the
+#' repositories being checked out, then please be aware that the package will be
+#' replaced with the version provided by the requested repositories. This could
+#' be a concern if your project uses \R packages from GitHub whose name matches
+#' that of an existing CRAN package, but is otherwise unrelated to the package
+#' on CRAN.
+#'
 #' @inheritParams renv-params
 #'
 #' @param repos The \R package repositories to use.
--- a/R/checkout.R
+++ b/R/checkout.R
@@ -2,21 +2,48 @@
 #' Checkout a Repository
 #'
 #' `renv::checkout()` can be used to install the latest packages available from
-#' the requested repositories.
+#' a (set of) package repositories.
+#'
+#' `renv::checkout()` is most useful with services like the Posit's
+#' [Package Manager](https://packagemanager.rstudio.com/), as it
+#' can be used to switch between different repository snapshots within an
+#' `renv` project. In this way, you can upgrade (or downgrade) all of the
+#' packages used in a particular `renv` project to the package versions
+#' provided by a particular snapshot.
 #'
 #' @inheritParams renv-params
 #'
-#' @param repos The \R package repositories to check out. When `NULL` (the default),
-#'   `renv` will attempt to determine the packages used in the project via
-#'   a call to [renv::dependencies()].
+#' @param repos The \R package repositories to check out.
 #'
 #' @param packages The packages to be installed. When `NULL` (the default),
-#'   all packages currently used in the project will be installed.
+#'   all packages currently used in the project will be installed, as
+#'   determined by [renv::dependencies()]. The recursive dependencies of these
+#'   packages will be included as well.
 #'
-#' @keywords internal
-checkout <- function(repos = getOption("repos"),
+#' @param date The snapshot date to use. When set, the associated snapshot as
+#'   available from the Posit's public
+#'   [Package Manager](https://packagemanager.rstudio.com/) instance will be
+#'   used. The repository path can be adjusted using the `renv.checkout.repos`
+#'   option. Ignored if `repos` is non-`NULL`.
+#'
+#' @examples
+#' \dontrun{
+#'
+#' # check out packages from PPM using the date '2023-01-02'
+#' renv::checkout(date = "2023-01-02")
+#'
+#' # alternatively, supply the full repository path
+#' renv::checkout(repos = "https://packagemanager.rstudio.com/cran/2023-01-02")
+#'
+#' # only check out some subset of packages (and their recursive dependencies)
+#' renv::checkout(packages = "dplyr", date = "2023-01-02")
+#'
+#' }
+#' @export
+checkout <- function(repos = NULL,
                      ...,
                      packages = NULL,
+                     date     = NULL,
                      clean    = FALSE,
                      project  = NULL)
 {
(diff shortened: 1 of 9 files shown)


R/checkout.R
```diff
@@ -2,21 +2,48 @@
...
-#'   a call to [renv::dependencies()].
+#' @param repos The \R package repositories to check out.
 #'
 #' @param packages The packages to be installed. When `NULL` (the default),
-#'   all packages currently used in the project will be installed.
+#'   all packages currently used in the project will be installed, as
+#'   determined by [renv::dependencies()]. The recursive dependencies of these
+#'   packages will be included as well.
+#'
+#' @param date The snapshot date to use. When set, the associated snapshot as
```
hadley, line 23: I think ideally we'd recognise the an existing repo named `RPSM` either in `getOptions("repos")` or the `repos` argument.

R/checkout.R
```diff
@@ -2,21 +2,48 @@
...
+#' renv::checkout(packages = "dplyr", date = "2023-01-02")
 #'
-#' @keywords internal
-checkout <- function(repos = getOption("repos"),
+#' }
+#' @export
+checkout <- function(repos = NULL,
                      ...,
                      packages = NULL,
+                     date     = NULL,
```
hadley, line 46: Given it's importance, I think `date` should come earlier in the arguments.

R/checkout.R
```diff
@@ -2,21 +2,48 @@
 #' Checkout a Repository
 #'
 #' `renv::checkout()` can be used to install the latest packages available from
-#' the requested repositories.
+#' a (set of) package repositories.
```
hadley, line 5: Maybe something like: It takes an renv library that tracks the idiosyncratic install of packages over time and gets it in a consistent state, as if you had installed all your packages on the same day.

I think it's also worth comparing to `update()`, and mentioning `checkout()` in the `update()` docs.

R/checkout.R
```diff
@@ -2,21 +2,48 @@
...
+#' `renv` project. In this way, you can upgrade (or downgrade) all of the
+#' packages used in a particular `renv` project to the package versions
+#' provided by a particular snapshot.
 #'
 #' @inheritParams renv-params
 #'
-#' @param repos The \R package repositories to check out. When `NULL` (the default),
-#'   `renv` will attempt to determine the packages used in the project via
-#'   a call to [renv::dependencies()].
+#' @param repos The \R package repositories to check out.
```
hadley, line 16:
```suggestion
#' @param repos The \R package repositories to use.
```

?

R/checkout.R
```diff
@@ -28,64 +55,142 @@ checkout <- function(repos = getOption("repos"),
...
+  dbs <- available_packages(type = "source")
+  if (is.null(dbs))
+    stop("no package repositories are available")
 
-    # ignore base packages (these are never distributed on CRAN)
-    priority <- renv_package_priority(package)
-    if (identical(priority, "base"))
-      return(NULL)
+  db <- renv_available_packages_flatten(dbs)
+  packages <- intersect(packages, db$Package)
```
hadley, line 94: I think it would be useful to message `setdiff(db$Package, packages)` so it's clear what's not updated.

R/checkout.R
```diff
@@ -2,21 +2,48 @@
...
 #' `renv::checkout()` can be used to install the latest packages available from
-#' the requested repositories.
+#' a (set of) package repositories.
+#'
+#' `renv::checkout()` is most useful with services like the Posit's
+#' [Package Manager](https://packagemanager.rstudio.com/), as it
+#' can be used to switch between different repository snapshots within an
+#' `renv` project. In this way, you can upgrade (or downgrade) all of the
+#' packages used in a particular `renv` project to the package versions
+#' provided by a particular snapshot.
```
hadley, line 12: Mention that it will also convert dev versions installed from Github to released versions? Or maybe in general that it matches the names of currently used packages to available packages from the specified repos?

R/checkout.R
```diff
@@ -28,64 +55,142 @@ checkout <- function(repos = getOption("repos"),
...
+  # create a lockfile matching this request
+  lockfile <- renv_lockfile_init(project)
+  lockfile$Packages <- records
+
+  # restore from that lockfile
+  restore(lockfile = lockfile, clean = clean)
+
+}
+
+renv_checkout_packages <- function(project) {
```
hadley, line 81: This surprised me — I had expected that you take the package names from the existing lock file.

R/checkout.R
kevinushey: Sorry, I don't understand -- can you elaborate?

R/checkout.R
kevinushey: Makes sense -- I use `...` here to force arguments to be named; should I relax that requirement (remove the usage of `...`)?

R/checkout.R
kevinushey: FWIW you do get a summary of the changes; e.g. if `pak` integration is enabled:

```
> renv::checkout(date = "2022-01-02")
* Snapshot date '2022-01-02' not available; using '2021-12-29' instead
* Querying repositories for available source packages ... Done!
✔ Updated metadata database: 2.73 MB in 5 files.                                              
ℹ R 4.3 aarch64-apple-darwin20 packages are missing from Bioconductor                          
✔ Updating metadata database ... done                                                         
                                                                                                
→ Will install 12 packages.
→ Will update 121 packages.
→ Will download 5 CRAN packages (4.69 MB), cached: 128 (0 B).
+ assertthat                0.2.1    ⬇ (53.25 kB)
+ BiocManager    1.30.20  → 1.30.16 👷🏽
+ bit            4.0.5    → 4.0.4   👷🏿‍♀️🔧
+ blob           1.2.4    → 1.2.2   👷🏿‍♂️
+ boot           1.3-28.1 → 1.3-28  👷🏼‍♀️
+ brew           1.0-8    → 1.0-6   👷🏿‍♀️
+ broom          1.0.4    → 0.7.10  👷🏻
< ... >
```

Do you think that's sufficient? Or should we explicitly call out packages which are installed from GitHub (or other remote sources) but don't appear to be available from the package repositories?

R/checkout.R
kevinushey: I don't think we want to use the lockfile -- the dependency tree is likely to change across different snapshots (e.g. IIRC `dplyr` only recently-ish took a dependency on `vctrs`) so we basically need to re-compute the dependency tree based on the packages used in the project.

In a sense, `renv::checkout()` is really like `renv::init()`, except instead of copying packages from the user / default library paths, we're copying them from a package repository.

R/checkout.R
kevinushey: FWIW `date` is really just a shortcut so you can just use `renv::checkout(date = "2022-01-02")` instead of `renv::checkout(repos = "https://packagemanager.rstudio.com/cran/2022-01-02")` or something similar.

R/checkout.R
```diff
@@ -28,64 +55,142 @@ checkout <- function(repos = getOption("repos"),
   renv_project_lock(project = project)
 
   # set new repositories
+  repos <- repos %??% renv_checkout_repos(date)
   options(repos = repos)
 
+  # TODO: Activate Bioconductor if it appears to be used by this project
```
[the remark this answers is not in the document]
kevinushey: Are there any other special considerations here for Bioconductor? (Could checking out an older snapshot also imply the need for an older Bioconductor release?) Feels like there could be some awkward conflicts.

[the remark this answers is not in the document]
hadley: Yeah, I'm not sure how this should work for BioC given that it's mostly tied to a specific R version, not a point in time. Maybe a separate BioC argument that at lets you get them consistent with a given BioC version? It would default to the version corresponding to the current version of R?

R/checkout.R
hadley: Oh, I hadn't thought about the dependency problem. Thanks for explaining!

R/checkout.R
hadley: Maybe there are really two commands here — set up my repos (with convenience syntax for setting dates) and re-install all packages from those repos (reporting where you can't).

kevinushey: The potential for users to have multiple PPM instances in their repositories makes me think we should do away with the `date` parameter, and ask users to just provide the full repository URL. This lets users take charge of what copy of PPM they're using, or whether they need a "mix" of snapshots from two different PPM instances for some reason.

The downside is (if you have scripts using `renv::checkout()`), if you need to change the repository URL in the future, you need to change all of those `renv::checkout()` calls to use the updated repository. Not sure if that's a real concern.

Otherwise, we might want to have something like:

- A project setting giving the "root" URL for the PPM instance,
- (Maybe?) A project setting giving the repository name to be used from the PPM instance (e.g. "cran"),
- `renv::checkout(date = <...>)` then makes use of the project settings when forming the full repository URL.

Or maybe (similar to Hadley's suggestion) we have a project setting giving a map from repository names to PPM "root" URLs, e.g.

```
{
    "RSPM": "https://pm.company.local/cran",
    "LOCAL": "https://pm.company.local/local"
}
```

and then the requested date would be applied to all "known" PPM URLs that have been registered in called to `renv::checkout(date = <...>)`. The only requirement then would be that the names of the `getOption("repos")` option would match the names in the project setting.

aronatkins: Some folks might have:

```r
options(repos = c(
    CRAN = "https://pm-dev.company.local/cran/2023-01-01"
))
```

and then when they deploy their content into some production environment, they'll want us to use:

```r
options(repos = c(
    CRAN = "https://pm-prod.company.com/cran/XXX"
))
```

This transition is not the same as "checkout", though. In this situation, folks want to use the same package graph but obtain those packages from a different repository URL.

Package Manager snapshot URLs (either date-stamps or transaction ids) are not valid across servers (except in the very narrow case where the repository is a simple, direct mirror of a single source like CRAN). Many installations use "approved CRAN" configurations. Rewriting from "dev" to "prod" essentially means that you need to drop the snapshot reference, and then you're asking for package binary incompatibility..

I really like the idea of "update my library to use 'current' versions" for whatever is the active set of repositories.

aronatkins: Aside: We should discourage `RSPM="https://pm.company.local/cran"` because `RSPM` is the server identifier included by RSPM in the `Repository: RSPM` package `DESCRIPTION` entry. It is unfortunately identifying the kind of the server, not the name of the repository.

hadley: Coming all the way back around, maybe it's still useful to have `date` create a repo using PPPM — that makes it easy for folks who can use PPPM, while not making it any harder for the folks with internal repos that need to supply the full url.

PR #1222, opened 2023-05-03
Reduce snapshot prompting

### LINKED ISSUE #1218: Snapshot shouldn't prompt to activate
Because using it without activation is a reasonable workflow, if you just want to capture the package dependencies you currently have installed.

hadley:
* Don't prompt to activate
* Ony prompt if overwriting existing file

Fixes #1218


kevinushey:
I wonder if instead we should have `renv_activate_prompt()` only perform prompting on snapshot if the working directory contains an `renv/library` folder? (In other words, if it looks like the user is working within a project with existing `renv` infrastructure, but that project hasn't been activated for some reason.)

https://github.com/rstudio/renv/blob/63c94fd5e9a6465d2a6328514ca29560dcf3f488/R/activate.R#L138-L144

hadley:
I think that makes sense — overwriting an existing lockfile is only risky if you're using a project library, because you then lose the ability to roll back any changes you have made. If you don't have a `renv/library`, there's no risk to snapshotting because you're always just capturing your current state.

kevinushey:
Right, but I wonder if we should still prompt to activate if the user calls `renv::snapshot()` within an `renv` project that isn't currently loaded?

In other words, keep the existing code here:

https://github.com/rstudio/renv/pull/1222/files?w=1#diff-5d849f3782ddfdd6b0053b841b4be729f4fdaa33c9262caf6757aa185e5f5658L138

but instead only prompt on snapshot if the `renv/library` directory exists, but that project isn't loaded:

https://github.com/rstudio/renv/blob/0ae3cba3edc7fa69a42119884363038762d0239b/R/activate.R#L139-L145

That is, add an extra check for `renv/library` if `action == "snapshot"`.

kevinushey:
I made the change I thought was appropriate here:

https://github.com/rstudio/renv/commit/a6c88c00f38168f63dbbc12371525af7906ecd98

Does that look okay?

hadley:
With that commit, I see:

```R
 create_project("~/Desktop/test-renv", open = FALSE)
#> ...
snapshot("~/Desktop/test-renv/")
#> The following package(s) will be updated in the lockfile:
#> 
#> # CRAN ===============================
#> - renv   [* -> 0.17.3-41]
#> 
#> The version of R recorded in the lockfile will be updated:
#> - R      [* -> 4.2.3]
#> 
#> Do you want to proceed? [Y/n]: 
```

So the prompt to activate is gone, which is good, but we still get prompted to "update" a lockfile that doesn't otherwise exist.

kevinushey:
Would you prefer some alternate text there? E.g. "The following packages will be recorded in the lockfile:"?

That is, whether we say "recorded" / "updated" depends on whether `renv.lock` existed before?

PR #1228, opened 2023-05-03
Tweak wording of activate prompt

hadley:
I think this makes it more clear what happens if you say no.

### DIFF
--- a/R/activate.R
+++ b/R/activate.R
@@ -149,16 +149,15 @@ renv_activate_prompt <- function(action, library, prompt, project) {
   fmt <- lines(
     "",
     "This project has not yet been activated.",
-    "Activating this project will ensure the project library is used when %s() is called.",
-    "Please see `?renv::activate` for more details.",
+    "This means that %s() will modify your global library.",
+    "Activate to use the project library.",
+    "See `?renv::activate` for more details.",
     ""
   )
-
   notice <- sprintf(fmt, action)
   vwritef(notice)
 
-  fmt <- "Would you like to activate this project before %s() is called?"
-  question <- sprintf(fmt, action)
+  question <- "Would you like to activate the project now?"
   response <- ask(question, default = TRUE)
   if (!response)
     return(FALSE)


R/activate.R
```diff
@@ -149,16 +149,15 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+    "See `?renv::activate` for more details.",
     ""
   )
-
   notice <- sprintf(fmt, action)
   vwritef(notice)
 
-  fmt <- "Would you like to activate this project before %s() is called?"
-  question <- sprintf(fmt, action)
+  question <- "Would you like to activate the project now?"
```
kevinushey, line 160:
```suggestion
  question <- "Would you like to activate this project now?"
```

R/activate.R
```diff
@@ -149,16 +149,15 @@ renv_activate_prompt <- function(action, library, prompt, project) {
   fmt <- lines(
     "",
     "This project has not yet been activated.",
-    "Activating this project will ensure the project library is used when %s() is called.",
-    "Please see `?renv::activate` for more details.",
+    "This means that %s() will modify your global library.",
```
kevinushey, line 152: This line would only be true if this were called in response to restore / clean; if we kept the snapshot prompt then it wouldn't be accurate (so I guess this PR depends on https://github.com/rstudio/renv/pull/1222).

kevinushey (commented):
I feel like we can still find some better text to use here. We might want to just have different tailored messages for the different activation scenarios (restore / clean / ...)

hadley:
Alternatively, maybe it would be better to error if the user declines to activate? It doesn't seem like we want folks using `restore()` to update the system libraries?

hadley:
If we merge #1243, the default prompt will become yes, we could easily make error if the user declines interatively and succeed if run non-interactively.  (But I do think it makes renv harder to understand if actions sometimes change the system library depending on what else is going on.)

hadley:
I've had another stab at this with all the recent changes, but I think the problem is that a Y/N just isn't quite enough. I think the message should really be something like:

```
This project has not yet been activated.
What do you want to do?

1. Activate the project and set up a project library.
2. Continue and use the global library
3. Cancel
```

kevinushey:
I like this idea as well -- do you want to file a separate issue to capture this, or just submit a new PR?

hadley:
I'll just make a new PR.

PR #1232, opened 2023-05-04
Vignette reorg

hadley:
I think this is at a good place to take a look — I haven't removed any content, just shuffled it around to new places. I've only added a small amount of content, as placeholders in `renv.Rmd` to point to the more detailed coverage in other vignettes.

Here's what the articles drop down looks like now:
Screenshot 2023-05-04 at 08 24 06


kevinushey:
Main nit: we should be consistent with using Capital Case versus Sentence case in the vignette titles.

kevinushey:
Thanks!

PR #1233, opened 2023-05-04
Build & deploy pkgdown site automatically

hadley:
I also used the tidytemplate template, which despite the name, is something we use for many non-tidyverse Posit packages. It's a pretty minor variation from the default, but I do think it looks little cleaner.


kevinushey (commented):
LGTM!

PR #1243, opened 2023-05-10
Refactor prompting

### LINKED ISSUE #1240: Change prompt default to yes
In `snapshot()` and `restore()`. And review other cases.

hadley:
The major user facing is change the default value of `default` for `proceed()` from `FALSE` to `TRUE`, so the interactive default is equivalent to the non-interactive behaviour.

I then refactored `proceed()` to include a `prompt` argument: I think this makes it more clear what happens when `prompt` is `FALSE`.

While updating the code to use the new `proceed()` behaviour I realised that there many repeated calls that aborted if the user said `n` at the prompt. So I introduced a new `check_can_proceed()` function that extracts out the repeated code.

And then while updating the code to use `check_can_proceed()` I realised that (unsurprisingly) most calls to `check_can_proceed()` came up after a message that was gated by a check for `prompt || renv_verbose()`. So I also added a `prompt` argument to `renv_verbose()` because I think this makes the connection between the two functions more clear.

Fixes #1240

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -167,10 +167,20 @@ ask <- function(question, default = FALSE) {
 
 }
 
-proceed <- function(default = FALSE) {
-  ask("Do you want to proceed?", default = default)
+proceed <- function(prompt = interactive(), default = TRUE) {
+  if (prompt) {
+    ask("Do you want to proceed?", default = default)
+  } else {
+    default
+  }
 }
 
+check_can_proceed <- function(prompt = interactive()) {
+  if (!proceed(prompt)) {
+    renv_report_user_cancel()
+    invokeRestart("abort")
+  }
+}
 # nocov end
 
 inject <- function(contents,
--- a/R/verbose.R
+++ b/R/verbose.R
@@ -1,5 +1,5 @@
 
-renv_verbose <- function() {
+renv_verbose <- function(prompt = interactive()) {
 
   verbose <- getOption("renv.verbose")
   if (!is.null(verbose))
@@ -9,6 +9,6 @@ renv_verbose <- function() {
   if (!is.na(verbose))
     return(as.logical(verbose))
 
-  interactive() || !renv_tests_running()
+  prompt || !renv_tests_running()
 
 }
--- a/R/clean.R
+++ b/R/clean.R
@@ -130,7 +130,7 @@ renv_clean_library_tempdirs <- function(project, prompt) {
     return(ntd())
 
   # nocov start
-  if (prompt || renv_verbose()) {
+  if (renv_verbose(prompt)) {
 
     renv_pretty_print(
       bad,
@@ -138,7 +138,7 @@ renv_clean_library_tempdirs <- function(project, prompt) {
       wrap = FALSE
     )
 
-    if (prompt && !proceed())
+    if (!proceed(prompt))
       return(FALSE)
 
   }
@@ -180,7 +180,7 @@ renv_clean_system_library <- function(project, prompt) {
     return(ntd())
 
   # nocov start
-  if (prompt || renv_verbose()) {
+  if (renv_verbose(prompt)) {
 
     renv_pretty_print(
       packages,
@@ -192,7 +192,7 @@ renv_clean_system_library <- function(project, prompt) {
       )
     )
 
-    if (prompt && !proceed())
+    if (!proceed(prompt))
       return(FALSE)
 
   }
@@ -227,7 +227,7 @@ renv_clean_unused_packages <- function(project, prompt) {
     return(ntd())
 
   # nocov start
-  if (prompt || renv_verbose()) {
+  if (renv_verbose(prompt)) {
 
     renv_pretty_print(
       removable,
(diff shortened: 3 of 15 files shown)


hadley:
LMK if you want me to add more tests, or if you want me to refactor `proceed()`/`ask()` to be testable (I have some existing work I can adapt from rsconnect).

kevinushey (commented):
I don't like this change; the intention in these code paths is that (in non-interactive sessions) you could set `options(renv.verbose = TRUE)` to get a textual summary of the changes `renv` will make (as the changes are executed), but not be prompted whether you need to proceed.

kevinushey:
Note also that `ask()` already assumes `TRUE` for these cases when running tests: https://github.com/rstudio/renv/blob/40aa593527e828fe392df24f566427ba04d2961d/R/utils.R#L119-L120

hadley:
> I don't like this change; the intention in these code paths is that (in non-interactive sessions) you could set options(renv.verbose = TRUE) to get a textual summary of the changes renv will make (as the changes are executed), but not be prompted whether you need to proceed.

As far as I know I didn't change that behaviour? You should see the summaries if `prompt` is `TRUE` or the user has done `options(renv.verbose = TRUE)`.

R/utils.R
```diff
@@ -167,10 +167,20 @@ ask <- function(question, default = FALSE) {
...
-  ask("Do you want to proceed?", default = default)
+proceed <- function(prompt = interactive(), default = TRUE) {
+  if (prompt) {
+    ask("Do you want to proceed?", default = default)
+  } else {
+    default
+  }
 }
 
+check_can_proceed <- function(prompt = interactive()) {
```
kevinushey, line 178: How do you feel about a different name; e.g.

```
halt(<value>)
cancel(<value>)
halt_if(<value>)
cancel_if(<value>)
```

R/utils.R
```diff
@@ -167,10 +167,20 @@ ask <- function(question, default = FALSE) {
 
 }
 
-proceed <- function(default = FALSE) {
-  ask("Do you want to proceed?", default = default)
+proceed <- function(prompt = interactive(), default = TRUE) {
+  if (prompt) {
```
kevinushey, line 171: What is the behavior if the user sets `prompt = TRUE` in a non-interactive session?

R/verbose.R
```diff
@@ -1,5 +1,5 @@
 
-renv_verbose <- function() {
+renv_verbose <- function(prompt = interactive()) {
```
kevinushey, line 2: Maybe my concern here is that "should we display some text" and "has the user asked for verbose output" are separate questions in my mind. Maybe we need a separate function here that collects those conditions, e.g.

```
renv_notification_required <- function(prompt) {
    prompt || renv_verbose()
}
```

or something like that?

kevinushey:
> As far as I know I didn't change that behaviour? You should see the summaries if prompt is TRUE or the user has done options(renv.verbose = TRUE).

Sorry, you are right! I misread the code. I still find this version harder to parse than the old code though. If the question is "do I want to display text to the user?" then the answer being "either if we're verbose, or if we're going to prompt the user" seems clearer as two separate checks, rather than wrapped into `renv_verbose()`.

R/verbose.R
hadley: Maybe `renv_needs_notification()`? I do think the two concerns are linked because you always want to notify the user when the prompt is true.

R/utils.R
hadley: I'm happy to change to whatever you want; I just used `check_`, because that prefix is something we use reasonably commonly in r-lib/tidyverse to indicate that you can ignore the return value, because the main point of the function is to signal an error if needed. For example, `check_installed("pkg")` errors if pkg isn't installed, and `check_string(x)` errors if `x` isn't a string.

R/verbose.R
hadley: `renv_should_notify()`?

R/utils.R
hadley: Hmmm, `ask()` has another layer of fallbacks, so it'll return `default`.

R/verbose.R
hadley: BTW why is it `can_proceed()` but `renv_verbose()`? i.e. why does one get the `renv_` prefix?

kevinushey: The rough convention I have: by default, internal functions should have names of the form `renv_<scope>_<name>`, unless they're generic enough to be useful in a variety of contexts (e.g. the stuff in `utils.R`).

`renv_verbose` probably could have a different name though.

kevinushey:
I'm trying to think more about the `prompt || renv_verbose()` pattern. I think the main intention here is:

- In interactive sessions, if we're prompting the user, we need to make sure that we display some report to the user before prompting them, so the 'verbose' option should effectively be ignored (we always display output). In other words, if you run `options(renv.verbose = FALSE); renv::snapshot(prompt = TRUE)`, then you should still get the lockfile diff report before being asked to snapshot.

- In non-interactive sessions, some users might decide they want to just turn off any of the reporting done by `renv`, so `options(renv.verbose = FALSE)` is a tool for that.

- When running tests, I try to disable most "verbose" output just to reduce the amount of chatter.

In other words, I think `renv.verbose` is mainly relevant for non-interactive sessions, for users who want `renv` to be quieter for some reason?

PR #1244, opened 2023-05-11
Refactor project get/set

### LINKED ISSUE #1058: Feature Request: function that tests if renv project is loaded
Use-case:

When opening an .Rproj that uses renv, a message is printed like:
```r
* Project '/path/to/project/' loaded. [renv 0.15.5]
```
I can manually inspect the console output for that message, but it would be nice to add a line to my code that tests that it was successfully loaded. This would be particularly useful for cases where I'm running scripts in batch mode. Then my scripts could include something like:
```r
stopifnot(renv_is_loaded)
```
This came up because I noticed that I was running scripts in batch mode from the wrong directory, so renv was not being sourced properly.

hadley:
* Eliminate `renv_project()` in favour of `renv_project_get()`.

* Use `renv_project_resolve()` in `load()`, like all other functions. Move `renv_project_find()` to `autoload.R`, the only place it's now used.

* Use `renv_project_set()`, `renv_project_get()`, and `renv_project_clear()` everywhere, instead of accessing `RENV_PROJECT` env var directly.

*  Implement `renv_project_is_active()` and use where appropriate. Plan is to export as `is_loaded()` to fix #1058.

* I don't understand how `renv_snapshot_task()` worked because, AFAICT, `project` wasn't previously defined.

### DIFF
--- a/R/activate.R
+++ b/R/activate.R
@@ -142,7 +142,7 @@ renv_activate_prompt <- function(action, library, prompt, project) {
     prompt &&
     interactive() &&
     is.null(library) &&
-    !identical(project, Sys.getenv("RENV_PROJECT"))
+    !renv_project_is_active(project)
 
   if (!ask)
     return(FALSE)
--- a/R/autoload.R
+++ b/R/autoload.R
@@ -81,3 +81,24 @@ renv_autoload_impl <- function() {
   TRUE
 
 }
+
+# TODO: this gets really dicey once the user starts configuring where
+# renv places its project-local state ...
+renv_project_find <- function(project = NULL) {
+
+  project <- project %||% getwd()
+
+  anchors <- c("renv.lock", "renv/activate.R")
+  resolved <- renv_file_find(project, function(parent) {
+    for (anchor in anchors)
+      if (file.exists(file.path(parent, anchor)))
+        return(parent)
+  })
+
+  if (is.null(resolved)) {
+    fmt <- "couldn't resolve renv project associated with path %s"
+    stopf(fmt, renv_path_pretty(project))
+  }
+
+  resolved
+}
--- a/R/load.R
+++ b/R/load.R
@@ -29,11 +29,7 @@ load <- function(project = NULL, quiet = FALSE) {
 
   renv_scope_error_handler()
 
-  project <- normalizePath(
-    project %||% renv_project_find(project),
-    winslash = "/",
-    mustWork = TRUE
-  )
+  project <- renv_project_resolve(project)
 
   action <- renv_load_action(project)
   if (action[[1L]] == "cancel") {
@@ -64,8 +60,7 @@ load <- function(project = NULL, quiet = FALSE) {
   # then unload the current project and reload the requested one
   switch <-
     !renv_metadata_embedded() &&
-    !is.na(Sys.getenv("RENV_PROJECT", unset = NA)) &&
-    !identical(project, renv_project())
+    !renv_project_is_active(project)
 
   if (switch)
     return(renv_load_switch(project))
@@ -331,7 +326,7 @@ renv_load_project <- function(project) {
 
   # record the active project in this session
   project <- renv_path_normalize(project, winslash = "/")
-  Sys.setenv(RENV_PROJECT = project)
+  renv_project_set(project)
 
   # update project list if enabled
   enabled <- renv_cache_config_enabled(project = project)
--- a/R/modify.R
+++ b/R/modify.R
@@ -38,7 +38,7 @@ renv_modify_impl <- function(project, changes) {
   else
     renv_modify_noninteractive(project, changes)
 
-  if (identical(renv_project(), project))
+  if (renv_project_is_active(project))
     renv_modify_fini(lockfile)
 
 }
--- a/R/project.R
+++ b/R/project.R
@@ -18,17 +18,24 @@
 #'
 #' }
 project <- function(default = NULL) {
-  renv_project(default = default)
+  renv_project_get(default = default)
 }
 
-renv_project_get <- function(default = NULL) {
-
+renv_project_get <- function(default = getwd()) {
   project <- Sys.getenv("RENV_PROJECT", unset = NA)
   if (is.na(project))
     return(default)
 
   project
+}
 
+renv_project_resolve <- function(project = NULL) {
+  project <- project %||% renv_project_get()
+  normalizePath(project, winslash = "/", mustWork = FALSE)
+}
+
+renv_project_is_active <- function(project = NULL) {
+  identical(renv_project_resolve(project), renv_project_get(default = NULL))
 }
 
 renv_project_set <- function(project) {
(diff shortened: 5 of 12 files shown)


R/autoload.R
```diff
@@ -81,3 +81,24 @@ renv_autoload_impl <- function() {
   TRUE
 
 }
+
+# TODO: this gets really dicey once the user starts configuring where
+# renv places its project-local state ...
+renv_project_find <- function(project = NULL) {
```
kevinushey, line 87: Why was this moved to `autoload.R`? By convention, files with the prefix `renv_<name>` and placed in `<name>.R`.

R/activate.R
```diff
@@ -142,7 +142,7 @@ renv_activate_prompt <- function(action, library, prompt, project) {
     prompt &&
     interactive() &&
     is.null(library) &&
-    !identical(project, Sys.getenv("RENV_PROJECT"))
+    !renv_project_is_active(project)
```
kevinushey, line 145:
```suggestion
    !renv_project_loaded(project)
```

R/project.R
```diff
@@ -18,17 +18,24 @@
...
 
   project
+}
+
+renv_project_resolve <- function(project = NULL) {
+  project <- project %||% renv_project_get()
+  normalizePath(project, winslash = "/", mustWork = FALSE)
+}
 
+renv_project_is_active <- function(project = NULL) {
```
kevinushey, line 37: This doesn't distinguish between "the user appears to have requested a particular project by setting the `RENV_PROJECT` environment variable" and "the project was actually loaded".

R/project.R
```diff
@@ -18,17 +18,24 @@
...
+renv_project_get <- function(default = getwd()) {
   project <- Sys.getenv("RENV_PROJECT", unset = NA)
   if (is.na(project))
     return(default)
 
   project
+}
+
+renv_project_resolve <- function(project = NULL) {
+  project <- project %||% renv_project_get()
```
kevinushey, line 33:
```suggestion
  project <- project %??% renv_project_get()
```

R/autoload.R
```diff
@@ -81,3 +81,24 @@ renv_autoload_impl <- function() {
...
+
+  project <- project %||% getwd()
```
kevinushey, line 89:
```suggestion
  project <- project %??% getwd()
```

R/load.R
```diff
@@ -64,8 +60,7 @@ load <- function(project = NULL, quiet = FALSE) {
   # then unload the current project and reload the requested one
   switch <-
     !renv_metadata_embedded() &&
-    !is.na(Sys.getenv("RENV_PROJECT", unset = NA)) &&
-    !identical(project, renv_project())
+    !renv_project_is_active(project)
```
kevinushey, line 63:
```suggestion
    !renv_project_loaded(project)
```

We might also need to do this stuff if any project is loaded (not just the current one)?

R/modify.R
```diff
@@ -38,7 +38,7 @@ renv_modify_impl <- function(project, changes) {
   else
     renv_modify_noninteractive(project, changes)
 
-  if (identical(renv_project(), project))
+  if (renv_project_is_active(project))
```
kevinushey, line 41:
```suggestion
  if (renv_project_loaded(project))
```

R/project.R
kevinushey, line 37:
```suggestion
renv_project_loaded <- function(project = NULL) {
```

R/load.R
```diff
@@ -29,11 +29,7 @@ load <- function(project = NULL, quiet = FALSE) {
 
   renv_scope_error_handler()
 
-  project <- normalizePath(
-    project %||% renv_project_find(project),
-    winslash = "/",
-    mustWork = TRUE
-  )
+  project <- renv_project_resolve(project)
```
kevinushey, line 32: Do we want to preserve the `renv_project_find()` behavior here? (I think we do)

R/activate.R
hadley: Are you suggesting a different name for this function?

R/project.R
hadley: I'm not sure you want to get into the details, but why have a separate `%??%` and `%||%` and why make `%??%` more like the `%||%` we use everywhere else? 😭

R/load.R
hadley: Yeah, I'll revert.

R/autoload.R
hadley: I moved it because that was the only place it was used, but since I'm reverting that change, I'll move it back.

hadley:
> Are we handling this correctly in any of the updated locations?

Yes; `renv_project_get()` is called with an explicit default in just about every location. I'll refactor to make `default` a required argument

> I think we should have a variable `_renv_project_path` that gets set when `renv::load()` is called, and use that wherever we need to know what the currently-loaded project is. (`RENV_PROJECT` is unreliable as it might be set by users, and it could be set even if the project is not yet loaded.)

Why would a user set `RENV_PROJECT`? How would it get set if the project is not yet loaded?

But I do agree that this should be an option, since it doesn't seem like you want subprocesses to inherit this value. If we change to the option, do you see any reason to continue supporting `RENV_PROJECT`?

> There's probably some conversation to be had about the distinction between `RENV_PROJECT` and the `renv.project.path` option.

There's some subtle bug that I haven't quite figured out due to `RENV_PROJECT` being set at the start of project loading and `renv.project.path` being set at the end. Assuming I can resolve this do you see any issue with eliminating that option?

R/activate.R
kevinushey: Yeah, that's my intent here.

kevinushey:
> Why would a user set RENV_PROJECT? How would it get set if the project is not yet loaded?

Good question. I guess I was imagining something like:

```
RENV_PROJECT=/path/to/project R -e 'renv::run(...)'
```

That is, having some way to load a project in a different location, but run using some separate working directory in an ergonomic way.

Now that I say it out loud, I'm not sure if anyone is actually using it this way...

> But I do agree that this should be an option, since it doesn't seem like you want subprocesses to inherit this value. If we change to the option, do you see any reason to continue supporting RENV_PROJECT?

I think that makes sense ... although I wonder what we should do in this scenario:

1. You work in an `renv` project, that project is loaded,
2. You `setwd("scripts")` or otherwise change the working directory,
3. You run something like `callr::r(function() { ... })`

Should that sub-process run with the same project loaded by default?

kevinushey:
> There's some subtle bug that I haven't quite figured out due to RENV_PROJECT being set at the start of project loading and renv.project.path being set at the end. Assuming I can resolve this do you see any issue with eliminating that option?

I think that makes sense. The intention was for `renv.project.path` to be set at the end of `renv::load()` so we could distinguish between the "requested" project and the "actually currently loaded" project.

But then I sort of feel like if this is something handled internally by `renv`, it should just be a variable in the `renv` namespace rather than an option.

hadley:
FWIW, as far as I can tell, `RENV_PROJECT` isn't documented anywhere, so it should be ok to remove.

In the callr scenario, it defaults to copying the `.libPaths()` from the current session, so you'll at least load packages from the right place. But probably `RENV_PROJECT` shouldn't be set, since we haven't actually loaded renv there.

kevinushey:
Sorry, I just remembered we're using it in the IDE:

https://github.com/rstudio/rstudio/blob/0e442b55fae1003f82bc3dc1cde2f037dc8815db/src/cpp/session/modules/SessionRenv.cpp#L39-L42

so we need to at least keep it alive for some amount of time for IDE compatibility.

kevinushey:
> I don't understand how renv_snapshot_task() worked because, AFAICT, project wasn't previously defined.

That looks like a bug; I'll see if I can fix.

PR #1248, opened 2023-05-12
rationalize usages of project accessors

kevinushey:
Similar approach to https://github.com/rstudio/renv/pull/1244.

### DIFF
--- a/R/project.R
+++ b/R/project.R
@@ -1,4 +1,8 @@
 
+# The path to the currently-loaded project, if any.
+# NULL when no project is currently loaded.
+`_renv_project_path` <- NULL
+
 #' Retrieve the Active Project
 #'
 #' Retrieve the path to the active project (if any).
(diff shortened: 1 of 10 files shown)


R/project.R
```diff
@@ -1,4 +1,8 @@
 
+# The path to the currently-loaded project, if any.
+# NULL when no project is currently loaded.
+`_renv_project_path` <- NULL
```
hadley, line 4: I don't understand why this works, because I thought the namespace locked all bindings when the package was loaded. To get around that problem, we usually do something like:

```R
the <- new.environment(parent = emptyenv)
the$project_path <- NULL
```

But I don't understand why it's not affecting you here

hadley: Ooooh, because you specifically unlock everything starting with `_` in `renv_binding_init()`

hadley (approved):
Should `renv_load_project()` also use `renv_project_set()`?

R/project.R
kevinushey: Another one of `renv`s dirty tricks. 🙂

kevinushey:
> Should renv_load_project() also use renv_project_set()?

My intention was to make sure `renv_project_set()` gets called at the end of all the `load()` gunk:

https://github.com/rstudio/renv/blob/fdc2a76dfcffe3f50917b498bc5064e12e3ec671/R/load.R#L713

so perhaps I should just remove https://github.com/rstudio/renv/blob/fdc2a76dfcffe3f50917b498bc5064e12e3ec671/R/load.R#L330 now?

PR #1250, opened 2023-05-12
Don't use renv with renv

hadley:
Experiment to see if we can move towards a more standard tidyverse workflow (which is easier for me to debug).


hadley:
I don't know why the older workflow is failing to build httpuv, but it shouldn't; switching to our standard workflow fixes the problem and means it's easier to fix future problems. 

I removed `appyveyor.yml` since our GHA windows actions should cover already the same cases, in a more standardised form. But you'll need to disable in https://github.com/rstudio/renv/settings/hooks.

I just copied the R versions that we usually use for tidyverse packages (release, dev, + 4 old versions), but since I think you want broader support for renv it might make more sense to declare a couple of specific older versions too. Just let me know what you want.

aronatkins:
Packrat CI currently follows the oldrel-4 pattern, even though we continue to use it in Connect and support restoring bundles for content created with older R. That gave us a great amount of version coverage without struggling to maintain working tests.

At one point in the past, I experimented with adding test coverage for all the old R versions, but struggled to get the right mix of package versions in place. I wanted repository snapshots (to test against historical versions of shiny and rmarkdown) but current versions (so the tests could use current testthat tooling). Right now, the Connect tooling tests R-3.6.3+.

Packrat embeds renv, so the two probably need to follow similar version-coverage patterns. I'm OK with whatever blend you decide to use.

kevinushey:
Thanks!

kevinushey:
I've also deleted the appveyor + Travis webhooks.

PR #1257, opened 2023-05-12
Proceed by default

### LINKED ISSUE #1240: Change prompt default to yes

hadley:
So we match the behaviour in non-interactive sessions (or when `prompt` has been set to `FALSE`).

Fixes #1240

(Hopefully this PR is easier to review 😆)

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -164,7 +164,7 @@ ask <- function(question, default = FALSE) {
 
 }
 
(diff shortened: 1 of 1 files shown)


kevinushey (approved):
Thanks; LGTM 🙂

PR #1269, opened 2023-05-16
Provide helper to compare new and old state

hadley:
The previous calls to waldo::compare() weren't very effective because it returns a value, it doesn't do anything. So now I include it in the text of the error message, which also makes it appear after the overall message instead of before it.

@kevinushey let me know if you'd prefer this organised differently or if you have better ideas for the function names.


kevinushey:
Maybe it'd be cleaner if we had a single function `renv_test_state()` that collected all of these various pieces into a single object? So we could do something like:

```
old <- renv_test_state()
# run test
new <- renv_test_state()
# compare old and new
```

And `renv_test_state()` would have the library paths, options, repository files, and so on.

hadley:
Oh yeah, that's a good idea.

tests/testthat/test-install.R
```diff
@@ -327,7 +327,10 @@ test_that("install() prefers cellar when available", {
   expect_equal(record$Source, "Cellar")
 
   prefix <- if (renv_platform_windows()) "file:///" else "file://"
-  uri <- paste0(prefix, renv_tests_path("local/skeleton"))
+  uri <- normalizePath(
```
kevinushey, line 330: What are we using `normalizePath()` here for? (Just feels awkward because this is a file URI not a path)

kevinushey (approved):

tests/testthat/test-install.R
hadley: Ooops, I should normalize the path of the url. It's my stab in the dark to make a test pass on windows. I think prior to #1272, the normalization was taken care of by someone else.

PR #1271, opened 2023-05-16
Drop renv_tests_init_envvars()

hadley:
Currently appears that it's only needed to reset the env vars set in `.Renviron` so all the tests pass (locally) if I just fail to set those.

@kevinushey opened for the purposes of discussion. If you think this is a reasonable approach, then I'll delete `renv_tests_init_envvars()`.


kevinushey:
I don't quite understand the motivation behind deleting `renv_tests_init_envvars()`; at least, removing `RETICULATE_PYTHON` is definitely necessary for the Python tests?

Do you want to just move this code somewhere else?

FWIW the project `.Renviron` was just there to make it easier to locally test `pak` integration while working on `renv`.

hadley:
My goal here is to discover exactly which env vars it's necessary to reset after each test file — ideally we want to get this set as small as possible. We might want to reset at the start of all tests, but otherwise, each test should be self-contained and leave envvars the way they found them. 

Maybe rather eliminating `renv_tests_init_envvars()` we should call it once in `setup.R` (so it's called once before all tests), rather than in `context()`, which is called once per file?

kevinushey:
> Maybe rather eliminating renv_tests_init_envvars() we should call it once in setup.R (so it's called once before all tests), rather than in context(), which is called once per file?

That makes sense to me. Conceptually, `renv_tests_init()` (and all the helpers) are intended to be run only once, before the test suite is run. (Or, before a single test is run, in the case of test_file)

hadley:
Do you think it's ok to drop the `.Renviron`? It's not a big, but usually these files in a project are anti-pattern (although I note you recover from the potential problem in `.Rprofile`). If we remove `.Renviron`, we could almost remove `.Rprofile`, apart from `system("git config core.hooksPath git/hooks")`. I don't think that's important for me (since I don't commit to main), but maybe you're still relying on it?

kevinushey:
> Do you think it's ok to drop the .Renviron?

That sounds fine to me.

> If we remove .Renviron, we could almost remove .Rprofile, apart from system("git config core.hooksPath git/hooks"). I don't think that's important for me (since I don't commit to main), but maybe you're still relying on it?

I'd like to keep it around; I think it's also nice in case we decide to introduce git hooks that we want to apply on other branches as well.

hadley:
Mostly superseded by #1282

PR #1272, opened 2023-05-16
Simplify path handling in tests

hadley:
Now that `renv_tests_root()` is no longer required in `tests/testhat.R` we can use `testthat::test_path()`.

I also removed the interactive working directory change when using RStudio, since in general that shouldn't be necessary. But I don't know why it was there in the first place, so I might be wrong.

### DIFF
--- a/R/tests.R
+++ b/R/tests.R
@@ -85,13 +85,6 @@ renv_tests_init_workarounds <- function() {
 
 }
 
-renv_tests_init_working_dir <- function() {
-  if (exists(".rs.getProjectDirectory")) {
-    home <- get(".rs.getProjectDirectory")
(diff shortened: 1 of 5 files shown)


kevinushey:
> I also removed the interactive working directory change when using RStudio, since in general that shouldn't be necessary. But I don't know why it was there in the first place, so I might be wrong.

This was mainly to help handle the case where I was using `devtools::load_all()`, trying to run some test code, but with the working directory still as `../renv` instead of `../renv/tests/testthat`. Presumedly that still works?

hadley:
Nothing in testthat will change the working directory; if you have problems it may suggest the need for `renv_tests_paths()`.

PR #1274, opened 2023-05-16
Create scoped repos helper

hadley:


R/tests.R
```diff
@@ -145,13 +145,28 @@ renv_tests_init_options <- function() {
 
 }
 
-renv_tests_init_repos <- function(repopath = NULL) {
+renv_tests_scope_repos <- function(envir = parent.frame()) {
```
hadley, line 148: Maybe `renv_test_scope()` should call this automatically? I think that would eliminate most changes to the tests.

kevinushey: I think it should -- IMHO, all "scoped" functionality should happen as part of a single `renv_test_scope()` call that's run at the start of a test.

kevinushey (approved):

PR #1282, opened 2023-05-17
Track envvars

hadley:


R/tests.R
```diff
@@ -10,23 +10,18 @@ renv_tests_scope <- function(packages = character(), project = NULL, envir = par
...
   renv_tests_init()
   renv_tests_scope_repos(envir = envir)
+  renv_scope_envvars(R_LIBS_USER = NULL, envir = envir)
 
   # ensure that attempts to restart are a no-op
   if (renv_rstudio_available())
     options(restart = function(...) TRUE)
 
-  # save local repositories
-  Sys.setenv(RENV_PATHS_LOCAL = renv_tests_path("local"))
```
hadley, line 19: This was only used by two tests, so they now do it themselves.

R/tests.R
```diff
@@ -10,23 +10,18 @@ renv_tests_scope <- function(packages = character(), project = NULL, envir = par
 
   renv_tests_init()
   renv_tests_scope_repos(envir = envir)
+  renv_scope_envvars(R_LIBS_USER = NULL, envir = envir)
```
hadley, line 13: `init()` changes this so this ensures we reset at the end of the test.

R/tests.R
```diff
@@ -53,20 +48,13 @@ renv_tests_scope <- function(packages = character(), project = NULL, envir = par
...
-  Sys.unsetenv("RENV_PROFILE")
-  Sys.unsetenv("RENV_PROJECT")
   Sys.unsetenv("RENV_PATHS_ROOT")
   Sys.unsetenv("RENV_PATHS_LIBRARY")
   Sys.unsetenv("RENV_PATHS_LIBRARY_ROOT")
   Sys.unsetenv("RENV_PATHS_LOCAL")
   Sys.unsetenv("RENV_PATHS_LOCKFILE")
   Sys.unsetenv("RENV_PATHS_RENV")
 
-  Sys.unsetenv("RENV_PYTHON")
```
hadley, line 65: Now handled by `renv_test_scope_python()`

R/tests.R
```diff
@@ -53,20 +48,13 @@ renv_tests_scope <- function(packages = character(), project = NULL, envir = par
 
 renv_tests_init_envvars <- function() {
 
-  Sys.unsetenv("RENV_PROFILE")
```
hadley, line 56: Now set/restored in the one profile test that needs it.

tests/testthat/helper-aaa.R
```diff
@@ -60,11 +60,17 @@ renv_test_state <- function() {
   opts$repos[opts$repos == "@CRAN@"] <- "https://cloud.r-project.org"
   opts <- opts[csort(names(opts))]
 
+  envvars <- as.list(Sys.getenv())
+  envvars <- envvars[grep("^RENV_DEFAULT_", names(envvars), invert = TRUE)]
```
hadley, line 64: Let me know if you think we should be tracking these env vars, but they didn't seem that important, and frequently changed.

R/tests.R
kevinushey: Maybe worth adding as a comment in the code so we remember the reason?

tests/testthat/helper-aaa.R
kevinushey: This looks fine to me too.

kevinushey (approved):

PR #1283, opened 2023-05-17
Drop withr dependency

hadley:
Now that I better understand the renv scope system


kevinushey (approved):
LGTM, but I think you said you wanted to use `withr` because the `renv` equivalents didn't handle exit handlers set on the global environment / at the top level? Or are you finding that we can get by without that?

hadley:
I now feel like it'll easier to add that to renv than switch to use withr everywhere.

PR #1286, opened 2023-05-18
Use testthat 3e

hadley:
This will allow us to use snapshot tests and parallel testing.

Main changes here are because testthat 3e is less aggressive about silencing output, so we have to do a little more ourselves.


R/scope.R
```diff
@@ -368,9 +368,9 @@ renv_scope_trace <- function(what, tracer, envir = NULL) {
   call <- sys.call()
   call[[1L]] <- base::trace
   call[["print"]] <- FALSE
-  eval(call, envir = parent.frame())
+  suppressMessages(eval(call, envir = parent.frame()))
```
kevinushey, line 371: Is this to suppress the `method`-related messages?

tests/testthat/test-bioconductor.R
```diff
@@ -12,7 +10,7 @@ test_that("packages can be installed, restored from Bioconductor", {
   renv_scope_options(repos = c(CRAN = "https://cloud.r-project.org"))
 
   install.packages("BiocManager", quiet = TRUE)
-  BiocManager::install('Biobase', quiet = TRUE, ask = FALSE)
+  suppressMessages(BiocManager::install('Biobase', quiet = TRUE, ask = FALSE))
```
kevinushey, line 13:
```suggestion
  suppressMessages(BiocManager::install("Biobase", quiet = TRUE, ask = FALSE))
```

kevinushey (approved):

R/scope.R
hadley: It suppresses this message:

```R
trace(mean, quote(1+ 1))
#> Tracing function "mean" in package "base"
```

PR #1287, opened 2023-05-18
Eliminate test-aaa.R and test-zzz.R

hadley:
* `test-aaa.R` appears to be a path test
* `test-zzz.R` is no longer needed since all tests ensure that they clean up after themselves


kevinushey:
For context, that test is really a pre-flight requirement for tests on CRAN: If we're not using a temporary directory for some reason (due to some internal bug / error), then we should fail immediately since otherwise we risk writing / updating files in unintended locations on the CRAN build server.

Is there a better place to move that? Maybe `setup.R`?

kevinushey:
For `test-zzz.R`, it should probably suffice to confirm that `R CMD check` doesn't complain about detritus.

hadley:
Ok, with that in mind, I think I like this approach better.  It always felt a bit weird to have functions in `setup.R` that are just called once, so I've moved them to `helper-scope.R`. I tweaked the names to better reflect that they are (mostly) scoped functions now, and added a comment to describe that you might occasionally need these interactively too (if stepping through a test line-by-line).  Now setup can just call `renv_test_scope_setup()` and verify that the root dir is as expected.

PR #1290, opened 2023-05-18
Track and eliminate files left in the tempdir

hadley:
I'm not sure how important this is, but most of the changes a straightforward, so I think it makes sense to merge this PR even if we decide to back out of systematically monitoring the tempdir.

@kevinushey I got stuck eliminating at `renv-library-*`, which I think is coming from `renv_libpaths_safe_tempdir()`.  It wasn't clear to me how to pick a scope to clean this up correctly. You should probably also carefully review the changes I made in `R/`, in case I got the scope wrong.

### DIFF
--- a/R/available-packages.R
+++ b/R/available-packages.R
@@ -107,6 +107,8 @@ renv_available_packages_query <- function(type, repos, quiet = FALSE) {
 renv_available_packages_query_impl_packages_rds <- function(url) {
   path <- file.path(url, "PACKAGES.rds")
   destfile <- tempfile("renv-packages-", fileext = ".rds")
+  defer(unlink(destfile))
+
   download(url = path, destfile = destfile, quiet = TRUE)
   suppressWarnings(readRDS(destfile))
 }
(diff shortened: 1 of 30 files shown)


R/available-packages.R
```diff
@@ -107,20 +107,26 @@ renv_available_packages_query <- function(type, repos, quiet = FALSE) {
 renv_available_packages_query_impl_packages_rds <- function(url) {
   path <- file.path(url, "PACKAGES.rds")
   destfile <- tempfile("renv-packages-", fileext = ".rds")
```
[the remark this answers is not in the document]
hadley: For some reason I had a mental blank against doing that in code in R.

hadley:
Merge conflicts resolved and I've switched to `revn_scope_tempfile()` in a lot more places. There are about 5 tests left that generate tempfiles but they're mostly generated outside of renv, so I'm going to leave them for now, and I've commented out `testfiles` from the state object.

R/package.R
```diff
@@ -443,8 +443,8 @@ renv_package_unpack <- function(package, path, subdir = "", force = FALSE) {
   }
 
   # create extraction directory
-  old <- tempfile("renv-package-old-")
-  new <- tempfile("renv-package-new-")
+  old <- renv_scope_tempfile("renv-package-old-")
+  new <- renv_scope_tempfile("renv-package-new-", envir = parent.frame())
```
kevinushey, line 447: any reason why we need the explicit environment here but not above?

kevinushey (approved):

hadley:
This is going to cause a bunch of merge conflicts with #1325, but I think it'll be easier to resolve them in this PR, so I'll wait until you've merged the other one.

R/package.R
hadley: Because the function returns `new`, so we want to make sure it's available to the caller. This might be too magical, and it might be better to not cleanup at all here.

PR #1292, opened 2023-05-18
Use `csort()` in more places

hadley:
`csort()` now is the only place that uses `renv_scope_locale()`, so I could simplify/inline that logic too, if you wanted.


kevinushey (approved):
LGTM! Let's keep `renv_scope_locale()` around anyhow.

PR #1295, opened 2023-05-19
Eliminate use of %||% with environments

hadley:
This is the first step towards unifying `%||%` and `%??%`. After a few abortive attempts, I figured out the best way to do this is to slowly move usage of `%||%` to `%??%`, since that makes it stricter. Then, once that's all done, I can rename `%??%` to `%||%`.

This first PR eliminates the use of `%||%` with environments by using argument defaults more in `scope.R`, and using `%??%` for environment caching elsewhere.

This includes a temporary warning in `%||%` and a temporary elimination of `warn = 1`. If you think this first step is sound, I'll remove these changes prior to merging.

### DIFF
--- a/R/scope.R
+++ b/R/scope.R
@@ -1,13 +1,11 @@
 
 renv_scope_tempdir <- function(pattern = "renv-tempdir-",
                                tmpdir = tempdir(),
-                               envir = NULL)
+                               envir = parent.frame())
 {
   dir <- tempfile(pattern = pattern, tmpdir = tmpdir)
   ensure_directory(dir)
   owd <- setwd(dir)
-
-  envir <- envir %||% parent.frame()
 
   defer({
     setwd(owd)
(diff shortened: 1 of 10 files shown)


R/scope.R
```diff
@@ -374,10 +347,8 @@ renv_scope_trace <- function(what, tracer, envir = NULL) {
 
 }
 
-renv_scope_var <- function(key, value, frame = NULL, envir = NULL) {
+renv_scope_var <- function(key, value, frame, envir = parent.frame()) {
```
hadley, line 350: I eliminated the default for `frame` here since I think it's sufficiently important that you should always supply explicitly (and that only changed a single call elsewhere).

kevinushey: That sounds reasonable to me.

R/scope.R
```diff
@@ -1,22 +1,20 @@
 
 renv_scope_tempdir <- function(pattern = "renv-tempdir-",
                                tmpdir = tempdir(),
-                               envir = NULL)
+                               envir = parent.frame())
```
kevinushey, line 4: For some reason, I was worried about `envir` being a promised and its value potentially depending on where / when that promise gets evaluated -- presumedly that's not actually an issue?

kevinushey (approved):

R/scope.R
hadley: That's what I thought you might be worried about, but no, it's not a problem.

PR #1296, opened 2023-05-19
Experiment: switch from `global(x)` to `the`

hadley:
This is just a spike to see what you think of the `the` syntax. If you like it, I'll convert `` `_renv_*` `` to use it; if you don't like it, I'll drop this PR and instead convert to `` `_renv_` `` style.


R/skip.R
```diff
@@ -3,10 +3,10 @@ skip_if_no_python <- function(python = NULL) {
 
   python <- python %||% "python3"
   key <- paste("tests", python, "installed", sep = ".")
-  installed <- global(key, {
+  installed <- the[[key]] %??% {
```
kevinushey, line 6: Does this need to assign `the[[key]]` as well?

R/skip.R
```diff
@@ -20,11 +20,11 @@ skip_if_no_virtualenv <- function(python = NULL) {
   skip_if_no_python(python)
 
   key <- paste("tests", python, "virtualenv.installed", sep = ".")
-  installed <- global(key, {
+  installed <- the[[key]] %??% {
```
kevinushey, line 23: Does this need to assign `the[[key]]` as well?

R/tests.R
```diff
@@ -136,7 +136,8 @@ renv_tests_diagnostics <- function() {
 
 # Cache absolute path to tests/testthat
 renv_tests_root <- function() {
-  global("tests.root", normalizePath(testthat::test_path(".")))
+  the$tests.root <- the$tests.root %??% normalizePath(testthat::test_path("."))
+  the$tests.root
```
kevinushey, line 140: This line might not be necessary? (The above should invisibly return the assigned value IIUC)

R/use.R
```diff
@@ -96,7 +96,8 @@ use <- function(...,
 }
 
 renv_use_libpath <- function() {
-  global("use.library", tempfile("renv-library-"))
+  the$use.library <- the$use.library %||% tempfile("renv-library-")
+  the$use.library
```
kevinushey, line 100: This line might not be necessary? (The above should invisibly return the assigned value IIUC)

kevinushey (commented):
Let me stew on this for a while before we approve / merge.

R/skip.R
hadley: Yup, good catch.

R/tests.R
hadley: It's not strictly necessary, but I think it's weird for `renv_test_root()` to return invisibly since you're not calling it for its side effect.

Maybe a clearer/simple change would be to wrap the assignment in `()`?

kevinushey: Ah, I hadn't thought about visibility of a return value as a signal that the function is used for its side effect.

Wrapping with `()` seems okay.

PR #1297, opened 2023-05-19
Run tests in parallel

hadley:
After setting `TESTTHAT_CPUS=4` in my `.Renviron`, this took test run time down from ~200s to 50s.


tests/testthat/setup.R
```diff
@@ -56,6 +56,9 @@ renv_tests_init_options <- function() {
 renv_tests_init_packages <- function() {
   requireNamespace("waldo", quietly = TRUE)
   renv_namespace_load("crayon")
+  renv_namespace_load("diffobj")
```
hadley, line 59: I think I should probably switch back to the previous approach; but maybe I've got all the important deps now?

kevinushey: It might be worthwhile just in case the dependencies (or recursive dependencies?) of these packages change in a way that break `renv` unit tests.

hadley:
That reminds me that I need to remove the pkgload remoteline because it won't actually get installed because pkgload isn't a direct dependency.

kevinushey:
> That reminds me that I need to remove the pkgload remoteline because it won't actually get installed because pkgload isn't a direct dependency.

That seems sort of surprising, but I'm guessing it's intentional? Is it worth considering `Remotes:` packages as dependencies even if they're not in Depends / Imports / Suggests?

kevinushey:
> After setting TESTTHAT_CPUS=4

Does testthat set this to an appropriate number on the user's behalf? Or should we set `TESTTHAT_CPUS` somewhere? It might make sense to use `parallel::detectCores()` as well.

hadley:
@kevinushey it defaults to 2 cores, and IIRC the reasoning is that number of cores to use for testing is really something user/computer specific rather than project specific. For example, if we set it to 2 in `.Renviron` we'd need to re-set it for GHA since they only have two virtual cores.

And agreed: https://github.com/r-lib/pak/issues/496.

hadley:
I'm confident the build failures are unrelated; I'm working on generally improving reliability in other PRs.

kevinushey (approved):

Issue #1293, opened 2023-05-18
`boostrap()` should use `stdout` not `stderr`


hadley:
i.e. it should use `cat()` instead of `message()`

So its feedback are the same colour as restore/snapshot etc.

PR #1299, opened 2023-05-20
use cat for bootstrap messages

kevinushey:
Closes https://github.com/rstudio/renv/issues/1293.

I also included some changes here to make sure I could run tests interactively, without those changes if I try to `devtools::load_all()` and start running tests line-by-line (that is, `Cmd + Enter` to run lines in the Console) I see errors like:

```
>   renv_tests_scope()
Error in `testthat::teardown_env()` at tests/testthat/setup.R:9:6:
! `teardown_env()` has not been initialized
ℹ This is an internal error that was detected in the testthat package.
  Please report it at <https://github.com/r-lib/testthat/issues> with a reprex and the full backtrace.
Run `rlang::last_trace()` to see where the error occurred.
```

I'm not sure if that's actually a testthat bug based on how I'm trying to do things here though.


tests/testthat/test-bootstrap.R
```diff
@@ -5,9 +5,10 @@ test_that("we can bootstrap the current version of renv", {
   skip_on_ci()
 
   renv_tests_scope()
+  renv_scope_options(renv.bootstrap.quiet = TRUE)
```
hadley, line 8: Might be simpler to just set once in `renv_scope_options()`?

hadley:
Maybe I accidentally fixed this in #1287, because I can't reproduce it on main:

```R
devtools::load_all(".")
renv_scope_options
```

hadley:
But I've seen some other test weirdness which should be fixed by https://github.com/rstudio/renv/pull/1306 and https://github.com/r-lib/testthat/pull/1790.

tests/testthat/helper-scope.R
```diff
@@ -1,5 +1,16 @@
-renv_tests_scope <- function(packages = character(), project = NULL, envir = parent.frame()) {
 
+renv_tests_scope <- function(packages = character(),
+                             project = NULL,
+                             envir = parent.frame())
+{
+  # source setup.R if necessary (for interactive scenarios)
+  running <- getOption("renv.tests.running", default = FALSE)
```
hadley, line 7: You should be able to revert all these changes now, right?

R/bootstrap.R
```diff
@@ -7,6 +7,19 @@
...
 }
 
+catf <- function(fmt, ..., appendLF = TRUE) {
+
+  quiet <- getOption("renv.bootstrap.quiet", default = FALSE)
+  if (quiet)
+    return(invisible())
+
+  msg <- sprintf(fmt, ...)
+  cat(msg, file = stdout(), sep = if (appendLF) "\n" else "")
```
kevinushey, line 17: TIL that R does something odd if `sep` contains a newline; in particular, it writes a trailing newline if `sep` _contains_ a newline. I guess I'm taking advantage of that here, but it's definitely a little odd.

`writeLines()` is more consistent in this regard.

tests/testthat/helper-scope.R
kevinushey: Don't I still need this if I want `setup.R` to be run before running a test?

hadley: No? It should be run automatically.

kevinushey: I'm not seeing that on main right now (unless I keep in the bit of code referenced here).

PR #1304, opened 2023-05-22
Identify all uses of %||% for non-NULL 0-length vectors

hadley:
Next step in the standardization of `%||%` semantics. After this `%||%` can be redefined as `function(x, y) if (is.null(x)) y else x` and `%??%` can be removed.  I'm not sure what to call the new function that's used specifically for 0 length vectors.

I'll revert the change to options and to `%||%` prior to merging.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -135,7 +135,7 @@ install <- function(packages = NULL,
   }
 
   # get and resolve the packages / remotes to be installed
-  remotes <- packages %||% renv_project_remotes(project, dependencies)
+  remotes <- packages %|0|% renv_project_remotes(project, dependencies)
   if (empty(remotes)) {
     vwritef("* There are no packages to install.")
     return(invisible(list()))
--- a/R/install.R
+++ b/R/install.R
@@ -105,9 +105,12 @@ install <- function(packages = NULL,
   renv_scope_error_handler()
 
   # allow user to provide additional package names as part of '...'
-  dots <- list(...)
-  names(dots) <- names(dots) %||% rep.int("", length(dots))
-  packages <- c(packages, dots[!nzchar(names(dots))])
+  if (!missing(...)) {
+    dots <- list(...)
+    names(dots) <- names(dots) %||% rep.int("", length(dots))
+
+    packages <- c(packages, dots[!nzchar(names(dots))])
+  }
 
   project <- renv_project_resolve(project)
   renv_project_lock(project = project)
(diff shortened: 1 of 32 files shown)


kevinushey (approved):

kevinushey:
> I'm not sure what to call the new function that's used specifically for 0 length vectors.

Or perhaps this is a sign that we have some functions returning zero-length vectors that should be returning `NULL` instead?

I feel like this was one of the thorniest examples: https://github.com/rstudio/renv/blob/8a855875014fff77455a014da362aaab31453048/R/paths.R#L43

hadley:
Can you give me a bit more context on that example? Because I looked at it, but it wasn't obvious to me why `""` was a more sensible default.

kevinushey:
I _think_ it was to defend against cases where `renv_paths_common()` as called with a `NULL` argument, e.g. a call like:

```
> renv:::renv_paths_common("test", NULL, NULL)
[1] ""
```

But I can't recall where that issue arose. We could try removing it and see where the failures are, and fix those with something more appropriate?

Given that `renv_paths_common()` is vectorized (intentionally) it feels like the right fix would live upstream anyhow.

hadley:
I agree that the right fix is to move the handling upstream, but that looks like a high risk change to me because it has to propagate out to every use of `renv_paths_cache()`, `renv_paths_local()` etc. So let me try eliminating `%|0|%` in favour of handling inline.

R/paths.R
```diff
@@ -40,13 +40,8 @@ renv_paths_common <- function(name, prefixes = NULL, ...) {
   else
     root
 
-  file.path(prefixed, ...) %||% ""
-
-}
-
-renv_paths_project <- function(..., project = NULL) {
```
hadley, line 47: `renv_paths_project()` wasn't used any where

kevinushey (approved):

R/install.R
```diff
@@ -135,7 +135,12 @@ install <- function(packages = NULL,
   }
 
   # get and resolve the packages / remotes to be installed
-  remotes <- packages %||% renv_project_remotes(project, dependencies)
+  if (length(packages) == 0) {
+    remotes <- renv_project_remotes(project, dependencies)
+  } else {
+    remotes <- packages
+  }
```
kevinushey, line 142:
```suggestion
  remotes <- if (length(packages) == 0) {
    renv_project_remotes(project, dependencies)
  } else {
    packages
  }
```

Not sure how you feel, but I usually prefer assigning the result of the `if` expression if all branches are just assigning the same variable.

kevinushey: We could also change the code here: https://github.com/rstudio/renv/pull/1304/files?w=1#diff-76ebee0a54b80584fcf1eff0cdf55ef6f17e47a6d80cb807b5939e8e50cac32cR110

Since that would help us preserve the NULL-ness of the incoming argument.

hadley: Oh yeah, let's change it there.

PR #1305, opened 2023-05-22
Download test refactoring

hadley:
* Use webfakes to generate a temporary server for testing
* Use `renv_binding_replace()` to modify and restore `utils:::makeUserAgent()`. Correctly restore value when function ends.

### DIFF
--- a/R/download.R
+++ b/R/download.R
@@ -206,7 +206,7 @@ renv_download_default <- function(url, destfile, type, request, headers) {
 
 }
 
-renv_download_default_agent_scope <- function(headers) {
+renv_download_default_agent_scope <- function(headers, envir = parent.frame()) {
 
   if (empty(headers))
     return(FALSE)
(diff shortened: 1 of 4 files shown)


R/tests.R
```diff
@@ -1,13 +1,3 @@
-
-renv_tests_program <- function(name) {
```
hadley, line 2: Was only used by the download test; we have other more specific helpers for python.

R/download.R
```diff
@@ -233,17 +229,16 @@ renv_download_default_agent_scope_impl <- function(headers, envir = NULL) {
   if (!ok)
     return(FALSE)
 
-  do.call("unlockBinding", list("makeUserAgent", utils))
-  defer(do.call("lockBinding", list("makeUserAgent", utils)), envir = envir)
```
hadley, line 237: Previously this restored the lock, but not the original value.

R/download.R
```diff
@@ -233,17 +229,16 @@ renv_download_default_agent_scope_impl <- function(headers, envir = NULL) {
...
-
   agent <- makeUserAgent(FALSE)
   all <- c("User-Agent" = agent, headers)
   headertext <- paste0(names(all), ": ", all, "\r\n", collapse = "")
 
-  assign("makeUserAgent", envir = utils, function(format = TRUE) {
+  renv_binding_replace("makeUserAgent", envir = utils, function(format = TRUE) {
```
kevinushey, line 236: Maybe we can use `renv_scope_var()` (or should we call it `renv_scope_binding()`?)

kevinushey (approved):

PR #1307, opened 2023-05-22
Provide helper to execute deferred handlers interactively

hadley:
This allows you to reset the state after running (e.g.) `renv_tests_scope()` interactively.

### DIFF
--- a/R/defer.R
+++ b/R/defer.R
@@ -10,3 +10,6 @@ defer <- function(expr, envir = parent.frame()) {
 
 }
 
+deferred_run <- function(envir = parent.frame()) {
+  renv_exit_handlers_execute(envir)
+}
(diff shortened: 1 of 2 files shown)


kevinushey:
What do you think about having an active binding to do this for you?

https://github.com/rstudio/renv/blob/8b70fe5bb604ad7e9477ad9ce1d02e8b9f0ab47f/tests/testthat/setup.R#L4-L10

So you can just write `done` and hit enter. Or maybe we just need some convention / shortcut that we can use in RStudio.

hadley:
I don’t have strong feelings; I just copied the name from withr. Let’s try the active binding and see how it feels.

kevinushey (approved):

PR #1308, opened 2023-05-22
Gate slow tests behind an env var

hadley:
Mostly converted from `skip_sometimes()` with a few additional new tests to make `test-bioconductor.R` and `test-python.R` run a bit faster.

### DIFF
--- a/.github/workflows/slow-tests.yaml
+++ b/.github/workflows/slow-tests.yaml
@@ -1,0 +1,38 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, master]
+  pull_request:
+    branches: [main, master]
+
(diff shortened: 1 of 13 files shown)


kevinushey (commented):
FWIW `skip_sometimes()` was also there to try and help prevent issues that arise with rate limiting on GitHub. There's almost surely a better solution though.

.github/workflows/slow-tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+    branches: [main, master]
+
+name: slow-tests
+
+jobs:
+  slow-tests:
+    runs-on: ubuntu-latest
+
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
```
hadley, line 15: Fix this

hadley:
I think the most skipping when `GITHUB_PAT` is absent seems to have taken care of the rate limiting stuff.

kevinushey (approved):

PR #1313, opened 2023-05-22
Don't namespace renv functions in tests

hadley:
(Except where needed)


kevinushey (approved):

tests/testthat/test-install.R
```diff
@@ -140,7 +140,7 @@ test_that("various remote styles can be used during install", {
   expect_true(renv_package_version("skeleton") == "1.0.2")
 
   # install from subdir
-  install("kevinushey/subdir:subdir")
+  install("kevinushey/subdir/subdir")
```
kevinushey, line 143: This is intentional; this is one of the places that `renv` deviates from the `remotes` spec.

kevinushey: The reason here being I wanted to resolve the ambiguity in `a/b/c`, since certain remotes might have that mean:

- organization a; user b; repository c,
- user a; repository b; sub-directory c

IIRC this is relevant for GitLab; see https://github.com/rstudio/renv/blob/bab78b78dede06e61c3db3c2163973c4f2747bab/tests/testthat/test-remotes.R#L184-L210

hadley: PR incoming.

kevinushey (approved):

R/scope.R
```diff
@@ -347,17 +347,15 @@ renv_scope_trace <- function(what, tracer, envir = parent.frame()) {
 
 }
 
-renv_scope_var <- function(key, value, frame, envir = parent.frame()) {
 
-  if (exists(key, envir = frame, inherits = FALSE)) {
-    saved <- get(key, envir = frame, inherits = FALSE)
-    assign(key, value, envir = frame, inherits = FALSE)
-    defer(assign(key, saved, envir = frame, inherits = FALSE), envir = envir)
+renv_scope_binding <- function(envir, symbol, replacement, frame = parent.frame()) {
```
kevinushey, line 351: I feel like we need different names here -- usually `envir` is used in these scope functions to mean something like "stack frame".

kevinushey: That's why I originally had:

```suggestion
renv_scope_binding <- function(frame, symbol, replacement, envir = parent.frame()) {
```

but maybe some name other than `frame` is better here -- `container`? Something else?

kevinushey (approved):

R/scope.R
kevinushey: Or perhaps all of these `scope` functions should accept an argument called `scope` instead of `envir` just to remove some of the naming ambiguity?

hadley: Yeah, I'll do another PR that renames to scope. 

The problem with the original names is that `envir` is the call frame and `frame` is the environment to modify, so they're perfectly wrong 😄

hadley: But I didn't mean to do that change in this PR, so I'll remove it.

PR #1314, opened 2023-05-22
WIP: Reduce/consolidate skips

hadley:


tests/testthat/test-acls.R
```diff
@@ -2,7 +2,7 @@
 test_that("renv_acls_reset() can reset ACLs", {
 
   skip_on_cran()
-  skip_if(!renv_platform_linux())
+  skip_on_os(c("windows", "mac"))
```
kevinushey, line 5: This test would fail on non-Linux Unix machines so it really is intended to be "Linux-only" rather than not-Windows / not-macOS.

tests/testthat/test-available-packages.R
```diff
@@ -70,7 +70,6 @@ test_that("available_packages() succeeds with unnamed repositories", {
 test_that("renv_available_packages_latest() respects pkgType option", {
 
   skip_on_cran()
-  skip_if(.Platform$pkgType == "source")
```
kevinushey, line 73: Can you check if this test succeeds if you run:

```
platform <- .Platform
platform$pkgType <- "source"
renv:::renv_binding_replace(".Platform", platform, envir = baseenv())
```

before running this test? IIRC I ran into some issues here when using R from Homebrew / compiled from sources for some reason.

tests/testthat/test-cache.R
```diff
@@ -188,7 +188,7 @@ test_that("invalid Built field is detected", {
 test_that("ACLs set on packages in project library are reset", {
 
   skip_on_cran()
-  skip_if(!renv_platform_linux())
+  skip_on_os(c("windows", "mac"))
```
kevinushey, line 191: Same as above Linux comment.

tests/testthat/test-config.R
```diff
@@ -89,7 +89,7 @@ test_that("multiple library paths can be set in RENV_CONFIG_EXTERNAL_LIBRARIES",
 test_that("cache symlinks are disabled if the cache and project library lie in different volumes", {
 
   skip_on_cran()
-  skip_if(renv_platform_unix())
+  skip_on_os(c("linux", "mac"))
```
kevinushey, line 92: Same as above Linux comment.

tests/testthat/test-download.R
```diff
@@ -114,7 +114,7 @@ test_that("downloads work with file URIs", {
 
 test_that("downloads work with UNC paths on Windows", {
   skip_on_cran()
-  skip_if_not(renv_platform_windows())
+  skip_on_os(c("linux", "mac"))
```
kevinushey, line 117: `skip_unless_os("windows")`?

tests/testthat/test-acls.R
kevinushey: In other words, `skip_unless_os("linux")`?

kevinushey: I also don't like `skip_on_os()` because I can never remember if macOS is "mac", "macos", "Darwin", or whatever else. I know I can look at the documentation but autocompletion fills in the blank if I have `skip_on_macos()` or something.

hadley: Can you explain more the distinction between unix and linux here? I didn't think we ran tests on non-Linux unix platforms so I don't understand why the distinction is important.

FWIW you can just call `skip_on_os("foo")` and the error message will help you 😄

tests/testthat/test-available-packages.R
hadley: You're right, it fails. I'm not sure why I removed it, but I'll restore it.

tests/testthat/test-download.R
kevinushey: I would prefer the full set, but I understand if you don't want to -- `skip_unless_os()` would make me happy enough.

hadley: I think it's easiest to start by adding the per OS helpers to renv, and I'll see how it feels.

kevinushey:
I'm going to close this as it's gotten stale -- feel free to re-open if you plan on finishing this up + getting it ready for merge.

PR #1325, opened 2023-05-23
separate 'setup' code from 'fixture' code

kevinushey:


R/defer.R
```diff
@@ -16,22 +13,8 @@ defer <- function(expr, envir = parent.frame()) {
 
 }
 
-deferred_run <- function(envir = parent.frame()) {
```
kevinushey, line 19: This felt superfluous since we have `renv_defer_execute()` but let me know if you'd like to keep it.

I also found a separate way to allow `done` to run deferred handlers on the global environment; see below in the PR and let me know if that's too gross or not.

R/utils.R
```diff
@@ -541,5 +540,7 @@ untar <- function(tarfile,
 }
 
 rep_named <- function(names, x) {
-  stats::setNames(rep_len(x, length(names)), names)
```
kevinushey, line 544: Just to avoid the "dependency" on stats.

R/defer.R
```diff
@@ -16,22 +13,8 @@ defer <- function(expr, envir = parent.frame()) {
...
-  id <- attr(envir, "__renv_defer_id__", exact = TRUE)
-  if (!is.null(id)) {
-    return(id)
-  }
-
-  # no id; bump the global count and add a new id
-  `_renv_defer_id` <<- `_renv_defer_id` + 1L
-  attr(envir, "__renv_defer_id__") <- as.character(`_renv_defer_id`)
-
+  format(envir)
```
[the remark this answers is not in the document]
hadley: Yeah, I think this is the best you can do without C code.

R/scope.R
```diff
@@ -375,3 +373,15 @@ renv_scope_umask <- function(umask, envir = parent.frame()) {
   defer(Sys.umask(oldmask), envir = envir)
   invisible(oldmask)
 }
+
+renv_scope_setwd <- function(dir, envir = parent.frame()) {
```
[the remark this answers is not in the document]
hadley: I was thinking `renv_scope_wd()`, since the set is sort of implied. There are also tons of places elsewhere in renv where this could be used.

R/scope.R
```diff
@@ -375,3 +373,15 @@ renv_scope_umask <- function(umask, envir = parent.frame()) {
...
+  owd <- setwd(dir)
+  defer(setwd(owd), envir = envir)
+  invisible(owd)
+}
+
+renv_scope_sandbox <- function(envir = parent.frame()) {
```
hadley, line 383: See also `renv_scoped_sandbox()` (which has a typo in the name, oops) in `test-sandbox.R`

R/tests.R
```diff
@@ -124,17 +127,21 @@ renv_tests_diagnostics <- function() {
...
 
 renv_tests_path <- function(path = NULL) {
-  if (is.null(path)) {
-    renv_tests_root()
-  } else {
-    file.path(renv_tests_root(), path)
-  }
+
+  # special case for NULL path
+  if (is.null(path))
```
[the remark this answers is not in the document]
hadley: I see this more as working around a buglet in `file.path()`, because IMO `file.path()` should silently drop all `NULL` arguments, but the early return is fine too.

tests/testthat/helper-scope.R
```diff
@@ -3,29 +3,24 @@ renv_tests_scope <- function(packages = character(),
...
-    sys.source(path, envir = globalenv())
-  }
-
-  # use local repositories in this scope
+  # use private repositories in this scope
   renv_tests_scope_repos(envir = envir)
 
+  # use sandbox in this scope
+  # TODO(Kevin): Without this, tests fail trying to install BiocManager?
+  renv_scope_sandbox(envir = envir)
```
hadley, line 11: It's worth thinking about whether it's better to apply that here, or in individual Bioc tests that need it.

tests/testthat/helper-scope.R
```diff
@@ -37,20 +32,18 @@ renv_tests_scope <- function(packages = character(),
...
-  .libPaths(lib)
+  renv_scope_libpaths(lib, envir = envir)
 
-  defer(envir = envir, {
-    setwd(owd)
-    unlink(lib, recursive = TRUE)
-    .libPaths(libpaths)
-  })
+  # make sure we clean up the library path when we're done too
+  defer(unlink(lib, recursive = TRUE), envir = envir)
```
hadley, line 38: You can eliminate this by using `renv_scope_tempfile()` instead of `tempfile()` above.

tests/testthat/helper-setup.R
```diff
@@ -0,0 +1,235 @@
...
+    RENV_PATHS_LIBRARY = NULL,
+    RENV_PATHS_LIBRARY_ROOT = NULL,
+    RENV_PATHS_LOCAL = NULL,
+    RENV_PATHS_LOCKFILE = NULL,
+    RENV_PATHS_RENV = NULL,
+    RENV_AUTOLOAD_ENABLED = FALSE,
+    envir = envir
+  )
+
+  if (is.na(Sys.getenv("GITHUB_PATH", unset = NA))) {
```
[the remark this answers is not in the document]
hadley: Ooops, yes.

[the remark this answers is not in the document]
hadley: FWIW there are ~15 instances of this pattern in renv, so maybe worth pulling out into a helper?

```R
renv_envvar_exists <- function(key) {
  is.na(Sys.getenv(key, unset = NA))
}
```

tests/testthat/helper-setup.R
```diff
@@ -0,0 +1,235 @@
...
+    renv.config.sandbox.enabled = TRUE,
+    restart = NULL,
+    renv.config.install.transactional = FALSE,
+    renv.tests.running = TRUE,
+    envir = envir
+  )
+
+}
+
+renv_tests_setup_tools <- function(envir) {
```
hadley, line 83: I think this could go in its own helper, outside of a function, so that its run by `load_all()`?

tests/testthat/helper-setup.R
```diff
@@ -0,0 +1,235 @@
...
+
+  if (interactive()) {
+
+    # create a 'done' object that, when printed, will
+    # run any pending defer handlers
+    envir <- attach(NULL, name = "renv:tools")
```
hadley, line 89: Given that you're changing from an active binding to a print method, are you sure that it needs to go into a special environment?

But TBH it feels like that this is a lot of gymnastics to save typing two characters, which autocomplete would provide for you anyway.

tests/testthat/helper-setup.R
```diff
@@ -0,0 +1,235 @@
...
+
+renv_tests_setup_repos <- function(envir = parent.frame()) {
+
+  # generate our dummy repository
+  repopath <- file.path(tempdir(), "repos")
+  contrib <- file.path(repopath, "src/contrib")
+  ensure_directory(contrib)
+
+  # save current directory
+  owd <- getwd()
```
hadley, line 159: Could use `renv_scope_wd()` below?

tests/testthat/helper-setup.R
```diff
@@ -0,0 +1,235 @@
...
+    })
+
+  }
+
+}
+
+renv_tests_setup_repos <- function(envir = parent.frame()) {
+
+  # generate our dummy repository
+  repopath <- file.path(tempdir(), "repos")
```
hadley, line 154: `renv_scope_tempfile()`?

tests/testthat/setup.R
```diff
@@ -1,30 +1,3 @@
...
-        "* running the tests may mutate system renv state"
-      ),
-      rootdir,
-      tempdir
-    )
-  }
-})
-
+# Primarily for non-interactive use, e.g. with `devtools::test()`.
+renv_tests_setup(envir = testthat::teardown_env())
```
hadley, line 3: Maybe just remove this and call unconditionally in `helper-zzz.R`?

hadley (approved):

R/tests.R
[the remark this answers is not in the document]
kevinushey: I agree; it really feels like there should be a straightforward way to handle this in a single call to `file.path()`. 😞

tests/testthat/helper-setup.R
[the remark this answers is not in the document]
kevinushey: https://github.com/rstudio/renv/issues/1329

R/scope.R
kevinushey: Thanks! I retooled those tests.

tests/testthat/helper-scope.R
kevinushey: Good question... it might be better to turn this off, since this would better reflect real-life usage of `renv` as the sandbox won't get activated until a project is activated.

The BiocManager issues could also reflect real problems users might encounter.

tests/testthat/helper-scope.R
kevinushey: Thanks!

tests/testthat/helper-setup.R
kevinushey: Good idea!

tests/testthat/helper-setup.R
```diff
@@ -0,0 +1,235 @@
+
+# Code that needs to run once, before a suite of tests is run.
+# Here, "suite of tests" might also mean "a single test" interactively.
+renv_tests_setup <- function(envir = parent.frame()) {
+
```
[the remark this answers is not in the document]
kevinushey: I tried to separate "setup" functions (`renv_setup_*`) from "fixture" functions (`renv_scope_*`), so that the "setup" functions only ever need to be called once, currently in `helper-zzz.R` or `setup.R`.

So, by design, I don't think this function can be called multiple times? If it is, that's programmer error, I think.

tests/testthat/helper-setup.R
```diff
@@ -0,0 +1,235 @@
...
+  # ensure that attempts to restart are a no-op
+  if (renv_rstudio_available())
```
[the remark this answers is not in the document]
kevinushey: Thanks! I'll remove this piece.

tests/testthat/helper-setup.R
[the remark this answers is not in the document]
hadley: Ok, makes sense.

PR #1341, opened 2023-05-25
Demphasise equip

### LINKED ISSUE #108: Does `equip()` give sufficient tools to install packages like nlopt?
@kevinushey - As previously flagged to you I've been developing my own equip function of sorts for our corporate R package. I've been testing this fairly consistently for a good few months now on a windows server build machine to strong success coupled with packrat.

Worth linking this to you as there are some cpp libraries that I required to download that I don't think your equip function does. You can find the function [here](https://github.com/dfe-analytical-services/dfeR/blob/feature/rtools_pkgs/R/setup.R). I've been testing mine vs yours and if memory serves me right packages such as RCurl don't compile with equip. 

Another important thing to point out is that I've had to add the following (the RWINLIB bit) to the front of my local_libs within the last month:

```
LOCAL_LIBS = -L"$(RWINLIB)/lib$(R_ARCH)" -L"$(LOCAL_SOFT)/lib$(R_ARCH)" -L"$(LOCAL_SOFT)/lib"'

```

From what I can tell the [rwinlib](https://github.com/rwinlib) system dependencies that are downloaded by a number of packages have change structure in anticipation for the new rtools and this has caused older versions of some packages to not compile e.g. rgdal 1.3-6. This points it back to the right place and I don't think has any negative effects as will only point to something valid if such packages are downloaded. 

Leave it to you if you think any of this should make it into equip.

Cheers,

Adam

### LINKED ISSUE #346: Suggestion: Highlight renv::equip() in starter guides and documentation
Hi Kevin, thanks for the great package.

I had several issues with compiling older package versions from source. The problem is exacerbated by the fact that once renv::restore() is stopped, the entire 30min+ package installation process begins anew.

As time progresses most people will be working with older package versions which CRAN will not keep as binaries. Could be useful to know about renv::equip() from the get go.

### LINKED ISSUE #1264: Soft deprecate `equip()`?

hadley:
Fixes #1264. Closes #346. Closes #108.

@kevinushey I wasn't sure exactly what you were thinking here, but this should be a good start at making `equip()` less prominent in the docs. Let me know if there's anything specific you'd like me to say in the equip docs.

### DIFF
--- a/R/equip.R
+++ b/R/equip.R
@@ -2,11 +2,11 @@
 #' Install Required System Libraries
 #'
 #' Equip your system with libraries commonly-used during compilation of
-#' \R packages. Currently only supported on Windows.
+#' base and recommended \R packages.
 #'
 #' @return This function is normally called for its side effects.
 #' @export
-#'
+#' @keywords internal
 #' @examples
 #' \dontrun{
 #'
--- a/R/equip.R
+++ b/R/equip.R
@@ -1,8 +1,9 @@
 
-#' Install Required System Libraries
+#' Install required system libraries
 #'
 #' Equip your system with libraries commonly-used during compilation of
-#' base and recommended \R packages.
+#' base and recommended \R packages. This was previously useful with older
+#' versions of R on windows, but is no longer terribly helpful.
 #'
 #' @return This function is normally called for its side effects.
 #' @export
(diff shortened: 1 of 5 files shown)


kevinushey (approved):
LGTM!

kevinushey:
We could also consider a blurb in the `equip` documentation stating that this was mainly necessary for older versions of R on Windows, but this is probably enough.

PR #1346, opened 2023-05-25
Implement renv_version_maj_min() and use it

hadley:
This is a relatively simple modification of `renv_version_components()` that makes it useful in more places.

### DIFF
--- a/R/version.R
+++ b/R/version.R
@@ -68,18 +68,22 @@ renv_version_match <- function(versions, request) {
 
 }
 
-renv_version_components <- function(version, n) {
+renv_version_maj_min <- function(version) {
+
+  if (!inherits(version, "numeric_version")) {
+    version <- numeric_version(version)
+  }
 
   # split version into parts
-  parts <- unclass(numeric_version(version))[[1L]]
+  parts <- unclass(version)[[1L]]
 
   # extend parts to size of n
-  diff <- max(n) - length(parts)
+  diff <- 2 - length(parts)
   if (diff > 0)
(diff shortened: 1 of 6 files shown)


R/version.R
```diff
@@ -68,18 +68,22 @@ renv_version_match <- function(versions, request) {
 
 }
 
-renv_version_components <- function(version, n) {
+renv_version_maj_min <- function(version) {
+
+  if (!inherits(version, "numeric_version")) {
+    version <- numeric_version(version)
+  }
```
kevinushey, line 75:
```suggestion
  version <- as.numeric_version(version)
```

R/version.R
```diff
@@ -68,18 +68,22 @@ renv_version_match <- function(versions, request) {
 
 }
 
-renv_version_components <- function(version, n) {
```
kevinushey, line 71: Why not keep `renv_version_components` and just have a helper `renv_version_majmin()` based on that?

hadley: Because it wasn't used anywhere else? 

I think we have some difference in philosophy when a function is no longer used, because I just want to get rid of it 😄

kevinushey: Here's an important question: do you have a giant box full of old cables somewhere in your house? 🙂 

(you can probably guess what my answer is)

hadley: I have a box that is organised by cable type and whenever I get more than five of one type, I throw the worst out 🤣

kevinushey (approved):

PR #1349, opened 2023-05-25
Snapshot experiment

hadley:
Experiment using snapshots for `snapshot()` and `status()`. Uncovered one bug (use of `renv_envvar_exists()` in activate script), one small infelicity (`snapshot()` would prompt to activate even when in tests) and required a few switches from `!renv_tests_running()` to `renv_verbose()` (but maybe this is incorrect and those if statement should just be removed?).

### DIFF
--- a/R/difftime.R
+++ b/R/difftime.R
@@ -1,6 +1,10 @@
 
 renv_difftime_format <- function(time, digits = 2L) {
 
+  if (is_testing()) {
+    return ("XXXX seconds")
+  }
+
   units <- attr(time, "units") %||% ""
   if (units == "secs" && time < 0.1) {
     time  <- time * 1000
--- a/tests/testthat/helper-setup.R
+++ b/tests/testthat/helper-setup.R
@@ -69,6 +69,7 @@ renv_tests_setup_options <- function(envir = parent.frame()) {
     renv.bootstrap.quiet = TRUE,
     renv.config.user.library = FALSE,
     renv.config.sandbox.enabled = TRUE,
+    renv.consent = TRUE,
     restart = NULL,
     renv.config.install.transactional = FALSE,
     renv.tests.running = TRUE,
--- a/tests/testthat/helper-snapshot.R
+++ b/tests/testthat/helper-snapshot.R
@@ -12,5 +12,7 @@ expect_snapshot <- function(x, ...) {
 strip_dirs <- function(x) {
   x <- gsub(getwd(), "<wd>", x, fixed = TRUE)
   x <- gsub(renv_paths_cache(), "<cache>", x, fixed = TRUE)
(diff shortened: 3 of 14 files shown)


tests/testthat/helper-setup.R
```diff
@@ -69,6 +69,7 @@ renv_tests_setup_options <- function(envir = parent.frame()) {
     renv.bootstrap.quiet = TRUE,
     renv.config.user.library = FALSE,
     renv.config.sandbox.enabled = TRUE,
+    renv.consent = TRUE,
```
kevinushey, line 72: Do you think we should add a check in https://github.com/rstudio/renv/blob/c972ddd2dc0d9aa83bb2b76411011e8782b1aac6/R/consent.R#L81-L88 for "tests running"?

kevinushey: Oh, I guess you already did (but we still need something here?)

kevinushey (approved):

tests/testthat/helper-setup.R
hadley: We could also start something like `rlang::is_interactive()`:

```R
is_interactive <- function() {
  opt <- getOption("rlang_interactive")
  if (!is.null(opt)) {
    return(opt)
  }
  
  interactive()
}
```

testthat sets `rlang_interactive` to `FALSE` by default, but you can still easily override in tests if you want to.

tests/testthat/helper-snapshot.R
```diff
@@ -0,0 +1,22 @@
...
+  # This shouldn't be so hard
+  eval.parent(
+    substitute(
+      testthat::expect_snapshot(x, ..., transform = strip_dirs)
+    )
+  )
+}
+
+strip_dirs <- function(x) {
+  x <- gsub(getwd(), "<wd>", x, fixed = TRUE)
```
kevinushey, line 13: The amount of stuff accumulating here makes me wonder if there might be a better alternative approach.

tests/testthat/helper-setup.R
hadley: Or to make it renv specific, it could be 

```R
is_interactive <- function() {
  !renv_tests_running() && interactive()
}
```

R/difftime.R
```diff
@@ -1,6 +1,10 @@
 
 renv_difftime_format <- function(time, digits = 2L) {
 
+  if (is_testing()) {
```
hadley, line 4: Are you ok with leaving this as is until I do a patch release of testthat? (Hopefully early next week)

kevinushey: I'm okay with that; should we add an issue or some other reminder to do so?

hadley: I'll make a note on my personal todo list.

PR #1350, opened 2023-05-26
Standardize output

hadley:
This has probably become too big to review easily, so it's probably best taken as discussion starter . Overall it ensures that we use `writef()` and friends for all user facing output, and ensure all output goes to `stdout()`, not `stderr()`. (Let me know if you need more justification for why I think this is the right approach).

Separately I think we should consider the names of `printf` and `writef`, as it's not obvious from their names how they differ (i.e. handling of vector inputs and whether or not there's a trailing line break).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: renv
 Type: Package
 Title: Project Environments
-Version: 0.17.3-67
+Version: 0.17.3-68
 Authors@R: c(
     person("Kevin", "Ushey", role = c("aut", "cre"), email = "kevin@rstudio.com"),
     person("Posit Software, PBC", role = c("cph", "fnd"))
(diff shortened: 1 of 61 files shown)


kevinushey:
> Separately I think we should consider the names of printf and writef, as it's not obvious from their names how they differ (i.e. handling of vector inputs and whether or not there's a trailing line break).

I'm open to alternatives here. Would you prefer a single function like `catf` and some argument that controls whether or not we append a line break?

kevinushey (approved):
Thanks; I'm on board.

hadley:
One thing that's worth thinking about is that there are maybe two uses of `writef()`:

```R
writef(c(title, lines, ""))
# vs
writef("Project path: %s", renv_path_pretty(project))
```

i.e. sometimes you have a vector of lines and sometimes you have a single line. And when you have multiple lines it doesn't make sense to have a format string.

Anyway, I'll merge this and we can iterate on names separately.

PR #1362, opened 2023-05-30
Polish hydrate docs

hadley:


R/hydrate.R
```diff
@@ -1,41 +1,24 @@
 
-#' Hydrate a Project
+#' Copy packages from user libraries to a project library
 #'
-#' Discover the \R packages used within a project, and then install those
-#' packages into the active library. This effectively allows you to fork the
-#' state of your default \R libraries for use within a project library.
+#' @description
+#' Copy/link the packages currently used by the project from a user library
+#' to the project library. This synchronises a project library with you have
```
ateucher, line 6:
```suggestion
#' to the project library. This synchronises a project library what with you have
```

ateucher (approved):
LGTM! I Like the added context as to when and how you would want to use it.

kevinushey:
The main reason I had an "exercise with caution" warning in `hydrate()` is that, by default, it doesn't touch packages that are already installed in the project library. In other words, it's possible that the project library is self-consistent; the default library paths are self-consistent, but calling `renv::hydrate()` could give you an inconsistent mix of those packages.

This is safe to do so when initializing a new project with a new project library, but is a bit risky if calling `renv::hydrate()` with a project library that already exists. So I feel like that should be be emphasized.

R/hydrate.R
```diff
@@ -1,41 +1,24 @@
...
+#' installed globally.
 #'
-#' It may occasionally be useful to use `renv::hydrate()` to update the packages
-#' used within a project that has already been initialized. However, be aware
-#' that it's possible that the packages pulled in may not actually be compatible
-#' with the packages already installed in the project library, so you should
-#' exercise caution when doing so.
+#' You generally don't need to call `hydrate()` directly as [init()] will call
+#' it for you. However, you may want to call it directly if you want to update
```
kevinushey, line 10: I'd go a bit further and say that this isn't recommended or is experts-only, given the caveats of mixing different packages from different libraries.

kevinushey (approved):

hadley:
Oooh, I didn't get that before.

PR #1367, opened 2023-05-30
Report on installed but unused packages

### LINKED ISSUE #749: Consider chattier message on status / snapshot
If you are using a renv mode that only snapshots identified dependencies and not the entire active library, it can be hard to understand why some packages are being excluded. 

While this behavior is often what you want, users have to be proactive in finding this very helpful FAQ: https://rstudio.github.io/renv/articles/faq.html#why-isn-t-my-package-being-snapshotted-into-the-lockfile-

It may help if renv::status and renv::snapshot were a little chattier when the snapshot type is not all. For example, a status message might say: 

8 packages are installed in the library but will not be snapshotted because don't appear to be used. See renv::dependencies to see what packages appear to be in use. Run renv::settings$snapshot.type("all") to tell renv to track all installed packages.

This is a long message and may get annoying overtime... a tradeoff to consider. 

FWIW I really like the renv::dependencies default, I sometimes forget about it though! This ticket was prompted by the parsnip package which has dependencies that are usually not identified by renv::dependencies.

hadley:
Fixes #749

### DIFF
--- a/R/status.R
+++ b/R/status.R
@@ -177,9 +177,6 @@ renv_status_check_synchronized <- function(project,
                                            library,
                                            packages)
 {
-  # flag set to FALSE if any of the below checks report out-of-sync
-  ok <- TRUE
-
   # extract record components
   lockfile <- renv_lockfile_records(lockfile)
   library  <- renv_lockfile_records(library)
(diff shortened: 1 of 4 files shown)


R/status.R
```diff
@@ -283,8 +283,19 @@ renv_status_check_synchronized <- function(project,
...
+  # installed/*/not used -------------------------------------------------------
+  records <- library %>%
+    exclude(packages) %>%
+    exclude(renv_packages_recommended())
+
+  if (length(records)) {
+    renv_pretty_print_records(
+      records,
+      preamble = "The following packages are installed but not used."
+    )
```
hadley, line 295: There's no postamble, because there at least four things you might want to do:

* Nothing, because you're using the packages interactively and they don't need to be inclued.
* You might want to run `clean()` to get rid of them
* You might want to add them to `_deps.R` to track them
* You might want to change the snapshot type

kevinushey:
IMHO this is of limited utility; `renv::status()` should tell you whether the library + lockfile are in a consistent state, and it's likely that users will have a variety of other packages installed in the project that they get "nagged" about every time they call `status()`.

At least by default, I think `renv::status()` should only provide feedback to the user if there's some kind of conflict, and if there's something they can (or should) do to resolve the conflict.

PR #1368, opened 2023-05-30
Add a clean argument to `deactivate()`

hadley:


NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `renv::deactivate()` gains a `clean` argument: when `TRUE` it will delete
+  all renv files/directories, leaving the project the way it was found.
```
kevinushey, line 5: Should we emphasize that this is only project-specific files and directories? (The global renv state / cache remains untouched)

kevinushey: Maybe not necessary for NEWS but perhaps worth emphasizing in the `?deactivate` docs.

kevinushey (approved):

NEWS.md
kevinushey: I guess that's already pretty clear from the docs there so I'll just resolve this!

PR #1370, opened 2023-05-31
Optionally install packages during `snapshot()`

### LINKED ISSUE #1198: Confusing prompt from `snapshot()`
Here's a (redacted) exchange received with {renv} v0.17.3 when running `snapshot()`:
```r
> renv::snapshot()
The following required packages are not installed:

        some, redacted, in_development,
        packages

Packages must first be installed before `renv` can snapshot them.
Consider installing these packages using `renv::install()`.
Use `renv::dependencies()` to see where this package is used in your project.

Do you want to proceed? [y/N]: 
```
It's not clear what "proceed" means here. `status()` gives effectively the same message but without the prompt to "proceed".

When I select `y`, `snapshot()` reports that `The lockfile is already up to date`, but it can be a bit unsettling to select `y` to verify (to oneself) that last fact.

hadley:
Fixes #1198


R/lockfile.R
```diff
@@ -174,17 +174,27 @@ renv_lockfile_create <- function(project,
...
-
-    renv_snapshot_libpaths(libpaths = libpaths,
-                           project  = project) %>%
-
-    renv_snapshot_filter(project  = project,
-                         type     = type,
-                         packages = packages,
-                         exclude  = exclude) %>%
+  repeat {
+    withRestarts(
```
hadley, line 178: I don't love the use of restarts, but this seemed like the easiest way to recompute the lockfile records, given the distance in the call sites.

R/pretty.R
```diff
@@ -35,6 +35,36 @@ renv_pretty_print <- function(values,
   writef(text)
 }
 
+renv_pretty_bullets <- function(values,
```
hadley, line 38: I'll probably pull this out into a separate PR. It's possible we can use it everywhere that `renv_pretty_print()` is currently used.

R/snapshot.R
```diff
@@ -986,15 +989,29 @@ renv_snapshot_filter_report_missing <- function(missing, type) {
...
-  TRUE
+  choices <- c(
+    "Install the packages, then snapshot",
+    "Snapshot, just using the currently installed packages",
+    "Cancel"
+  )
+  choice <- tryCatch(
+    menu(choices, graphics = FALSE, title = "What do you want to do?"),
+    interrupt = function(cnd) 0
+  )
```
hadley, line 1006: This also needs a bit of work to make it testable, and it'll be worth also using where `utils::select.list()` currently is.

R/snapshot.R
```diff
@@ -971,6 +971,9 @@ renv_snapshot_filter_impl <- function(project, records, packages, type) {
 }
 
 renv_snapshot_filter_report_missing <- function(missing, type) {
+  if (!interactive() || !renv_verbose()) {
```
hadley, line 974: This logic is wrong — still need to show the message even if not interactive.

R/snapshot.R
kevinushey: What are the pros and cons of `menu()` vs `select.list()`?

R/pretty.R
kevinushey: It would also be a good opportunity to revisit the `wrap` argument for `renv_pretty_print`; I think we almost always use `wrap = FALSE` so that is probably the better default.

R/lockfile.R
```diff
@@ -174,17 +174,27 @@ renv_lockfile_create <- function(project,
...
-  renv_lockfile_records(lockfile) <-
-
-    renv_snapshot_libpaths(libpaths = libpaths,
-                           project  = project) %>%
-
-    renv_snapshot_filter(project  = project,
-                         type     = type,
-                         packages = packages,
-                         exclude  = exclude) %>%
+  repeat {
```
kevinushey, line 177: Can we add a comment here indicating the intention here is to allow the restart below to recompute the lockfile?

kevinushey (commented):
LGTM; I can't think of a better solution other than finding a way to pull out `renv_snapshot_filter_report_missing` but that seems tough to untangle.

R/snapshot.R
hadley: Lets discuss in #1372, but in brief, I find the returned value to be easier to work with.

kevinushey: 👍

tests/testthat/_snaps/snapshot.md
```diff
@@ -126,3 +154,40 @@
...
+      
+      What do you want to do?
+      
+      1: Snapshot, just using the currently installed packages
+      2: Install the packages, then snapshot
+      3: Cancel
+      
+      Selection: 2
+      
+      Retrieving 'file://<tempdir>/renv-repos-519265f184d7/src/contrib/egg_1.0.0.tar.gz' ...
```
hadley, line 177: Need to fix this path

R/lockfile.R
```diff
@@ -174,17 +174,30 @@ renv_lockfile_create <- function(project,
...
+          records  = records,
+          type     = type,
+          packages = packages,
+          exclude  = exclude
+        )
+        # skipped if recomputeRecords restart is used
+        break
+      },
+      recomputeRecords = function() {
+        renv_dynamic_reset()
```
kevinushey, line 194: I wonder if we should also set some value to ensure we don't prompt the user again, just in case? (To avoid potentially recursively prompting the user.)

This shouldn't happen if `install()` works correctly, but it feels like a good idea to have a safeguard given the structure of the code here.

R/snapshot.R
```diff
@@ -976,21 +975,31 @@ renv_snapshot_filter_report_missing <- function(missing, type) {
...
     "Packages must first be installed before renv can snapshot them.",
-    "Consider installing these packages using `renv::install()`.",
     if (type %in% "explicit")
       "If these packages are no longer required, consider removing them from your DESCRIPTION file."
     else
       "Use `renv::dependencies()` to see where this package is used in your project."
   )
   renv_pretty_print(
-    values = csort(unique(missing)),
+    values = unique(missing),
```
kevinushey, line 984: Do we still want to sort here (not csort)?

R/lockfile.R
hadley: Yeah, maybe the `repeat` is just the wrong structure all together. Let me think about it.

R/snapshot.R
hadley: Ooops yes. And since this is a common need (and it'd be nice to keep all uses of `sort()` in pretty printing), I've made a note to add a sort argument to `renv_pretty_print()`.

R/lockfile.R
```diff
@@ -174,17 +174,36 @@ renv_lockfile_create <- function(project,
...
+      NULL
+    }
+  )
 
-    renv_snapshot_filter(project  = project,
-                         type     = type,
-                         packages = packages,
-                         exclude  = exclude) %>%
+  # We re-computing and re-filtering records once, if requested by user
+  if (is.null(records)) {
```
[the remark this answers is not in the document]
kevinushey: Much better 😄

kevinushey (approved):

hadley:
@kevinushey would you mind taking a look in your windows VM to figure out what's wrong with https://github.com/rstudio/renv/actions/runs/5164075054/jobs/9302761838?pr=1370#step:6:225?

kevinushey:
Looks like it might be due to path separators?

```
> tempdir()
[1] "C:\\temp\\RtmpmKiprC"
> getwd()
[1] "C:/temp/RtmpmKiprC/renv-test-36fc460a4fe2"
```

Maybe https://github.com/rstudio/renv/blob/055735df0c132161e46cfaaaf4c15bd6d5402da8/tests/testthat/helper-snapshot.R#L13 should be a regex or something accepting either kind of separator? Or just normalize `getwd()` before checking?

hadley:
I have tried like 7 different ways of normalizing the path, but it's totally possible I screwed it up every time 😄

hadley:
I'm relieved that this wasn't simple 😆

PR #1372, opened 2023-05-31
Provide a 3 choice menu when prompting to activate

hadley:
Follow up to #1228


R/activate.R
```diff
@@ -156,22 +156,27 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
-    "Please see `?renv::activate` for more details.",
-    ""
+  renv_activate_prompt_impl(project)
+
+
+}
+
+renv_activate_prompt_impl <- function(project = NULL) {
+  choices <- c(
+    "Activate the project, setting up a project library",
```
kevinushey, line 166:
```suggestion
    "Activate the project and use the project library.",
```

R/activate.R
```diff
@@ -156,22 +156,27 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+    "Continue, using the system library",
```
kevinushey, line 167:
```suggestion
    "Do not activate the project; use the current library paths.",
```

R/activate.R
```diff
@@ -156,22 +156,27 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+    "Cancel"
```
kevinushey, line 168: Do you think this is necessary? Since `menu()` will tell you how to cancel:

```
> menu("Testing")

1: Testing

Selection: 
Enter an item from the menu, or 0 to exit
Selection: 0
[1] 0
```

although I guess it's not displayed when the menu is first displayed.

kevinushey: (Is there prior precedent for this from other R packages?)

R/activate.R
```diff
@@ -156,22 +156,27 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
-    "This project has not yet been activated.",
-    "Activating this project will ensure the project library is used when %s() is called.",
-    "Please see `?renv::activate` for more details.",
-    ""
+  renv_activate_prompt_impl(project)
+
+
+}
+
+renv_activate_prompt_impl <- function(project = NULL) {
```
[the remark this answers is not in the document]
kevinushey: I think I prefer this version, since it's obvious from the code which answer is associated with which choice.

R/activate.R
```diff
@@ -156,22 +156,27 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+    "Cancel"
   )
-  writef(fmt, action)
 
-  fmt <- "Would you like to activate this project before %s() is called?"
-  question <- sprintf(fmt, action)
-  response <- ask(question, default = TRUE)
-  if (!response)
-    return(FALSE)
+  writef("This project has not yet been activated.")
```
kevinushey, line 171: I feel like we need a single-line blurb explaining why this prompt is being shown, like "It looks like you're working within an renv project, but that project has not been loaded yet. What would you like to do?"

R/activate.R
```diff
@@ -156,22 +156,27 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+  choice <- menu(choices, "What do you want to do?", default = 2)
```
kevinushey, line 172:
```suggestion
  choice <- menu(choices, "What would you like to do?", default = 2)
```

Also, does menu have a 'default' argument?

kevinushey:
```suggestion
  choice <- menu(choices, "How would you like to proceed?", default = 2)
```

Just word-smithing; not sure if one of these is better than the other.

R/activate.R
kevinushey:
```suggestion
    "Cancel."
```

R/activate.R
kevinushey: Oh, I see you defined your own version of menu below.

R/activate.R
hadley: Yeah, I think the `0` option is pretty hard to discover. There's some related art in rsconnect, but that's because I solved a similar problem a few of weeks ago.

R/activate.R
```diff
@@ -156,22 +156,27 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
-    "Activating this project will ensure the project library is used when %s() is called.",
-    "Please see `?renv::activate` for more details.",
-    ""
+  renv_activate_prompt_impl(project)
+
+
+}
+
+renv_activate_prompt_impl <- function(project = NULL) {
+  choices <- c(
```
kevinushey, line 165: We might also want some way to cache the user's response per-session so we don't bug them multiple times (maybe a separate PR)

R/activate.R
```diff
@@ -156,22 +156,26 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+  )
+  choices <- c(
+    activate = "Activate the project and use the project library.",
+    continue = "Do not activate the project and use the current library paths.",
+    cancel = "Cancel and resolve the problem another way."
+  )
+
+  choice <- menu(choices, title, default = 2)
+  switch(choice,
+    activate = {activate(project = project); TRUE},
```
kevinushey, line 177:
```suggestion
    activate = { activate(project = project); TRUE },
```

R/utils.R
```diff
@@ -163,6 +163,46 @@ proceed <- function(default = TRUE) {
...
+  }
+
+  if (!interactive()) {
+    writef(c("Not interactive. Will:", choices[[default]]))
+    return(default)
+  }
+
+  tryCatch(
+    idx <- utils::menu(choices, paste(title, collapse = "\n"), graphics = FALSE),
+    interrupt = function(cnd) 0L
```
kevinushey, line 196:
```suggestion
    interrupt = function(cnd) "cancel"
```

to simplify code below?

R/utils.R
```diff
@@ -163,6 +163,46 @@ proceed <- function(default = TRUE) {
...
+  )
+  if (idx == 0L) {
+    "cancel"
+  } else {
+    names(choices)[selected]
```
kevinushey, line 201: You set `idx` above but use `selected` here?

R/utils.R
```diff
@@ -163,6 +163,46 @@ proceed <- function(default = TRUE) {
...
+    return(names(choices)[selected])
+  }
+
+  if (!interactive()) {
+    writef(c("Not interactive. Will:", choices[[default]]))
+    return(default)
+  }
+
+  tryCatch(
+    idx <- utils::menu(choices, paste(title, collapse = "\n"), graphics = FALSE),
```
kevinushey, line 195:
```suggestion
  idx <- tryCatch(
    utils::menu(choices, paste(title, collapse = "\n"), graphics = FALSE),
```

since `tryCatch()` is always producing a usable value?

R/utils.R
hadley: Still `menu()` might return 0L, so it doesn't help that much.

R/activate.R
```diff
@@ -156,22 +156,26 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
-  if (!response)
-    return(FALSE)
 
-  activate(project = project)
-  return(TRUE)
+}
 
+renv_activate_prompt_impl <- function(action, project = NULL) {
+  title <- c(
+    sprintf("It looks like you've called renv::%s() in a non-renv project.", action),
```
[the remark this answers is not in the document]
hadley: Hmmmm, I guess I was a bit confused by the focus of the function on "activate", rather than "load". I wonder if it's worth distinguishing between the two cases (i.e. non-renv vs not-loaded renv) or just writing something generic. Previously we had:

> This project has not yet been activated.

So maybe 

> It looks like you've called renv::%s() in a project that hasn't been activated yet

?

R/activate.R
```diff
@@ -156,22 +156,26 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+    "How would you like to proceed?"
+  )
+  choices <- c(
+    activate = "Activate the project and use the project library.",
+    continue = "Do not activate the project and use the current library paths.",
+    cancel = "Cancel and resolve the problem another way."
```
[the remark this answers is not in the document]
hadley: Does "issue" sound better than "problem"?  Or "situation"? 

My goal was to make it more clear that by cancelling you're opting to do something other than the available choices.

R/utils.R
hadley: Oops yes.

R/utils.R
hadley: Good catch. I'll fix this by hand.

R/utils.R
kevinushey: Ah right -- this is our wrapper over `utils::menu()` so we need to transform the numeric indices into the actual selected values.

R/activate.R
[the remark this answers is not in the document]
kevinushey: Yeah, I like that.

For extra context, the main reason why I imagined this prompt being necessary: imagine you share an `renv` project with someone else, and so you just share the project + the lockfile. Your collaborator starts working and runs `renv::restore()`.

This way, we can see they probably intended to load the project first so they could use a project library, and also that we can "activate" the project so the project library is loaded by default in the future.

R/activate.R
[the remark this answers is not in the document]
kevinushey: I like "situation". A plain "Cancel the current operation." might also be fine.

I guess my question is, what are the "other way"s a user could resolve a situation like the above? Intentionally call `renv::activate()` themselves beforehand? Use `renv::restore(library = "path/to/custom/library")`?

R/activate.R
```diff
@@ -156,22 +156,26 @@ renv_activate_prompt <- function(action, library, prompt, project) {
...
+  )
+
+  choice <- menu(choices, title, default = 2)
```
kevinushey, line 175: Do we accept named defaults here, for when "choices" is a named character vector?

R/activate.R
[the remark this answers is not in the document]
hadley: Ah, that makes sense.

R/activate.R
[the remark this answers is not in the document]
hadley: I guess the option I'm thinking of is "oops, I've called `restore()` in the wrong directory and I need to change my working directory"

R/activate.R
hadley: Yup, and that's easier to reason about.

R/activate.R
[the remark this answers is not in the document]
hadley: How about "Cancel, and resolve the situation on your own." — I think that better conveys you're opting to take control.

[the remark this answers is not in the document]
kevinushey: > I guess the option I'm thinking of is "oops, I've called restore() in the wrong directory and I need to change my working directory"

Oh, that's a good point!

> How about "Cancel, and resolve the situation on your own." — I think that better conveys you're opting to take control.

I like it!

Issue #1359, opened 2023-05-30
problem restoring package from r-universe


hansvancalster:
I'm trying to restore a package from `renv.lock` which is hosted on a r-universe repo. 
It fails with `Error in read.dcf(file.path(pkgname, "DESCRIPTION"), c("Package", "Type")) `.
The RemoteSha in the `renv.lock` refers to an older (development) version of  the current lintr 3.0.2.9000 on that repo. Could that be related to this problem?
 
Here is a reprex (the .Rprofile has pkgType = "both" as option, which I needed to override an organisation specific Rprofile.site ):

*Local `.Rprofile` detected at `C:\R\temp\prtclsrc\.Rprofile`*

``` r
getOption("repos")
#>                                        CRAN 
#>               "https://cloud.r-project.org" 
#>                                        INLA 
#> "https://inla.r-inla-download.org/R/stable" 
#>                                        inbo 
#>               "https://inbo.r-universe.dev"
x <- jsonlite::read_json("renv.lock")
x$Packages$lintr
#> $Package
#> [1] "lintr"
#> 
#> $Version
#> [1] "3.0.2.9000"
#> 
#> $Source
#> [1] "Repository"
#> 
#> $Repository
#> [1] "https://inbo.r-universe.dev"
#> 
#> $RemoteUrl
#> [1] "https://github.com/r-lib/lintr"
#> 
#> $RemoteRef
#> [1] "HEAD"
#> 
#> $RemoteSha
#> [1] "3a8d4957ce33a7b0613ee917d887209118cf63d2"
#> 
#> $Requirements
#> $Requirements[[1]]
#> [1] "R"
#> 
#> $Requirements[[2]]
#> [1] "backports"
#> 
#> $Requirements[[3]]
#> [1] "codetools"
#> 
#> $Requirements[[4]]
#> [1] "cyclocomp"
#> 
#> $Requirements[[5]]
#> [1] "digest"
#> 
#> $Requirements[[6]]
#> [1] "glue"
#> 
#> $Requirements[[7]]
#> [1] "knitr"
#> 
#> $Requirements[[8]]
#> [1] "rex"
#> 
#> $Requirements[[9]]
#> [1] "stats"
#> 
#> $Requirements[[10]]
#> [1] "utils"
#> 
#> $Requirements[[11]]
#> [1] "xml2"
#> 
#> $Requirements[[12]]
#> [1] "xmlparsedata"
#> 
#> 
#> $Hash
[...]

PR #1373, opened 2023-05-31
handle <pkg>_<sha>.zip installation

kevinushey:
Closes https://github.com/rstudio/renv/issues/1359.

### DIFF
--- a/R/r.R
+++ b/R/r.R
@@ -149,9 +149,8 @@ r_cmd_install <- function(package, path, ...) {
   path <- renv_path_normalize(path, winslash = "/", mustWork = TRUE)
 
   # unpack source packages in zip archives
-  iszip <- renv_archive_type(path) %in% "zip"
   unpack <-
-    iszip &&
+    renv_archive_type(path) %in% "zip" &&
     renv_package_type(path) %in% "source"
 
   if (unpack) {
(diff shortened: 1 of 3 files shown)


R/r.R
```diff
@@ -149,15 +149,39 @@ r_cmd_install <- function(package, path, ...) {
...
+    # build expected pattern
+    fmt <- "^%s(?:_%s)?\\.zip$"
+    regexps <- .standard_regexps()
+    pattern <- sprintf(fmt, regexps$valid_package_name, regexps$valid_package_version)
+    matches <- grepl(pattern, basename(path), perl = TRUE)
+
+    # if we don't have a match, then we'll have to install it ourselves
+    if (!matches) {
+      dir <- renv_scope_tempfile("r-cmd-install-")
+      ensure_directory(dir)
```
hadley, line 177: What do you think of writing these temporary directories like this?

```suggestion
      dir <- ensure_directory(renv_scope_tempfile("r-cmd-install-"))
```

R/r.R
```diff
@@ -149,15 +149,39 @@ r_cmd_install <- function(package, path, ...) {
...
+  if (renv_platform_windows() && iszip) {
+
+    # build expected pattern
+    fmt <- "^%s(?:_%s)?\\.zip$"
+    regexps <- .standard_regexps()
+    pattern <- sprintf(fmt, regexps$valid_package_name, regexps$valid_package_version)
+    matches <- grepl(pattern, basename(path), perl = TRUE)
+
+    # if we don't have a match, then we'll have to install it ourselves
+    if (!matches) {
```
hadley, line 175: I wonder if it would be safer to just rename to `package_1.0.0.zip`?

hadley (approved):

R/r.R
kevinushey: I prefer to avoid nesting the scope helpers; IMHO the code is easier to read in its original form.

R/r.R
```diff
@@ -158,6 +158,30 @@ r_cmd_install <- function(package, path, ...) {
...
+    renv_file_type(path) == "file" &&
+    renv_archive_type(path) %in% "zip" &&
+    renv_package_type(path) %in% "binary"
+
+  if (unpack) {
+
+    # unpack into temporary directory in library path
+    library <- renv_libpaths_active()
+    exdir <- renv_scope_tempfile(package, tmpdir = library)
+    unzip(path, exdir = exdir)
```
hadley, line 173: It just occurred to me, but can we just install from that directory?

kevinushey: You mean unpack into the library path directly? We could, but this is a bit safer in case multiple processes are trying to install into the same library at the same time. (really, we should be using locking but I got lazy).

hadley: No, I mean call `install.packages()` (or the `R CMD install` equivalent) on the unzipped directory, instead of the zipped file.

kevinushey: Ohh, yes! I'll try that.

PR #1374, opened 2023-05-31
Require user to confirm installation

### LINKED ISSUE #587: prompt=TRUE does nothing
`renv::install(..., prompt=TRUE)` does nothing. `renv` just installs all packages without any confirmation. I'm surprised "install regardless of the version changes to existing packages" is the default in R. That's the opposite of conda, which defaults to prompting the user after listing all proposed changes to the existing environment. It would be great to have the same with renv. 

```
> sessionInfo()
R version 4.0.3 (2020-10-10)
Platform: x86_64-conda-linux-gnu (64-bit)
Running under: Ubuntu 18.04.5 LTS

Matrix products: default
BLAS/LAPACK: /ebio/abt3_projects/Georg_animal_feces/envs/function/lib/libopenblasp-r0.3.10.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C
 [9] LC_ADDRESS=C               LC_TELEPHONE=C
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] renv_0.12.3

loaded via a namespace (and not attached):
[1] compiler_4.0.3 tools_4.0.3
```

hadley:
Fixes #587

### DIFF
--- a/R/patch.R
+++ b/R/patch.R
@@ -73,7 +73,10 @@ renv_patch_tar <- function() {
 
   # TAR appears to be set but invalid; override it
   # and warn the user
-  newtar <- Sys.which("tar") %""% "internal"
+  newtar <- Sys.which("tar")
+  if (!nzchar(newtar))
+    newtar <- "internal"
+
   Sys.setenv(TAR = newtar)
 
   # report to the user
--- a/R/path.R
+++ b/R/path.R
@@ -12,10 +12,7 @@ renv_path_absolute <- function(path) {
 
 renv_path_aliased <- function(path) {
 
-  home <-
-    Sys.getenv("HOME") %""%
-    Sys.getenv("R_USER")
-
+  home <- Sys.getenv("HOME", unset = Sys.getenv("R_USER"))
   if (!nzchar(home))
     return(path)
 
--- a/R/r.R
+++ b/R/r.R
@@ -148,17 +148,41 @@ r_cmd_install <- function(package, path, ...) {
   # normalize path to package
   path <- renv_path_normalize(path, winslash = "/", mustWork = TRUE)
 
-  # unpack .zip archives before install
+  # unpack .zip source archives before install
   # https://github.com/rstudio/renv/issues/1359
+  ftype <- renv_file_type(path)
+  atype <- renv_archive_type(path)
+  ptype <- renv_package_type(path)
+
   unpack <-
-    renv_file_type(path) == "file" &&
-    renv_archive_type(path) %in% "zip"
+    ftype == "file" &&
+    atype == "zip" &&
+    ptype == "source"
 
   if (unpack) {
     newpath <- renv_package_unpack(package, path, force = TRUE)
     if (!identical(newpath, path)) {
       path <- newpath
       defer(unlink(path, recursive = TRUE))
+    }
+  }
+
+  # rename binary .zip files if necessary
+  rename <-
+    ftype == "file" &&
+    atype == "zip" &&
+    ptype == "binary"
+
+  if (rename) {
+    regexps <- .standard_regexps()
+    fmt <- "^%s(?:_%s)?\\.zip$"
+    pattern <- sprintf(fmt, regexps$valid_package_name, regexps$valid_package_version)
+    if (!grepl(pattern, basename(path), perl = TRUE)) {
+      dir <- renv_scope_tempfile(package)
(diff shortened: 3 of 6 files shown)


tests/testthat/_snaps/install.md
```diff
@@ -13,3 +19,27 @@
...
+      Downloading 1 package(s) and their dependencies
+      Done
+      
+      The following package(s) will be installed into the library
+      - bread     [1.0.0]
+      - breakfast [1.0.0]
+      - oatmeal   [1.0.0]
+      - toast     [1.0.0]
+      
+      Installing oatmeal [1.0.0] ...
```
hadley, line 36: Given that the versions are now reported above, what do you think about dropping from the individual status lines?

tests/testthat/_snaps/install.md
```diff
@@ -3,6 +3,12 @@
     Code
       install("bread@0.1.0")
     Output
+      Downloading 1 package(s) and their dependencies
+      Done
+      
+      The following package(s) will be installed into the library
```
hadley, line 9: Drop "into the library" ?

R/install.R
```diff
@@ -792,3 +796,10 @@ renv_install_remotes_update <- function(records, project) {
   records
 
 }
+
+renv_install_report <- function(records) {
+  renv_pretty_print_records(
+    records,
+    preamble = "The following package(s) will be installed into the library"
```
kevinushey, line 803:
```suggestion
    preamble = "The following package(s) will be installed into the library:"
```

R/retrieve.R
```diff
@@ -28,10 +28,12 @@ retrieve <- function(packages) {
   }
   renv_scope_options(HTTPUserAgent = agent)
 
+  writef("Downloading %s package(s) and their dependencies", length(packages))
```
kevinushey, line 31:
```suggestion
  writef("Downloading %s and their dependencies.", nplural("package", length(packages)))
```

tests/testthat/_snaps/install.md
kevinushey: Dropping the versions, I assume you mean? That sounds good to me.

tests/testthat/_snaps/install.md
kevinushey: Yeah, let's drop it. Maybe:

```
The following packages will be installed:

- abc
- xyz

Packages will be installed into ~/path/to/library.
```

Reporting the library path might be useful just to confirm the library path is what the user expects?

kevinushey: Or maybe we just always print this and remove the tests-running conditional: https://github.com/rstudio/renv/blob/6ad88516825f92f6d0fd919dfd8395a1d07462ec/R/install.R#L197-L205

R/install.R
kevinushey: Might also be a good opportunity to move from "package(s)" to just checking the number of packages to be installed?

kevinushey (approved):

R/retrieve.R
hadley: That doesn't quite work because in the one package case it should really be "Downloading 1 package and its dependencies".  We could use `nggettext()` here or create some helper:

```R
writef(
 ngettext(
   length(packages),
   "Downloading %s package and its dependencies",
   "Downloading %s packages and their dependenices",
 ),
 length(packages)
)
```

But that's going to be a lot of work.

tests/testthat/_snaps/install.md
hadley: Ok, I'll propose some changes in separate PR.

R/retrieve.R
kevinushey: Good point. Let's leave it alone.

tests/testthat/_snaps/install.md
hadley: Moved that into the preamble, and removed tweaked other test specific code.

PR #1375, opened 2023-05-31
Eliminate emitter and wrap arguments

hadley:
* All calls either do set `wrap = FALSE` or (IMO) should set `wrap = FALSE`
* Use `-` instead of tab to make a bulleted list
* The emitter argument didn't have any affect


kevinushey (approved):
LGTM, thanks!

PR #1378, opened 2023-06-01
Experiment: make `renv_pretty_print()` more compact

hadley:
I'm not sure how I feel about this. I like that it makes each call feel more like a related block (reserving empty lines to separate outputs) but maybe it becomes a bit too dense?


kevinushey (approved):
I don't feel too strongly either way, but I feel like the extra breathing room is useful when the package list is large. I'll let you decide if you want to merge and if we change our minds later we can roll it back.

PR #1379, opened 2023-06-01
Compact installation logging

hadley:
* Drop version since now shown above
* Put each package on one line
* Align package name and verb
* Only show timing if greater than a second

The tests only cover a subset of the installation modes, so I also tested interactively with

```R
with_renv({
  init()

  writeLines(c('library("rlang")', "library(memoise)"), "_deps.R")
  install()
  install(c("rlang", "memoise", "fastmap"), rebuild = TRUE)

  renv_scope_options(renv.config.install.build = TRUE)
  install("r-lib/cli", rebuild = TRUE)
})
```

Initial install looks like this:

```
Installing rlang   ... OK [installed binary]
Caching rlang      ... OK [moved]
Installing fastmap ... OK [installed binary]
Caching fastmap    ... OK [moved]
Installing memoise ... OK [installed binary]
Caching memoise    ... OK [moved]
```

Installing from cache looks like this:

```
Installing rlang   ... OK [linked from cache]
Installing fastmap ... OK [linked from cache]
Installing cachem  ... OK [linked from cache]
Installing memoise ... OK [linked from cache]
```

And building from source looks like this:

```
Building cli   ... OK [from source]
Installing cli ... OK [built from source in 7.2 seconds]
Caching cli    ... OK [moved]
```

IMO using a single line for each action and carefully aligning the `...` makes the results much easier to quickly scan.

The download logs still span multiple lines, but I don't see an obvious way to resolve this given the length of the urls:

```
Retrieving 'https://cran.rstudio.com/bin/macosx/big-sur-arm64/contrib/4.2/rlang_1.1.1.tgz' ...
	OK [file is up to date]
Retrieving 'https://cran.rstudio.com/bin/macosx/big-sur-arm64/contrib/4.2/memoise_2.0.1.tgz' ...
	OK [file is up to date]
Retrieving 'https://cran.rstudio.com/bin/macosx/big-sur-arm64/contrib/4.2/fastmap_1.1.1.tgz' ...
	OK [file is up to date]
```

Maybe it's ok to just produce long lines and allow the client to wrap them?


R/cache.R
```diff
@@ -106,7 +106,7 @@ renv_cache_path_components <- function(path) {
 
 }
 
-renv_cache_synchronize <- function(record, linkable = FALSE) {
+renv_cache_synchronize <- function(record, linkable = FALSE, max_width = NULL) {
```
kevinushey, line 109: I don't like the idea of adding a print-specific argument to this function. Should it be an option or something else?

kevinushey (commented):
Note that one motivating reason for the extra newline between the "installing" line and the "completed" line was to make CI logs more clear; I'm not sure if this is still an issue but I recall some CI instances buffering by newline and so if you got stuck installing one package you wouldn't see that in the CI output.

kevinushey:
One option for URL shortening would be to print something like:

```
Retrieving rlang 1.1.1 {binary/source} from CRAN ...
```

where the repository name would be drawn from whatever the name in `getOption("repos")` is.

kevinushey:
Either way though I don't feel comfortable merging unless we can convince ourselves that "slow" package retrieval / installation is still reported appropriately on CI (e.g. if you "froze" during one of those steps)

hadley:
Do you have any sense of what CI that was? i.e. what would it take to convince you that buffering wasn't a problem? Would it be adequate to show what happens in a GHA log?

kevinushey:
IIRC this first came up with Packrat restores with RStudio Connect, so perhaps it wasn't CI but instead the way Connect processed restore logs...

Verifying this would be hard though; we'd basically need a test case where `renv` freezes for some extended period of time while installing a package, and then we'd try to view the streaming logs from the GitHub Action to see if the initial "Installing..." bit was reported.

Maybe we can take this change, and then one day (if someone complains) we can roll it back?

kevinushey:
I also sort of wonder if, in the compact display, we could just remove the caching line altogether, and maybe bundle that time spent into the "install" time?

hadley:
Ok, that makes sense to me. I can also spend a few minutes trying to hack the R build process to make it take an arbitrarily long amount of time so we could at least use that as a test case.

I like the idea of combining the caching action/time into the install time. I could still include something about caching in the message:

```
Installing rlang   ... OK [installed binary and moved to cache]
Installing fastmap ... OK [installed binary and moved to cache]
Installing memoise ... OK [installed binary and moved to cache]
```

kevinushey:
It might also be fine to leave the caching notifications out altogether. The main reason I included it was just in case moving a package into the cache happened to be slow for some reason, which might theoretically be a concern if the cache is on a network drive or something.

That said, we shouldn't spend _too_ much time here since in the end we'll want to use `pak` by default and so will get `pak`'s installation log.

hadley:
In that case, let's maybe just leave the messaging as is in this PR? It's a decent improvement and wasn't that much work to do.

kevinushey (approved):

PR #1380, opened 2023-06-01
Only report slow dependency discovery for implicit type

hadley:
This was motivated by #1198 where in one of the examples I noticed:

```
NOTE: Dependency discovery took 11 seconds during snapshot.
Consider using .renvignore to ignore files -- see `?dependencies` for more information.
```

That occurs because the current code includes time spent waiting for the prompt. That seemed like a potentially confusing red herring to me , so I moved the time tracking to `renv_snapshot_filter_implicit()`. I think this is correct as (a) it's the only type that's likely to be slow and (b) it's the only type affected by `.renvignore`.

### DIFF
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -876,8 +876,6 @@ renv_snapshot_dependencies <- function(project, type) {
 
 renv_snapshot_filter <- function(project, records, type, packages, exclude) {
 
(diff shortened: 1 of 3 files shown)


kevinushey (approved):
LGTM; thanks!

PR #1382, opened 2023-06-01
Simplify restart-after-install messaging

### LINKED ISSUE #1371: use clearer message when prompting use to restart after install
E.g. I saw:

```
> renv::install("rstudio/renv")
Retrieving 'https://api.github.com/repos/rstudio/renv/tarball/a02b1b28dc1b1752b005ec8fdca1426c5ba4b956' ...
	OK [downloaded 497 Kb in 0.91 seconds]
Installing renv [0.17.3-76] ...
	OK [built from source in 4.7 seconds]
Copying renv [0.17.3-76] into the cache ...
	OK [copied to cache in 37 milliseconds]
Installed 1 package into library at path "~/Library/R/arm64/4.3/library".
The following package(s) have been updated:

	renv [installed version 0.17.3-76 != loaded version 0.17.3.76]

Consider restarting the R session and loading the newly-installed packages.
```

Perhaps we should just do the regular record diff printout here?

https://github.com/rstudio/renv/blob/a02b1b28dc1b1752b005ec8fdca1426c5ba4b956/R/install.R#L666-L711

Note that the main intention here is to let the user know they installed a package that was already loaded and so should (probably) restart the R session.

hadley:
Fixes #1371


R/install.R
```diff
@@ -658,44 +658,16 @@ renv_install_postamble <- function(packages) {
 
   # only diagnose packages currently loaded
   packages <- renv_vector_intersect(packages, loadedNamespaces())
-  if (empty(packages))
-    return(TRUE)
 
-  # get version of package in library
```
hadley, line 664: I realise it's slightly bad form to delete comments, but IMO based on the names, it's pretty obvious what the code is doing.

R/install.R
```diff
@@ -658,44 +658,16 @@ renv_install_postamble <- function(packages) {
...
-  # get version of package currently loaded
   loaded <- map_chr(packages, renv_namespace_version)
 
-  # collect into data.frame
-  data <- data.frame(
-    Package   = packages,
-    Installed = installed,
-    Loaded    = loaded,
-    stringsAsFactors = FALSE
+  renv_pretty_print(
```
hadley, line 665: `renv_pretty_print()` does nothing if `values` is empty of `renv_verbose()` is not `TRUE`.

kevinushey (approved):
LGTM!

Do you think we need to emphasize in the printed output that these are packages which are already loaded, hence why we're suggesting a restart?

FWIW pak chooses to notify the user before installation, e.g.

```
> pak::pkg_install("rlang")
 
→ Will install 1 package.
→ The package (1.88 MB) is cached.
+ rlang   1.1.1 

! rlang is loaded in the current R session, you probably need to restart R after the installation.

ℹ No downloads are needed, 1 pkg (1.88 MB) is cached
✔ Got rlang 1.1.1 (aarch64-apple-darwin20) (1.88 MB)
✔ Installed rlang 1.1.1  (38ms)                                
✔ 1 pkg: added 1, dld 1 (1.88 MB) [998ms]                      
```

hadley:
Yeah, that makes sense to me.

Agreed that the pak message should come after install.

PR #1384, opened 2023-06-02
Refactor renv_dependencies_discover_description()

hadley:
* Determine project root once, and then use it for settings.
* Rearrange code and update comments to make flow a bit easier to understand


kevinushey (approved):
LGTM!

PR #1385, opened 2023-06-02
Refactor dependency handling in install()

hadley:
Now we first discover all dependencies then update those generic records with specifics found in the `DESCRIPTION`.

### DIFF
--- a/R/description.R
+++ b/R/description.R
@@ -145,40 +145,6 @@ renv_description_parse_field <- function(field) {
 
 }
 
-renv_description_remotes <- function(descpath) {
-
-  # read Remotes field from DESCRIPTION
-  desc <- renv_description_read(path = descpath)
-  remotes <- desc[["Remotes"]]
-  if (is.null(remotes))
-    return(NULL)
-
-  # parse each remote entry
-  entries <- strsplit(remotes, "\\s*,\\s*", perl = TRUE)[[1L]]
-  parsed <- map(entries, renv_description_remotes_parse)
-
-  # ensure named
-  names(parsed) <- map_chr(parsed, `[[`, "Package")
-
-  # and return
-  parsed
-
-}
-
-renv_description_remotes_parse <- function(entry) {
-
-  status <- catch(renv_remotes_resolve(entry))
-
-  if (inherits(status, "error")) {
-    fmt <- "failed to resolve remote '%s' from project DESCRIPTION file; skipping"
-    warningf(fmt, entry)
-    return(NULL)
-  }
-
-  status
-
-}
-
 renv_description_resolve <- function(path) {
 
   case(
--- a/R/install.R
+++ b/R/install.R
@@ -118,7 +118,6 @@ install <- function(packages = NULL,
   }
   install_all <- is.null(packages)
 
-
   project <- renv_project_resolve(project)
   renv_project_lock(project = project)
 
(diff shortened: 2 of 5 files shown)


tests/testthat/test-install.R
```diff
@@ -275,21 +274,12 @@ test_that("install() installs inferred dependencies", {
   skip_on_cran()
   renv_tests_scope("breakfast")
 
-  # use dummy library path
```
hadley, line 278: Not needed since `renv_test_scope()` does this.

tests/testthat/test-install.R
```diff
@@ -275,21 +274,12 @@ test_that("install() installs inferred dependencies", {
...
-  renv_scope_libpaths(templib)
-
   # try installing packages
   install()
 
   # validate that we've installed breakfast + deps
   expect_true(renv_package_installed("breakfast"))
 
-  # try calling install once more; nothing should happen
-  records <- install()
```
hadley, line 290: `install()` now returns one record, but my understanding is that's correct because `install()` should always re-install.

R/description.R
```diff
@@ -145,40 +145,6 @@ renv_description_parse_field <- function(field) {
 
 }
 
-renv_description_remotes <- function(descpath) {
```
hadley, line 148: Removed because it's basically a duplicate of `renv_project_remotes_description_remotes()`

R/install.R
```diff
@@ -142,25 +144,23 @@ install <- function(packages = NULL,
...
   }
 
   # resolve remotes
-  remotes <- lapply(remotes, renv_remotes_resolve)
+  remotes <- lapply(packages, renv_remotes_resolve)
   names(remotes) <- extract_chr(remotes, "Package")
 
-  # update records with requested remotes
+  # apply version specifications and Remotes from DESCRIPTION
+  remotes <- renv_install_remotes_update(remotes, project, all = install_all)
```
hadley, line 158: Maybe should only do this when `packages` is `NULL`

R/project.R
```diff
@@ -101,29 +101,12 @@ renv_project_type_impl <- function(path) {
...
-  )
-
-  packages <- sort(unique(deps$Package))
-  ignored <- c("renv", renv_project_ignored_packages(project))
-  setdiff(packages, ignored)
-
-}
-
-renv_project_remotes_description <- function(project, descpath, fields = NULL) {
+  if (!file.exists(descpath))
```
kevinushey, line 105: IIUC this is safe since the only time version-specific constraints could show up would be in the DESCRIPTION file; is that worth including as a comment? (as to why we don't just use `dependencies()`)

tests/testthat/test-install.R
kevinushey: Normally `renv::install()` (called without arguments) should be a no-op if all installed packages are up-to-date, e.g.

```
> renv::install()
Downloading 24 package(s) and their dependencies
Done

* There are no packages to install.
```

R/install.R
```diff
@@ -142,25 +144,23 @@ install <- function(packages = NULL,
     return(renv_pak_install(packages, libpaths, project))
   }
 
-  # get and resolve the packages / remotes to be installed
-  remotes <- packages %||% renv_project_remotes(project, dependencies)
-  if (empty(remotes)) {
+  packages <- packages %||% renv_snapshot_dependencies(project, "implicit")
```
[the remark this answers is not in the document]
hadley: Yeah, these PRs are preliminary to some behaviour changes.

R/project.R
hadley: Yeah, this function is now specifically about getting package constraints from the `DESCRIPTION`, rather than also finding package names from code.

tests/testthat/test-install.R
hadley: I think either you or I was confused by our slack thread, because I thought that this was what you wanted.

kevinushey (approved):

tests/testthat/test-install.R
```diff
@@ -275,15 +274,11 @@ test_that("install() installs inferred dependencies", {
...
-  templib <- renv_scope_tempfile("renv-library-")
-  ensure_directory(templib)
-  renv_scope_libpaths(templib)
-
   # try installing packages
-  install()
+  records <- install()
 
   # validate that we've installed breakfast + deps
+  expect_length(records, 2L)
```
kevinushey, line 281: I think this should be 4, since:

```
> tools:::package_dependencies("breakfast", recursive = TRUE)
$breakfast
[1] "oatmeal" "toast"   "bread"  
```

hadley: And indeed it is 4 😄

PR #1389, opened 2023-06-02
Snapshot quarto and pandoc versions

### LINKED ISSUE #615: consider encoding version of pandoc used into lockfile?
For projects using `rmarkdown`, since having an appropriate version of `pandoc` available is required by `rmarkdown` in most workflows.

### LINKED ISSUE #1143: question: can renv capture what version of Quarto is installed?
Imagine that one creates a tool in R--for example, a package or golem app--that creates reports with Quarto.

For all packages involved in creating report content, renv can capture the details in a lock file.

For Quarto, can one snapshot the Quarto version, even though Quarto CLI (unlike rmarkdown) is not a package?

If this is out of scope for renv, what is the best solution for this type of problem? I see that the quarto package has functions that could help in checking that a user's environment is properly provisioned. The `quarto_path()` function returns the path of one's Quarto install. The `quarto_version()` function fetches the version of Quarto installed. Used in combination, these functions could help us determine whether Quarto is installed and, if so, whether it's version is OK for the tool's needs.

hadley:
Fixes #615. Fixes #1143.

### DIFF
--- a/R/cite.R
+++ b/R/cite.R
@@ -84,7 +84,7 @@ cite <- function(type = c("plain", "bibtex"),
     "output.html"
   )
 
-  system2("pandoc", args)
+  system2(pandoc_path(), args)
   system("open output.html")
 
 }
--- a/R/lockfile-external.R
+++ b/R/lockfile-external.R
@@ -36,6 +36,9 @@ pandoc_version <- function() {
   }
 
   out <- system2(path, "--version", stdout = TRUE)
+  if (!is.null(attr(out, "status"))) {
+    return(NULL)
+  }
 
   # From pandoc::pandoc_version()
   re <- "^pandoc(?:\\.exe)? ([\\d.]+).*$"
@@ -57,5 +60,10 @@ quarto_version <- function() {
   }
 
   # from quarto::quarto_version
-  system2(path, "--version", stdout = TRUE)
+  out <- system2(path, "--version", stdout = TRUE)
+    if (!is.null(attr(out, "status"))) {
+    return(NULL)
+  }
+
+  out
 }
--- a/R/lockfile-external.R
+++ b/R/lockfile-external.R
@@ -5,6 +5,9 @@ renv_lockfile_external <- function() {
   )
 }
 
+
+# pandoc ------------------------------------------------------------------
+
 pandoc_path <- function() {
   if (!is.na(path <- Sys.getenv("RSTUDIO_PANDOC", unset = NA))) {
     return(file.path(path, "pandoc"))
@@ -17,17 +20,6 @@ pandoc_path <- function() {
   NULL
 }
 
-quarto_path <- function() {
-  if (!is.na(path <- Sys.getenv("QUARTO_PATH", unset = NA))) {
-    return(path)
-  }
-
-  if (nzchar(path <- Sys.which("quarto"))) {
-    return(unname(path))
-  }
-
-  NULL
-}
 
 pandoc_version <- function() {
   path <- pandoc_path()
@@ -35,6 +27,7 @@ pandoc_version <- function() {
     return(NULL)
   }
 
+  local_pandoc_safe_environment()
   out <- system2(path, "--version", stdout = TRUE)
   if (!is.null(attr(out, "status"))) {
     return(NULL)
@@ -51,6 +44,43 @@ pandoc_version <- function() {
     version <- paste(version, "9999", sep = ".")
 
   version
+}
+
+# https://github.com/rstudio/rmarkdown/blob/10863429024e236/R/pandoc.R#L720-L758
+local_pandoc_safe_environment <- function(code, envir = parent.frame()) {
+
+  if (renv_envvar_exists("LC_ALL")) {
+    renv_scope_envvars(LC_ALL = NULL, envir = envir)
+  }
+
+  if (renv_envvar_exists("LC_CTYPE")) {
+    renv_scope_envvars(LC_CTYPE = NULL, envir = envir)
+  }
+
+  if (renv_platform_linux()) {
+    if (!renv_envvar_exists("HOME"))
+      stop("The 'HOME' environment variable must be set before running Pandoc.")
+
+    if (!renv_envvar_exists("LANG") || identical(Sys.getenv("LANG"), "en_US"))
+      renv_scope_envvars(LANG = "en_US.UTF-8", envir = envir)
+  }
+
+  invisible()
+}
+
+
+# quarto ------------------------------------------------------------------
+
+quarto_path <- function() {
+  if (!is.na(path <- Sys.getenv("QUARTO_PATH", unset = NA))) {
+    return(path)
+  }
+
+  if (nzchar(path <- Sys.which("quarto"))) {
+    return(unname(path))
+  }
+
+  NULL
 }
 
 quarto_version <- function() {
--- a/R/lockfile-external.R
+++ b/R/lockfile-external.R
@@ -1,3 +1,10 @@
+renv_lockfile_external <- function() {
+  list(
+    quarto = quarto_version(),
+    pandoc = pandoc_version()
+  )
+}
+
 pandoc_path <- function() {
   if (!is.na(path <- Sys.getenv("RSTUDIO_PANDOC", unset = NA))) {
     return(file.path(path, "pandoc"))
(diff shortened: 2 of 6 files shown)


R/lockfile.R
```diff
@@ -5,7 +5,9 @@ renv_lockfile_init <- function(project) {
 
   lockfile$R        <- renv_lockfile_init_r(project)
   lockfile$Python   <- renv_lockfile_init_python(project)
-  lockfile$Packages <- list()
+  lockfile$quarto   <- renv_lockfile_quarto()
```
kevinushey, line 8: I feel like we should collect these into a common lockfile entry, e.g. `lockfile$External$quarto` or something like that.

R/lockfile.R
```diff
@@ -5,7 +5,9 @@ renv_lockfile_init <- function(project) {
...
+  lockfile$pandoc   <- renv_lockfile_pandoc()
+  lockfile$Packages <- list(ver)
```
kevinushey, line 10:
```suggestion
  lockfile$Packages <- list()
```

R/snasphot-external.R
```diff
@@ -0,0 +1,54 @@
...
+  NULL
+}
+
+pandoc_version <- function() {
+  path <- pandoc_path()
+  if (is.null(path)) {
+    return(NULL)
+  }
+
+  out <- system2(path, "--version", stdout = TRUE)
```
kevinushey, line 31: Do we need this? https://github.com/rstudio/rmarkdown/blob/10863429024e236389a48f40ab1485620d53d311/R/pandoc.R#L720-L758

hadley: Hmmmm, I'm guessing no given that the motivating issue was filed in 2014, and I've never heard of it.

aronatkins: I remember this! Pandoc would consume massive amounts of memory (and then get OOM killed) if the "right" environment variables were not set. I think we are very careful about setting these variables in Connect and the IDE and avoid this problem.

hadley: Ok, if you think it's a potential problem, I'll replicate the env vars.

kevinushey: I definitely recall seeing this (I think mainly on Windows?) as well. I think we're better off keeping that code around just in case; it doesn't hurt.

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+  list(
+    quarto = quarto_version(),
+    pandoc = pandoc_version()
+  )
+}
+
+
+# pandoc ------------------------------------------------------------------
+
+pandoc_path <- function() {
```
kevinushey, line 11:
```suggestion
renv_pandoc_path <- function() {
```

By convention, (almost) all internal functions in should have an `renv_` prefix, unless they're generally-applicable utility functions. Can we continue that convention?

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+
+  if (nzchar(path <- Sys.which("pandoc"))) {
+    return(unname(path))
+  }
+
+  NULL
+}
+
+
+pandoc_version <- function() {
```
kevinushey, line 24:
```suggestion
renv_pandoc_version <- function() {
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+  path <- pandoc_path()
```
kevinushey, line 25:
```suggestion
  path <- renv_pandoc_path()
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+  if (is.null(path)) {
+    return(NULL)
+  }
+
+  local_pandoc_safe_environment()
```
kevinushey, line 30:
```suggestion
  renv_scope_pandoc()
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+
+  version <- gsub(re, "\\1", out[[1]], perl = TRUE)
+  if (grepl("-nightly-", out[[1]]))
+    version <- paste(version, "9999", sep = ".")
+
+  version
+}
+
+# https://github.com/rstudio/rmarkdown/blob/10863429024e236/R/pandoc.R#L720-L758
+local_pandoc_safe_environment <- function(code, envir = parent.frame()) {
```
kevinushey, line 50:
```suggestion
renv_scope_pandoc <- function(code, envir = parent.frame()) {
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+
+  if (renv_envvar_exists("LC_ALL")) {
+    renv_scope_envvars(LC_ALL = NULL, envir = envir)
```
kevinushey, line 53: Is it worth collecting all the environment variables we want to set in a single list, and then calling `renv_scope_envvars()` once at the bottom?

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+      renv_scope_envvars(LANG = "en_US.UTF-8", envir = envir)
+  }
+
+  invisible()
+}
+
+
+# quarto ------------------------------------------------------------------
+
+quarto_path <- function() {
```
kevinushey, line 74:
```suggestion
renv_quarto_path <- function() {
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  if (nzchar(path <- Sys.which("quarto"))) {
+    return(unname(path))
+  }
+
+  NULL
+}
+
+quarto_version <- function() {
```
kevinushey, line 86:
```suggestion
renv_quarto_version <- function() {
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+  path <- quarto_path()
```
kevinushey, line 87:
```suggestion
  path <- renv_quarto_path()
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+  if (is.null(path)) {
+    return(NULL)
+  }
+
+  # from quarto::quarto_version
+  out <- system2(path, "--version", stdout = TRUE)
+    if (!is.null(attr(out, "status"))) {
+    return(NULL)
```
kevinushey, line 95: indent off here?

R/cite.R
```diff
@@ -84,7 +84,7 @@ cite <- function(type = c("plain", "bibtex"),
     "output.html"
   )
 
-  system2("pandoc", args)
+  system2(pandoc_path(), args)
```
kevinushey, line 87:
```suggestion
  system2(renv_pandoc_path(), args)
```

kevinushey: but we should probably get rid of `cite()`, it was an internal utility that I tried putting together that doesn't really belong in renv.

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
+renv_lockfile_external <- function() {
+  list(
+    quarto = quarto_version(),
```
kevinushey, line 3:
```suggestion
    quarto = renv_quarto_version(),
```

R/lockfile-external.R
```diff
@@ -0,0 +1,99 @@
...
+    pandoc = pandoc_version()
```
kevinushey, line 4:
```suggestion
    pandoc = renv_pandoc_version()
```

R/lockfile.R
```diff
@@ -6,6 +6,7 @@ renv_lockfile_init <- function(project) {
   lockfile$R        <- renv_lockfile_init_r(project)
   lockfile$Python   <- renv_lockfile_init_python(project)
   lockfile$Packages <- list()
+  lockfile$External <- renv_lockfile_external()
```
kevinushey, line 9: I think you need to update https://github.com/rstudio/renv/blob/334511e6ee93cc473b943701a155656fe12590e6/R/lockfile.R#L146-L164 as well so that `External` comes last.

kevinushey (commented):
I'm still a bit worried about collaborative scenarios where the pandoc / quarto versions could start flipping around if different people had different versions of the tools available (e.g. because they were using different versions of RStudio). This could lead to noisy git diffs which users have complained about in the past.

PR #1391, opened 2023-06-05
Add example for setting renv.auth per package as function

bartekch:
`renv.auth` option for specific package could be set as a list, but also as a function with single argument record, similarly, but not exactly the same, to general renv.auth function.

I had a case where passing the option as a function was the most convenient solution, and I'd appreciate if this was explicitly listed as a viable alternative in vignette. Hope it's not some unrecommended solution or anything.


kevinushey (approved):
Thanks!

PR #1394, opened 2023-06-05
Simplify handling of `Suggested` dependencies

hadley:
* Top-level `DESCRIPTION` is now treated the same way regardless of whether it's a "Package" or not.
* `settings$package.dependency.fields()` no longer affects transitive dependencies.
* Document how "development dependencies" are found in `dependencies()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `dependencies()` now marks `Suggested` packages listed in `DESCRIPTION` files
+  as development dependencies regardless of whether or not they're in 
+  "package" project.
+
 * If `renv::snapshot()` finds missing packages, a new prompt allows you to 
   install them before continuing (#1198).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,9 @@
   as development dependencies regardless of whether or not they're in 
   "package" project.
 
+* `settings$package.dependency.fields()` now only affects packages installed
+  directly by the user, not downstream dependencies of those packages.
+
 * If `renv::snapshot()` finds missing packages, a new prompt allows you to 
   install them before continuing (#1198).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,8 +2,8 @@
 # renv 0.18.0  (UNRELEASED)
 
 * `dependencies()` now marks `Suggested` packages listed in `DESCRIPTION` files
-  as development dependencies regardless of whether or not they're in 
-  "package" project.
+  as development dependencies regardless of whether or not they're a "package" 
+  project.
 
 * `settings$package.dependency.fields()` now only affects packages installed
   directly by the user, not downstream dependencies of those packages.
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -145,10 +145,14 @@
 #'   * `"fatal"`: errors are fatal and stop execution.
 #'   *  `"ignored"`: errors are ignored and not reported to the user.
 #'
-#' @param dev Boolean; include 'development' dependencies as well? That is,
-#'   packages which may be required during development but are unlikely to be
-#'   required during runtime for your project. By default, only runtime
-#'   dependencies are returned.
+#' @param dev Boolean; include development dependencies? These packages are
+#'   typically required when developing the project, but not when running it
+#'   (i.e. you want them installed when humans are working on the project but
+#'   not when computers are deploying it).
+#'
+#'   Development dependencies include packages listed in the `Suggests` field
+#'   of a `DESCRIPTION` found in the project root, and roxygen2 or devtools if
+#'   their use is implied by other project metadata.
 #'
 #' @return An \R `data.frame` of discovered dependencies, mapping inferred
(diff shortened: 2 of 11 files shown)


R/dependencies.R
```diff
@@ -536,25 +540,34 @@ renv_dependencies_discover_description <- function(path,
...
     path = path,
-    type = type
+    proj_desc = proj_desc
   )
 
+  # Infer a dependency on roxygen2 and devtools if RoxygenNote present
+  if (!is.null(dcf$RoxygenNote)) {
+    data <- c(
+      data,
+      list(renv_dependencies_list(path, c("roxygen2", "devtools"), dev = TRUE))
```
hadley, line 561: IMO this only needs to add roxygen2, not devtools, but that would be a behaviour change.

tests/testthat/test-dependencies.R
```diff
@@ -150,22 +150,24 @@ test_that("no warnings are produced when crawling dependencies", {
...
+    stringsAsFactors = FALSE
+  )
+
   writeLines(c("Type: Project", "Suggests: bread"), con = "DESCRIPTION")
   deps <- dependencies(dev = TRUE)
-  expect_true(nrow(deps) == 1)
-  expect_true(deps$Package == "bread")
-  expect_true(deps$Dev)
-})
+  expect_equal(deps[c("Package", "Dev")], expected)
```
hadley, line 165: This construction is equivalent to the previous three lines but will give a more informative failure.

kevinushey (approved):

tests/testthat/test-dependencies.R
```diff
@@ -195,14 +210,6 @@ test_that("Suggest dependencies are ignored by default", {
   expect_false(renv_package_installed("egg"))
 })
 
-test_that("Suggest dependencies are used when requested", {
```
hadley, line 198: Moved to `install.R` which is a location.

hadley:
@kevinushey It's worth taking another look at this now, since I've also slightly changed the meaning of `settings$package.dependency.fields()`.

NEWS.md
```diff
@@ -1,6 +1,13 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `dependencies()` now marks `Suggested` packages listed in `DESCRIPTION` files
```
kevinushey, line 4: What happens if a project contains DESCRIPTION files in sub-folders?

R/dependencies.R
```diff
@@ -512,49 +516,49 @@ renv_dependencies_discover_renv_lock <- function(path) {
 renv_dependencies_discover_description <- function(path,
                                                    fields = NULL,
                                                    subdir = NULL,
-                                                   project = NULL)
+                                                   project = NULL,
+                                                   is_dependency = FALSE)
```
kevinushey, line 520: How do you feel about using `fields = "most"` in the callers, and avoiding the addition of a new parameter here? See also https://github.com/rstudio/renv/blob/557433bafa4677221a1461588290dc71421bc664/R/description.R#L168-L196.

R/dependencies.R
```diff
@@ -579,13 +583,12 @@ renv_dependencies_discover_description_impl <- function(dcf, field, path, type)
...
 
   # create dependency list
-  dev <- field == "Suggests" && type != "package"
   renv_dependencies_list(
     path,
     extract_chr(matches, 2L),
     extract_chr(matches, 3L),
     extract_chr(matches, 4L),
-    dev
+    dev = field == "Suggests"
```
kevinushey, line 591: FWIW I think the reason I had this was to avoid issues with DESCRIPTION files located in sub-directories within a project, but that imply separate considerations (e.g. should we really be parsing those for dependencies at all from the top-level?)

R/package.R
```diff
@@ -276,7 +276,7 @@ renv_package_dependencies_impl <- function(package,
   assign(package, location, envir = visited, inherits = FALSE)
 
   # find its dependencies from the DESCRIPTION file
-  deps <- renv_dependencies_discover_description(location, fields)
+  deps <- renv_dependencies_discover_description(location, is_dependency = TRUE)
```
kevinushey, line 279: `fields = "most"`?

R/dependencies.R
kevinushey: Oops, I guess I mean "strong", not "most".

hadley: Oh yeah, that works much better. That also required a change to `renv_project_remotes()`, but I think for the better — now no-arg `install()` will respect the active profile.

NEWS.md
hadley: That shouldn't have changed — a `DESCRIPTION` in a sub-folder will use `settings$package.dependency.fields`> A `DESCRIPTION` in the project root also uses `Suggests` and `Config/renv/profiles/{profile}/dependencies`.

kevinushey: We discussed on Slack, but maybe we can reconsider this part (only look at DESCRIPTION files in the project root)? For a separate PR though, so if this preserves the existing behavior then I think we're happy.

kevinushey (approved):
LGTM!

Issue #1396, opened 2023-06-06
snapshot(exclude = "myPackage") failure


russHyde:
The 'exclude' parameter in renv::snapshot() doesn't exclude a local package.

I'm writing a shiny app as a package ("golemDeploy", built with {golem}) and am having some problems using snapshot.

There is a top-level `./app.R` file, that is required to start the app after deployment on a Connect server.
It looks like this:

```
pkgload::load_all(export_all = FALSE,helpers = FALSE,attach_testthat = FALSE)
options( "golem.app.prod" = TRUE)
golemDeploy::run_app()
```

Now, I don't want {golemDeploy} to be included in the renv.lock for this project. So I'm using `renv::snapshot(exclude = "golemDeploy")`. But snapshot fails in two different ways here:

When `golemDeploy` is not installed locally, snapshot fails because that package isn't installed (which seems odd since I've asked for that package to be excluded)

```r
> renv::snapshot(exclude = "golemDeploy")
The following required packages are not installed:

	golemDeploy

Packages must first be installed before `renv` can snapshot them.
Consider installing these packages using `renv::install()`.
Use `renv::dependencies()` to see where this package is used in your project.
```

When golemDeploy is installed locally, snapshot states that it will be updating that package in the renv.lock even though I've asked for it to be excluded. (Proceeding here means that an renv.lock is written, but golemDeploy is not included in the renv.lock, contrary to what the message implies)

```r
> # when golemDeploy is installed locally:
> renv::snapshot(exclude = "golemDeploy")
The following package(s) will be updated in the lockfile:

# (Unknown Source) ===================
- golemDeploy   [0.0.0.9000 -> *]

Do you want to proceed? [y/N]:
```

(Admittedly, I can just remove the `golemDeploy::` prefix from the app.R file, but it seems like there is a bug with 'exclude' in renv::snapshot() here)

kevinushey:
Thanks for the bug report; this should be better handled in the development version of `renv`. Can you install it with:

```
renv::install("rstudio/renv")
```

and let us know if the issue seems fixed on your end?

russHyde:
Thanks.
Now, when the package is installed, renv::snapshot outputs the following:

```r
> renv::snapshot(exclude = "golemDeploy")
The following package(s) were installed from an unknown source:

- golemDeploy [0.1.0]

renv may be unable to restore these packages in the future.
Consider reinstalling these packages from a known source (e.g. CRAN).

Do you want to proceed? [Y/n]:
```

[...]

PR #1397, opened 2023-06-06
improve handling of exclude

kevinushey:
Closes https://github.com/rstudio/renv/issues/1396.

### DIFF
--- a/R/load.R
+++ b/R/load.R
@@ -187,6 +187,10 @@ renv_load_minimal <- function(project) {
   if (length(lockfile))
     renv_load_python(project, lockfile$Python)
 
+  # TODO: Should this be set in non-interactive cases too?
+  if (interactive())
+    renv_project_set(project)
+
   invisible(project)
 
 }
@@ -611,6 +615,10 @@ renv_load_bioconductor_validate <- function(project, version) {
 }
 
 renv_load_switch <- function(project) {
+
+  # skip when testing
+  if (is_testing())
+    return(project)
 
   # safety check: avoid recursive unload attempts
   unloading <- getOption("renv.unload.project")
--- a/R/lockfile.R
+++ b/R/lockfile.R
@@ -172,33 +172,29 @@ renv_lockfile_create <- function(project,
   libpaths <- libpaths %||% renv_libpaths_all()
   type <- type %||% settings$snapshot.type(project = project)
 
+  # use a restart, so we can allow the user to install packages before snapshot
+  lockfile <- withRestarts(
+    renv_lockfile_create_impl(project, libpaths, type, packages, exclude),
+    renv_recompute_records = function() {
+      renv_dynamic_reset()
+      renv_lockfile_create_impl(project, libpaths, type, packages, exclude)
+    }
+  )
+}
+
+renv_lockfile_create_impl <- function(project, libpaths, type, packages, exclude) {
+
   lockfile <- renv_lockfile_init(project)
 
   records <- renv_snapshot_libpaths(libpaths = libpaths, project = project)
 
-  records <- tryCatch(
-    renv_snapshot_filter(
-      project  = project,
-      records  = records,
-      type     = type,
-      packages = packages,
-      exclude  = exclude
-    ),
-    renv_recompute_records = identity
+  records <- renv_snapshot_filter(
+    project  = project,
+    records  = records,
+    type     = type,
+    packages = packages,
+    exclude  = exclude
   )
-
-  # We re-computing and re-filtering records once, if requested by user
-  if (inherits(records, "renv_recompute_records")) {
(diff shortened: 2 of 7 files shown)


R/snapshot.R
```diff
@@ -969,20 +970,17 @@ renv_snapshot_filter_report_missing <- function(missing, type) {
     # do nothing
   } else if (choice == "install") {
     install(missing, prompt = FALSE)
-    # User will only see this in exceptional circumstances as it is
-    # caught once by renv_lockfile_create()
-    stop(errorCondition(
-      message = "Failed to restart snapshotting after install",
-      class = "renv_recompute_records"
-    ))
+    renv_condition_signal("renv_recompute_records")
```
kevinushey, line 973: FWIW it feels like we should be setting a flag or something indicating whether we would prompt the user here.

R/snapshot.R
```diff
@@ -885,40 +885,40 @@ renv_snapshot_filter <- function(project, records, type, packages, exclude) {
...
-    packages = renv_snapshot_filter_packages(project, records, packages),
+    all      = renv_snapshot_filter_all(project, records, exclude),
+    custom   = renv_snapshot_filter_custom(project, records, exclude),
+    explicit = renv_snapshot_filter_explicit(project, records, exclude),
+    implicit = renv_snapshot_filter_implicit(project, records, exclude),
+    packages = renv_snapshot_filter_packages(project, records, packages, exclude),
     stopf("unknown snapshot type '%s'", type)
   )
 
-  if (length(exclude))
```
kevinushey, line 896: The PR effectively moves this piece into each of the filter routines, so that filters are processed at the right time.

R/snapshot.R
```diff
@@ -885,40 +885,40 @@ renv_snapshot_filter <- function(project, records, type, packages, exclude) {
...
+  missing <- setdiff(packages, c(names(records), ignored, exclude))
   if (!`_renv_status_running`)
     renv_snapshot_filter_report_missing(missing, type)
 
   # ignore packages as defined by project
   used <- setdiff(packages, ignored)
 
   # include transitive dependencies
+  #
+  # TODO: what if a package that a user requested be ignored pops up here; should
```
hadley, line 920: IMO, I think it'd be sufficient to document that `exclude` won't exclude transitive dependencies.

R/snapshot.R
hadley: I think this needs to be a stop, otherwise if `install()` doesn't get everything the first time, and the user tries to `install()` again, then it will appear to succeed.  I don't think the previous behaviour was that great either, but at least you got a clear error.

tests/testthat/test-snapshot.R
```diff
@@ -376,18 +376,6 @@ test_that("a project using explicit snapshots is marked in sync appropriately",
 
 })
 
-test_that("we can explicitly exclude some packages from snapshot", {
```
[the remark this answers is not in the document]
hadley: Agree that it doesn't make sense to allow this.

R/lockfile.R
```diff
@@ -184,13 +184,11 @@ renv_lockfile_create <- function(project,
...
     ),
-    renv_recompute_records = function(cnd) {
-      NULL
-    }
+    renv_recompute_records = identity
   )
 
   # We re-computing and re-filtering records once, if requested by user
-  if (is.null(records)) {
+  if (inherits(records, "renv_recompute_records")) {
```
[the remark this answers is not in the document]
hadley: Also it's not directly related to the goal of this PR?

hadley (approved):

R/load.R
```diff
@@ -187,6 +187,10 @@ renv_load_minimal <- function(project) {
   if (length(lockfile))
     renv_load_python(project, lockfile$Python)
 
+  # TODO: Should this be set in non-interactive cases too?
+  if (interactive())
+    renv_project_set(project)
```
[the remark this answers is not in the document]
kevinushey: That's what I saw too. 😭

R/snapshot.R
kevinushey: I think I have the right solution now:

- Move the lockfile creation impl into a separate function
- Invoke that from the top level in a `withRestarts` handler
- Check for relevant restarts when prompting the user, and only prompt to install if such a restart exists

The action at a distance is sort of hard to reason about, but I believe it's correct.

R/lockfile.R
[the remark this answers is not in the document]
kevinushey: I think my main concern about a sentinel value like `NULL` is that functions might reasonably return `NULL` in some cases, so it's hard to reason that the `NULL` was definitely produced because of the condition.

Either way though I think I landed on something we'll both agree is better?

PR #1399, opened 2023-06-07
Don't `snapshot()` development dependencies

### LINKED ISSUE #1019: `Suggests` dependencies included by default
Hi,

I'm experiencing (I think), the same issue as https://github.com/rstudio/renv/issues/823. But since that is a little stale (and on an older version) I thought I'd raise the issue here again.

Using, renv 0.15.5

Reproduce as before,

1. Create a package (for instance with `usethis::create_package("suggestdemo")`)
2. Run `renv::init()` and restart the R session
3. Add the `Suggests`-field to the `DESCRIPTION` file (for instance `Suggests: knitr`)
4. Run `renv::status()`

The result is

```R
> renv::status()
* The project is already synchronized with the lockfile.
The following package(s) are used in the project, but are not installed:

	knitr

Consider installing these packages, and then using `renv::snapshot()`
to record these packages in the lockfile.
```

Note also,

```R
> renv::settings$package.dependency.fields()
[1] "Imports"   "Depends"   "LinkingTo"
```

And finally,

```R
> renv::dependencies(dev = TRUE)
Finding R package dependencies ... Done!
                           Source Package Require Version   Dev
1 D:/code/suggestdemo/DESCRIPTION   knitr                 FALSE
2   D:/code/suggestdemo/renv.lock    renv                 FALSE
```

My expectation would be to not see `knitr` in `renv::status` and to see it listed as a development TRUE dependency in `renv::dependencies`

Thanks

hadley:
Fixes #1019

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -143,7 +143,7 @@ install <- function(packages = NULL,
     return(renv_pak_install(packages, libpaths, project))
   }
 
-  packages <- packages %||% renv_snapshot_dependencies(project, "implicit")
+  packages <- packages %||% renv_snapshot_dependencies(project, include_dev = TRUE)
   if (empty(packages)) {
     writef("* There are no packages to install.")
     return(invisible(list()))
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `snapshot()` and `status()` no longer track development dependencies.
+  `install()` will continue to install them (#1019).
+
 * `dependencies()` now marks `Suggested` packages listed in `DESCRIPTION` files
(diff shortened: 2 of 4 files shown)


kevinushey (approved):

R/install.R
```diff
@@ -143,7 +143,7 @@ install <- function(packages = NULL,
     return(renv_pak_install(packages, libpaths, project))
   }
 
-  packages <- packages %||% renv_snapshot_dependencies(project, "implicit")
+  packages <- packages %||% renv_snapshot_dependencies(project, dev = TRUE)
```
kevinushey, line 146: Do we need to document the fact that `install()` now respects the snapshot type? (The change probably makes sense)

hadley: Oh yeah, good catch.

PR #1400, opened 2023-06-07
Tweak install log

### LINKED ISSUE #1386: tweak display of installation log
```
> renv::install()
Downloading 24 package(s) and their dependencies
Done

* There are no packages to install.
```

The output makes it look like 24 packages _need_ to be downloaded, but really it's just we inferred the project depends on 24 packages, which will need to be downloaded and installed. (Some of which might already be downloaded and cached.)

Also, by convention, progress headers should usually end with an ellipsis, e.g. something like:

```
* Updating 24 package(s) and their dependencies ... Done!
* There are no packages to install.
```

I also noticed that we're now getting prompted _after_ download; is that expected?

```
> renv::install()
Downloading 24 package(s) and their dependencies
Retrieving 'https://cloud.r-project.org/bin/macosx/big-sur-arm64/contrib/4.3/devtools_2.4.5.tgz' ...
        OK [downloaded 420.8 Kb in 0.12 seconds]
Done

The following package(s) will be installed:

- devtools [2.4.5]

Packages will be installed into "~/Library/R/arm64/4.3/library"

Do you want to proceed? [Y/n]: y
```

hadley:
Fixes #1386

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -191,6 +191,8 @@ install <- function(packages = NULL,
     return(invisible(list()))
   }
 
+  libpath <- renv_path_pretty(renv_libpaths_active())
+  writef(header(sprintf("Installing package(s) into %s", libpath)))
   if (prompt || renv_verbose())
     renv_install_report(records)
   cancel_if(prompt && !proceed())
@@ -727,10 +729,7 @@ renv_install_preflight <- function(project, libpaths, records) {
 renv_install_report <- function(records) {
   renv_pretty_print_records(
     records,
-    preamble = "The following package(s) will be installed:",
-    postamble = sprintf(
-      "Packages will be installed into %s", renv_path_pretty(renv_libpaths_active())
-    )
+    preamble = "The following package(s) will be installed:"
   )
 }
 
--- a/R/pretty.R
+++ b/R/pretty.R
@@ -49,7 +49,7 @@ renv_pretty_print_records <- function(records,
   text <- sprintf("- %s [%s]", format(packages), descs)
 
   all <- c(
-    preamble, if (length(postamble)) "",
+    preamble, if (length(preamble)) "",
     text, "",
     postamble, if (length(postamble)) ""
   )
--- a/R/retrieve.R
+++ b/R/retrieve.R
@@ -28,12 +28,16 @@ retrieve <- function(packages) {
   }
   renv_scope_options(HTTPUserAgent = agent)
 
-  writef("Downloading %s package(s) and their dependencies", length(packages))
+  writef(header("Downloading package(s)"))
   # TODO: parallel?
   handler <- state$handler
   for (package in packages)
     handler(package, renv_retrieve_impl(package))
-  writef(c("Done", ""))
+
+  if (is.null(state$downloaded)) {
+    writef("Nothing needed")
+  }
+  writef("")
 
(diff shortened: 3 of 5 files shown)


R/pretty.R
```diff
@@ -49,7 +49,7 @@ renv_pretty_print_records <- function(records,
   text <- sprintf("- %s [%s]", format(packages), descs)
 
   all <- c(
-    preamble, if (length(postamble)) "",
+    preamble, if (length(preamble)) "",
```
hadley, line 52: Ooops

R/retrieve.R
```diff
@@ -28,12 +28,16 @@ retrieve <- function(packages) {
...
-  writef("Downloading %s package(s) and their dependencies", length(packages))
+  writef(header("Downloading package(s)"))
   # TODO: parallel?
   handler <- state$handler
   for (package in packages)
     handler(package, renv_retrieve_impl(package))
-  writef(c("Done", ""))
+
+  if (is.null(state$downloaded)) {
+    writef("Nothing needed")
```
hadley, line 38: I think we also need to print something if no downloads actually occurred.

R/pretty.R
kevinushey: 🙈

R/retrieve.R
kevinushey:
```suggestion
    writef("[no downloads required]")
```

I like using adornments for placeholder text like this just to help make it clear that it's not "regular" output. I'm open to alternatives on wordsmithing though.

tests/testthat/_snaps/install.md
```diff
@@ -3,15 +3,14 @@
     Code
       install("bread@0.1.0")
     Output
-      Downloading 1 package(s) and their dependencies
-      Done
+      # Downloading package(s) =============
```
kevinushey, line 6: Let's change the format of `header()` here, the `#` and `=` feel too big and noisy. (use `--`?)

R/install.R
```diff
@@ -191,6 +191,8 @@ install <- function(packages = NULL,
     return(invisible(list()))
   }
 
+  libpath <- renv_path_pretty(renv_libpaths_active())
+  writef(header(sprintf("Installing package(s) into %s", libpath)))
```
kevinushey, line 195:
```suggestion
  writef(header(sprintf("Installing packages into %s", libpath)))
```

R/retrieve.R
```diff
@@ -28,12 +28,16 @@ retrieve <- function(packages) {
   }
   renv_scope_options(HTTPUserAgent = agent)
 
-  writef("Downloading %s package(s) and their dependencies", length(packages))
+  writef(header("Downloading package(s)"))
```
kevinushey, line 31:
```suggestion
  writef(header("Downloading packages"))
```

R/install.R
kevinushey: I think it's okay to drop `(s)` here, since this reads like the name of a "stage" of the install process; that is, the "Installing packages" stage.

tests/testthat/_snaps/install.md
hadley: Yeah, I was thinking I'd do that in a follow up PR.

kevinushey (approved):

PR #1407, opened 2023-06-07
test that binary packages can be loaded after install

### LINKED ISSUE #1275: load binary packages ahead of use
Installing binary packages can often hide problems with an environment until runtime.

Take, for example, the `Dockerfile`:

```dockerfile
FROM rstudio/r-base:4.3-jammy

RUN apt-get update && apt-get -y install pkg-config

ARG REPOSITORYhttps://packagemanager.rstudio.com/cran/__linux__/jammy/latest
RUN R -s -e "install.packages('rjags', repos=c(CRAN='${REPOSITORY}'))"
RUN R -s -e "library(rjags)"
```

Building this image:

```bash
docker build \
    --progress=plain \
    --no-cache -f Dockerfile.renv .
```

This example errs because the `library` statement fails to load the `rjags` package (because `jags` is missing).

```
#7 [4/4] RUN R -s -e "library(rjags)"
#7 0.385 Loading required package: coda
#7 0.427 Error: package or namespace load failed for ‘rjags’:
#7 0.427  .onLoad failed in loadNamespace() for 'rjags', details:
#7 0.427   call: dyn.load(file, DLLpath = DLLpath, ...)
#7 0.427   error: unable to load shared object '/opt/R/4.3.0/lib/R/library/rjags/libs/rjags.so':
#7 0.427   libjags.so.4: cannot open shared object file: No such file or directory
#7 0.427 Execution halted
#7 ERROR: process "/bin/sh -c R -s -e \"library(rjags)\"" did not complete successfully: exit code: 1
```

This same situation can be seen with renv:

```dockerfile
FROM rstudio/r-base:4.3-jammy

RUN apt-get update && apt-get -y install pkg-config

ARG REPOSITORY=https://packagemanager.rstudio.com/cran/__linux__/jammy/latest

RUN mkdir /content && echo "library(rjags)" > /content/script.R

WORKDIR /content

RUN R -s -e "install.packages('renv', repos=c(CRAN='${REPOSITORY}'))"
RUN R -s -e "renv::init(bare = TRUE, repos=c(CRAN='${REPOSITORY}'))"
RUN R -s -e "renv::install('rjags', repos=c(CRAN='${REPOSITORY}'))"
RUN R -s -e "renv::snapshot()"
RUN R -s -f "script.R"
```

The error:

```
 > [9/9] RUN R -s -f "script.R":
#0 0.537 Loading required package: coda
#0 0.580 Error: package or namespace load failed for ‘rjags’:
#0 0.580  .onLoad failed in loadNamespace() for 'rjags', details:
#0 0.580   call: dyn.load(file, DLLpath = DLLpath, ...)
#0 0.580   error: unable to load shared object '/root/.cache/R/renv/cache/v5/R-4.3/x86_64-pc-linux-gnu/rjags/4-14/dc00e7bcbec8cd953930ec12fc9c8819/rjags/libs/rjags.so':
#0 0.580   libjags.so.4: cannot open shared object file: No such file or directory
#0 0.580 Execution halted
```

Part of this is rooted in how R handles package installation: The `R CMD install` command installs source and binary packages differently.

R CMD install source/binary switch:
https://github.com/wch/r-source/blob/3a15e29d1b5a94e06b99cb9315c562fd8f71bc5b/src/library/tools/R/install.R#L419-L422

R CMD install package load test (inside do_install_source):
https://github.com/wch/r-source/blob/3a15e29d1b5a94e06b99cb9315c562fd8f71bc5b/src/library/tools/R/install.R#L1814-L1852
https://github.com/wch/r-source/blob/3a15e29d1b5a94e06b99cb9315c562fd8f71bc5b/src/library/tools/R/install.R#L1854-L1861

R CMD install do_install_binary does not attempt load:
https://github.com/wch/r-source/blob/3a15e29d1b5a94e06b99cb9315c562fd8f71bc5b/src/library/tools/R/install.R#L480-L502

When first added, only to the source path (and used library, not external process):
https://github.com/wch/r-source/commit/7bc912f48ad621d223bf26106d8e4ddcf405921a

Could renv attempt to validate binary packages by loading them before finalizing the installation? That would let us identify problems at restore-time rather than downstream when running the content.

kevinushey:
Closes #1275.


R/install.R
```diff
@@ -582,6 +582,16 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
   else
     r_cmd_install(package, path)
 
+  # if we just installed a binary package, check that it can be loaded
+  # (source packages are checked by default on install)
+  if (isbin) {
+    status <- catch(renv_install_test(package))
```
kevinushey, line 588: Perhaps we should only do this step on Linux? Or is there some fear that binaries from Windows or macOS could fail to load in some scenarios?

R/install.R
```diff
@@ -590,6 +600,36 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
 
 }
 
+renv_install_test <- function(package) {
```
kevinushey, line 603: We should respect `--no-test-load` or something similar here.

R/install.R
```diff
@@ -590,6 +600,36 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
...
+  rlibs <- paste(renv_libpaths_all(), collapse = .Platform$path.sep)
+  renv_scope_envvars(R_LIBS = rlibs, R_LIBS_USER = "NULL", R_LIBS_SITE = "NULL")
+
+  # make sure R_TESTS is unset here, just in case
+  # https://github.com/wch/r-source/blob/1c0a2dc8ce6c05f68e1959ffbe6318a309277df3/src/library/tools/R/install.R#L76-L79
+  renv_scope_envvars(R_TESTS = NULL)
+
+  # the actual code we'll run in the other process
+  code <- substitute({
+    setTimeLimit(60)
```
kevinushey, line 615: Or should we just rely on R's own timeout facilities? I think that's only available in recent-ish versions of R.

tests/testthat/test-retrieve.R
```diff
@@ -233,10 +237,25 @@ test_that("explicit path to binary packages work", {
...
 
+  renv_tests_scope()
+  renv_scope_local()
+
+  # make sure we have a binary package in the cellar to test with
+  srcpath <- renv_tests_path("local/skeleton/skeleton_1.0.1.tar.gz")
+  binpath <- renv_tests_path("local/skeleton/skeleton_1.0.1.tgz")
+  defer(unlink(binpath))
+
+  # create the binary
```
kevinushey, line 248: The old binary was actually built with an older version of R, and so would fail to load when the test suite was run with a newer version of R.

R/install.R
kevinushey: (This might theoretically be relevant depending on how binaries get built on PPM?)

aronatkins: It is possible to "accidentally" use the wrong binary package in any environment.

From https://bugs.r-project.org/show_bug.cgi?id=18549:

```r
lib <- file.path(getwd(), "lib")
dir.create(lib)
.libPaths(c(lib, .libPaths()))
install.packages("https://cran.r-project.org/bin/macosx/big-sur-arm64/contrib/4.2/Rcpp_1.0.10.tgz")
# installation is permitted without error
library(Rcpp)
#> Error: package or namespace load failed for ‘Rcpp’ in dyn.load(file, DLLpath = DLLpath, ...):
#>  unable to load shared object '/Users/aron/rcpp/lib/Rcpp/libs/Rcpp.so':
#>   dlopen(/Users/aron/rcpp/lib/Rcpp/libs/Rcpp.so, 0x0006): tried: '/Users/aron/rcpp/lib/Rcpp/libs/Rcpp.so' (mach-o file, but is an incompatible architecture (have (arm64), need (x86_64h)))
#> In addition: Warning message:
#> package ‘Rcpp’ was built under R version 4.2.0
```

That said, my experience is mostly with Linux packages and these problems occurring because folks cross-wire distribution binaries or lack system dependencies.

R/install.R
```diff
@@ -590,6 +600,46 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
...
+  opts <- r_cmd_install_option(package, c("install.opts", "INSTALL_opts"), FALSE)
+  if (is.character(opts)) {
+    flags <- unlist(strsplit(opts, "\\s+", perl = TRUE))
+    if ("--no-test-load" %in% flags)
+      return(TRUE)
+  }
+
+  # make sure we use the current library paths in the launched process
+  rlibs <- paste(renv_libpaths_all(), collapse = .Platform$path.sep)
+  renv_scope_envvars(R_LIBS = rlibs, R_LIBS_USER = "NULL", R_LIBS_SITE = "NULL")
```
aronatkins, line 615: The reference code in R also clears `R_ENVIRON_USER`:

https://github.com/wch/r-source/blob/1c0a2dc8ce6c05f68e1959ffbe6318a309277df3/src/library/tools/R/check.R#L273-L276

R/install.R
```diff
@@ -590,6 +600,46 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
...
+
+  # the actual code we'll run in the other process
+  code <- substitute({
+    setTimeLimit(60)
+    options(warn = 1)
+    library(package)
+  }, list(package = package))
+
+  # write it to a tempfile
+  script <- renv_scope_tempfile("renv-install-")
```
aronatkins, line 629: I really like the addition of all these scoped helpers. It's liberating to stop second-guessing "what about the cleanup!?!"

R/install.R
aronatkins: What about adding an option to let folks disable this step, then we can control the per-OS default and let folks override that default choice if they choose.

aronatkins (approved):

R/install.R
```diff
@@ -590,6 +600,46 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
 
 }
 
+renv_install_test <- function(package) {
```
aronatkins, line 603: This feels like a function that can be called from a vendored renv; does that sound right? I'm not seeing any renv-specific logic/settings that would cause trouble to packrat.

Given that, I'm inclined to abandon my packrat-equivalent implementation and update our vendored renv once this merges.

Any reason to have a separate implementation in packrat?

Should we annotate this function somehow to indicate that it is a semi-public interface?

R/install.R
kevinushey: Right now you can opt out with `options(INSTALL_opts = "--no-test-load")`, which is what you would normally do for source package installation as well.

R/install.R
kevinushey: Thanks; I missed that!

R/install.R
kevinushey: Yep, this should be safe to call from a vendored `renv`.

> Should we annotate this function somehow to indicate that it is a semi-public interface?

Any thoughts on how we should do this? Just a comment, or something slightly more assertive?

R/install.R
aronatkins: That's fair. I was thinking that the load-test of binary packages might want to be controlled by a separate toggle, since it's a new feature and different than the default behavior in R.

I was also thinking about the Packrat side, where we will want a global toggle, just to give us an "out" in case something unexpected appears in some external installation.

R/install.R
aronatkins: I don't know that we have much precedent. In the case of packrat, we decided to export `.snapshotImpl` because it was directly called by rsconnect, and having it exported was a reminder that it was part of the public interface.

That said, we don't perfectly follow this pattern. Connect calls `packrat:::appDataDir`, for example, which we never exported. Packrat calls `renv:::download`, which is another example. There is no distinction between "internal" and "public" calls when using the vendored renv.

Mostly, I'm asking for "future you" so we can somehow mark that internal functions are treated as public by the vendoring packages. I don't have any real preference about how to indicate these semi-public functions, but I would hate to end up in a situation where we worry about _all_ of renv effectively being treated as public.

hadley (approved):

aronatkins:
If this PR continues to iterate, could we pull in `main`? I want to experimentally vendor this change into Packrat and play with it over there, and it sounds like I need https://github.com/rstudio/renv/pull/1409.

kevinushey:
Sure, will do.

kevinushey:
Rebased on top of main.

R/autoload.R
```diff
@@ -37,7 +37,7 @@ renv_autoload_impl <- function() {
 
   # check if we're disabled
   enabled <- Sys.getenv("RENV_AUTOLOAD_ENABLED", unset = "TRUE")
-  if (falsy(enabled, FALSE))
+  if (enabled)
```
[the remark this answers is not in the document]
kevinushey:
```
> as.logical("True")
[1] TRUE
```

too clever?

R/dependencies.R
```diff
@@ -471,9 +471,11 @@ renv_dependencies_discover_impl <- function(path) {
...
 
-  result <- tryCatch(callback(path), error = warning)
-  if (inherits(result, "condition"))
+  result <- catch(callback(path))
+  if (inherits(result, "error")) {
+    warning(result)
     return(NULL)
+  }
 
   result
```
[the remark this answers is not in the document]
kevinushey: I still prefer returning and handling the condition object directly. I think the reason I don't like this form is you can't return from the scope where `tryCatch()` was called directly from the handler, e.g.

```
> test <- function() {
+   tryCatch(
+     stop("ouch"),
+     error = function(e) {
+       return(NULL)
+     }
+   )
+   print("i got here")
+ }
> 
> test()
[1] "i got here"
```

Whereas, if you just return and handle the condition object you get to handle it with "regular" control flow. In other words, you need to test the result of the tryCatch handler anyways, so I find it easier to just return and handle the condition object up front.

In a perfect world we'd be able to do something like:

```
.[result, error] <- try { something() }
```

and then `error` would be non-NULL if an error occurred.

R/install.R
kevinushey: That's a good idea. I like escape hatches.

R/autoload.R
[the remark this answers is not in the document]
aronatkins: Too clever. Let's say I do:

```bash
RENV_INSTALL_TEST_LOAD=1
```

Then, we get:
```
#15 0.854 Error in if (enabled) return(TRUE) :
#15 0.854   argument is not interpretable as logical
#15 0.854 Calls: <Anonymous> ... playActions -> installPkg -> <Anonymous> -> <Anonymous>
#15 0.854 Execution halted
```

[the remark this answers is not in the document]
kevinushey: Yeah; I figured that was going to cause trouble. I'll fix it.

R/install.R
```diff
@@ -582,6 +582,16 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
...
     r_cmd_install(package, path)
 
+  # if we just installed a binary package, check that it can be loaded
+  # (source packages are checked by default on install)
+  if (isbin) {
+    status <- catch(renv_install_test(package))
+    if (inherits(status, "error")) {
+      unlink(installpath, recursive = TRUE)
+      stop(status)
+    }
```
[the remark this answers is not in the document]
aronatkins: Another thought -- have `renv_install_test` return `TRUE` when it does work and validates the package, `FALSE` when it does no work, and throw an error when it fails to validate the package.

R/install.R
```diff
@@ -590,6 +600,54 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
 
 }
 
+renv_install_test <- function(package) {
+
+  # add escape hatch, just in case
+  enabled <- Sys.getenv("RENV_INSTALL_TEST_LOAD", unset = renv_platform_linux())
```
aronatkins, line 606: This is backwards! 🙈 

```r
enabled <- renv_platform_linux()
if (enabled) {
    return(TRUE)
}
```

aronatkins:
I have a proof-of-concept integration of this change with packrat. I needed to bypass the `RENV_INSTALL_TEST_LOAD` logic, but can generate results like the following (occurring within a Docker container):

```
#15 [installation 7/8] RUN R -s -e "packrat::restore()"
#15 0.764 Installing coda (0.19-4) ...
#15 1.254 	OK (downloaded binary - local path)
#15 1.254 Installing packrat (0.9.1-1) ...
#15 5.968 	OK (built source)
#15 5.969 Installing rjags (4-14) ...
#15 6.444 /opt/R/4.3.0/lib/R/bin/R --vanilla -s -f '/tmp/RtmprBMzpk/renv-install-76b402047'
#15 6.444 ================================================================================
#15 6.444
#15 6.444 Loading required package: coda
#15 6.444 Error: package or namespace load failed for ‘rjags’:
#15 6.444  .onLoad failed in loadNamespace() for 'rjags', details:
#15 6.444   call: dyn.load(file, DLLpath = DLLpath, ...)
#15 6.444   error: unable to load shared object '/content/packrat/lib/x86_64-pc-linux-gnu/4.3.0/rjags/libs/rjags.so':
#15 6.444   libjags.so.4: cannot open shared object file: No such file or directory
#15 6.444 Execution halted
#15 6.480
#15 6.480 LOAD ERROR
#15 6.480
#15 6.480 Error: error testing if 'rjags' can be loaded [error code 1]
#15 6.480 Execution halted
```

The "downloaded binary - local path" output is part of my change to make sure that we were following the path that I expected. It won't remain.

R/install.R
```diff
@@ -590,6 +600,54 @@ renv_install_package_impl <- function(record, quiet = TRUE) {
...
+  writeLines(deparse(code), con = script)
+
+  # check that the package can be loaded in a separate process
+  renv_system_exec(
+    command = R(),
+    args    = c("--vanilla", "-s", "-f", renv_shell_path(script)),
+    action  = sprintf("testing if '%s' can be loaded", package)
+  )
+
+  TRUE
```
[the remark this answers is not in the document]
kevinushey: I think what I really want is a way to return FALSE, as well as some information about the error that occurred.

Right now, errors in `renv_system_exec()` are printed like so:

https://github.com/rstudio/renv/blob/47e0d46d489057de87f1e004d560ef7a11cdbadb/R/system.R#L54-L74

but I wonder if that should instead be part of the error message itself that gets thrown. Still ... for a separate PR, maybe?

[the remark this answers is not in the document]
kevinushey: I feel like part of the reason I went this route is because R truncated long error messages by default, though.

R/install.R
kevinushey: Ugh! Thanks for catching that.

PR #1410, opened 2023-06-08
Document and test auto-snapshotting

### LINKED ISSUE #1249: Add tests for automatic snapshots
As described in https://github.com/rstudio/renv/blob/feature/project-rationalize/inst/config.yml#L20-L26.

hadley:
Fixes #1249

### DIFF
--- a/R/snapshot-auto.R
+++ b/R/snapshot-auto.R
@@ -11,14 +11,19 @@ renv_snapshot_auto <- function(project) {
   `_renv_snapshot_running` <<- TRUE
   defer(`_renv_snapshot_running` <<- FALSE)
 
+  lockfile <- renv_lockfile_path(project = project)
+  old <- file.info(lockfile)$mtime
+
   # passed pre-flight checks; snapshot the library
   # validation messages can be noisy; turn off for auto snapshot
   status <- catch(renv_snapshot_auto_impl(project))
   if (inherits(status, "error"))
     return(FALSE)
 
-  lockfile <- renv_lockfile_path(project = project)
-  writef("* Automatic snapshot has updated '%s'.", renv_path_aliased(lockfile))
+  new <- file.info(lockfile)$mtime
+  if (old != new)
+    writef("* Automatic snapshot has updated '%s'.", renv_path_aliased(lockfile))
+
   TRUE
 
 }
@@ -36,7 +41,7 @@ renv_snapshot_auto_impl <- function(project) {
 
 }
 
-renv_snapshot_auto_enabled <- function(project) {
+renv_snapshot_auto_enabled <- function(project = renv_project_get()) {
 
   # don't auto-snapshot if disabled by user
   enabled <- config$auto.snapshot()
@@ -44,7 +49,7 @@ renv_snapshot_auto_enabled <- function(project) {
     return(FALSE)
 
   # only automatically snapshot the current project
-  if (renv_project_loaded(project))
+  if (!renv_project_loaded(project))
     return(FALSE)
 
   # don't auto-snapshot if the project hasn't been initialized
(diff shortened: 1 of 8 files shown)


R/snapshot-auto.R
```diff
@@ -11,14 +11,19 @@ renv_snapshot_auto <- function(project) {
...
   # passed pre-flight checks; snapshot the library
   # validation messages can be noisy; turn off for auto snapshot
   status <- catch(renv_snapshot_auto_impl(project))
   if (inherits(status, "error"))
     return(FALSE)
 
-  lockfile <- renv_lockfile_path(project = project)
-  writef("* Automatic snapshot has updated '%s'.", renv_path_aliased(lockfile))
+  new <- file.info(lockfile)$mtime
+  if (old != new)
```
hadley, line 24: Only message if the snapshot has actually changed.

R/snapshot-auto.R
```diff
@@ -36,15 +41,15 @@ renv_snapshot_auto_impl <- function(project) {
...
+renv_snapshot_auto_enabled <- function(project = renv_project_get()) {
 
   # don't auto-snapshot if disabled by user
   enabled <- config$auto.snapshot()
   if (!enabled)
     return(FALSE)
 
   # only automatically snapshot the current project
-  if (renv_project_loaded(project))
+  if (!renv_project_loaded(project))
```
hadley, line 52: Ooops

R/snapshot-auto.R
```diff
@@ -64,18 +69,14 @@ renv_snapshot_auto_enabled <- function(project) {
...
 
-renv_snapshot_auto_update <- function(project) {
+renv_snapshot_auto_update <- function(project = renv_project_get() ) {
 
   # check for enabled
   if (!renv_snapshot_auto_enabled(project = project))
     return(FALSE)
 
-  # get path to project library
-  libpath <- renv_paths_library(project = project)
```
hadley, line 74: `renv_snapshot_auto_enabled()` already checks that a library exists

tests/testthat/helper-setup.R
```diff
@@ -12,6 +12,9 @@ renv_tests_setup <- function(envir = parent.frame()) {
   if (!once())
     return()
 
+  # remove automatic tasks so we can capture explicitly in tests
+  renv_task_unload()
```
hadley, line 16: I think this is the right principle; otherwise running the tests interactively and non-interactively behaves differently. It does mean you need to remember `renv_snapshot_task()` manually, but I don't see a way around that.

R/snapshot-auto.R
kevinushey: `renv_snapshot_auto_update()` also gets called in `renv_snapshot_task()`, which doesn't perform that validation, though, so we could run into an unexpected error if someone deleted the `renv/library` folder with automatic snapshots enabled.

R/snapshot-auto.R
kevinushey: 🙈 🙉

R/snapshot-auto.R
```diff
@@ -64,18 +69,14 @@ renv_snapshot_auto_enabled <- function(project) {
 
 }
 
-renv_snapshot_auto_update <- function(project) {
+renv_snapshot_auto_update <- function(project = renv_project_get() ) {
```
kevinushey, line 72: Would it be worth having this function return a boolean value indicating if the lockfile was updated / changed? (basically thinking of if we should move the `old != new` check a bit further inside the call stack)

kevinushey (approved):

R/snapshot-auto.R
hadley: Oh, I'll delete the duplicated code the other way then.

R/snapshot-auto.R
hadley: I made `renv_snapshot_auto_impl()` return the logical; I think that's what you were thinking.

PR #1412, opened 2023-06-08
Keep source when installing packages

### LINKED ISSUE #522: keep source when installing with system git
If I install a package with `devtools::install_git` I do have source references, when doing `restore` source references are lost. 

Would be nice to have an option to keep the source references (or is there one already?), or maybe even do it automatically for all installed packages by default. Or maybe a `keep_source` per package flag in renv.lock ?

hadley:
Fixes #522

I don't think there's any downside to doing this by default?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `install()` now keeps source when installing packages from source (#522).
+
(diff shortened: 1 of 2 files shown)


kevinushey (approved):

king-of-poppk:
> I don't think there's any downside to doing this by default?

- #1688 
- #1713
- https://github.com/mlr-org/mlr3book/pull/739

PR #1420, opened 2023-06-09
Only use PPM for binary installs

### LINKED ISSUE #927: renv does not honour the type="source" argument when installing from renv but keeps installing binary versions of package
**Expected behaviour:** 

- step 1 is using renv to detect binary packages appropriate for the client system and uses those as default behaviour. 
- step 2 and 3 honour the type="source" argument and stick to download sources

**Actual behaviour**

- step 1 is ok
- step 2 and 3 are installing binaries despite type="source" being set. 

**Reproducible example** 

See below. Run in  docker container `rocker/r-ver:4.1.2` after running
```bash
apt-get update && apt-get install libxt6 libxtst6 pandoc vim
R -q -e "install.packages(c('styler','reprex'))
```

**Note**: Example below uses admin privileges to install renv but if renv is already present this step can easily be omitted. 

``` r
# Set up immutable URL from public RSPM 
options(repos = "https://packagemanager.rstudio.com/all/2022-01-31+MTo3OTU2NzYxLDI6NDUyNjIxNTs5RDdCN0EyMg")

# pre-requisites: setup workdir and install renv into R library

testdir<-paste0(Sys.getenv("HOME"), "/renv-test")
dir.create(testdir)
setwd(testdir)
install.packages("renv", paste0(Sys.getenv("R_HOME"),"/library/"))
#> Updating HTML index of packages in '.Library'
#> Making 'packages.html' ... done

# step 1: Create fresh renv, install SuppDists package and clean up renv in workdir

renv::init()
#> * Initializing project ...
#> * Discovering package dependencies ... Done!
#> * Copying packages into the cache ... Done!
#> * Lockfile written to '~/renv-test/renv.lock'.
renv::install("SuppDists")
#> Retrieving 'https://packagemanager.rstudio.com/all/__linux__/focal/2022-01-31+MTo3OTU2NzYxLDI6NDUyNjIxNTs5RDdCN0EyMg/src/contrib/SuppDists_1.1-9.7.tar.gz' ...
#>  OK [downloaded 331.9 Kb in 2.4 secs]
#> Installing SuppDists [1.1-9.7] ...
#>  OK [installed binary]
#> Moving SuppDists [1.1-9.7] into the cache ...
#>  OK [moved to cache in 1.8 milliseconds]

system("rm -rf renv* .Rprofile")

# step 2: Repeat step 1 with type="source" 

renv::init()
#> * Initializing project ...
#> * Discovering package dependencies ... Done!
#> * Copying packages into the cache ... Done!
#> * Lockfile written to '~/renv-test/renv.lock'.
renv::install("SuppDists", type="source")
#> Installing SuppDists [1.1-9.7] ...
#>  OK [linked cache]

system("rm -rf renv* .Rprofile")

# step 3: Repeat step 2 but additionally cleaning up renv cache befor running renv 

system(paste0("rm -rf ",Sys.getenv("HOME"), "/.cache"))

renv::init()
#> * Initializing project ...
#> * Discovering package dependencies ... Done!
#> * Copying packages into the cache ... Done!
#> * Lockfile written to '~/renv-test/renv.lock'.
renv::install("SuppDists", type="source")
#> Retrieving 'https://packagemanager.rstudio.com/all/__linux__/focal/2022-01-31+MTo3OTU2NzYxLDI6NDUyNjIxNTs5RDdCN0EyMg/src/contrib/SuppDists_1.1-9.7.tar.gz' ...
#>  OK [downloaded 331.9 Kb in 1.1 secs]
#> Installing SuppDists [1.1-9.7] ...
#>  OK [installed binary]
#> Moving SuppDists [1.1-9.7] into the cache ...
#>  OK [moved to cache in 1.6 milliseconds]

system("rm -rf renv* .Rprofile") 
```

<sup>Created on 2022-02-02 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
Fixes #927

An alternative fix would be to change the default for `pkgType` to "both" (or maybe binary?). But I'm not sure what the implications of that would be on linux.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -129,6 +129,7 @@ install <- function(packages = NULL,
   libpaths <- renv_libpaths_resolve(library)
   renv_scope_libpaths(libpaths)
 
+  is_binary <- !identical(type, "source")
   type <- type %||% getOption("pkgType")
   renv_scope_options(pkgType = type)
 
@@ -185,7 +186,7 @@ install <- function(packages = NULL,
   )
 
   # retrieve packages
-  records <- retrieve(names(remotes))
+  records <- retrieve(names(remotes), is_binary = is_binary)
   if (empty(records)) {
     writef("* There are no packages to install.")
     return(invisible(list()))
--- a/R/retrieve.R
+++ b/R/retrieve.R
@@ -4,7 +4,7 @@
 # this routine retrieves a package + its dependencies, and as a side
 # effect populates the restore state's `retrieved` member with a
 # list of package records which can later be used for install
-retrieve <- function(packages) {
+retrieve <- function(packages, is_binary = TRUE) {
 
   # confirm that we have restore state set up
   state <- renv_restore_state()
@@ -15,7 +15,7 @@ retrieve <- function(packages) {
   options(repos = renv_repos_normalize())
 
   # transform repository URLs for PPM
-  if (renv_ppm_enabled()) {
+  if (is_binary && renv_ppm_enabled()) {
     repos <- getOption("repos")
     renv_scope_options(repos = renv_ppm_transform(repos))
   }


R/install.R
```diff
@@ -129,6 +129,7 @@ install <- function(packages = NULL,
   libpaths <- renv_libpaths_resolve(library)
   renv_scope_libpaths(libpaths)
 
+  is_binary <- !identical(type, "source")
```
kevinushey, line 132: You could also do something like `renv_scope_envvars(RENV_PPM_ENABLED = "FALSE")` to avoid needing to pass this along the call stack, and also ensure transformation is handled consistently anywhere else we do this, e.g.

https://github.com/rstudio/renv/blob/47e0d46d489057de87f1e004d560ef7a11cdbadb/R/pak.R#L60-L62

kevinushey:
As an aside, R doesn't let you use `type = "binary"` with a plain `install.packages()`, e.g.

```
> install.packages("rlang", type = "binary")
Error in install.packages("rlang", type = "binary") : 
  type 'binary' is not supported on this platform
```

R/install.R
hadley: Or maybe new `renv_scope_ppm_repos()` which could automatically check `renv_ppm_enabled()`?

kevinushey: Good idea!

kevinushey:
I'm going to try and take a look here in a separate PR.

PR #1421, opened 2023-06-09
vendor: default version and trapped download failures

### LINKED ISSUE #1418: vendor: branch install encounters untar error
```r
remotes::install_github("rstudio/renv", "bugfix/binary-install-test")
renv:::vendor()
#> * Downloading from GitHub ... FAILED
#> Error in utils::untar(tarfile = tarfile, files = files, list = list, exdir = exdir,  : 
#>   invalid 'tarfile' argument
```

aronatkins:
Traps download failures, as we were attempting to `untar(FALSE)` before. Default the vendor version to the current renv version. Print the version when downloading from GitHub.

This lets me successfully:

```
remotes::install_github("rstudio/renv", "aron-vendor-untar-fix")
renv:::vendor()
```

Feel free to discard/reimplement if the intent was some other interaction with `vendor`.

Fixes #1418

Note: Rooted on https://github.com/rstudio/renv/pull/1407 because that's the change I wanted to use.

### DIFF
--- a/R/vendor.R
+++ b/R/vendor.R
@@ -44,7 +44,7 @@ vendor <- function(version    = NULL,
   }
 
   # get renv sources
-  sources <- sources %||% renv_vendor_sources(version)
+  sources <- sources %||% renv_vendor_sources(version %||% renv_metadata_version())
 
   # re-compute renv version from sources
   version <- renv_description_read(path = sources, field = "Version")
--- a/R/vendor.R
+++ b/R/vendor.R
@@ -20,9 +20,10 @@
 #' need to rely on renv internal functions, we strongly recommend testing
 #' your usages of these functions to avoid potential breakage.
 #'
-#' @param version The version of renv to vendor. If `NULL` (the default),
-#'   the current version of renv will be used. Ignored if `sources`
-#'   is non-`NULL`.
+#' @param version The version of renv to vendor. A Git tag targets a CRAN
+#'   version and "main" references the latest development version of renv. Use
+#'   a Git branch name or commit SHA to target other versions. Ignored when
+#'   `sources` is non-`NULL`.
 #'
 #' @param sources The path to local renv sources to be vendored.
 #'
@@ -44,7 +45,12 @@ vendor <- function(version    = NULL,
   }
 
   # get renv sources
-  sources <- sources %||% renv_vendor_sources(version %||% renv_metadata_version())
+  if (is.null(sources)) {
+    if (is.null(version)) {
+      stop("Specify an renv version or a path to renv sources")
+    }
+    sources <- renv_vendor_sources(version)
+  }
 
   # re-compute renv version from sources
   version <- renv_description_read(path = sources, field = "Version")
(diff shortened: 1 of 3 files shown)


R/vendor.R
```diff
@@ -45,7 +45,7 @@ vendor <- function(version    = NULL,
   }
 
   # get renv sources
-  sources <- sources %||% renv_vendor_sources(version)
+  sources <- sources %||% renv_vendor_sources(version %||% renv_metadata_version())
```
kevinushey, line 48: I think this now needs to be a commit SHA rather than a version, now that we're no longer tagging every renv commit. @hadley, what do you think the right approach is? If `version` is NULL, should we just use the SHA associated with the tip of main?

R/vendor.R
```diff
@@ -150,6 +150,9 @@ renv_vendor_imports <- function() {
 renv_vendor_sources <- function(version = renv_metadata_version()) {
 
   tarball <- renv_bootstrap_download_github(version)
```
kevinushey, line 152: Maybe we should alter this function so that it either returns a path, or emits an error (instead of FALSE)?

R/vendor.R
aronatkins: With these changes, it used e7ec59ff4a92ec15583ac60faa1100f5b490af34 after I had installed this branch with remotes.

```r
renv:::renv_metadata_version()
#> [1] "e7ec59ff4a92ec15583ac60faa1100f5b490af34"
``` 

From the DESCRIPTION:

```
RemoteSha: e7ec59ff4a92ec15583ac60faa1100f5b490af34
```

There is definitely a question about how folks are going to use vendor... will they use vendor to incorporate the code for the version of renv installed locally, or should we require a remote reference? If we have a remote reference, then there is a risk of that remote version needing some alternate strategy for composing all the code.

I don't think that the vendor function can be used with old renv, for example.

I would prefer if `renv:::vendor` lost its inferred argument values; force us to provide values.

R/vendor.R
aronatkins: Any option is better than attempting to `untar(FALSE)` 🤣

Do you want to make those changes separately or do you want me to iterate, here?

R/vendor.R
hadley: IMO I think we should just default it to `main` — that's likely what you will want most of the time when re-vendoring anyway.  But I'm also fine with leaving version with no default so that you have to pick a sha yourself.

R/vendor.R
hadley: I think that's hard to do without changing how `renv_bootstrap_download()` works (along with all other `renv_bootstrap_download_*` functions). It's a bit icky, but I think given that the vendor code is used by only a few people, we should just go with the code in this PR.

R/vendor.R
aronatkins: I've made (version or sources) required.

kevinushey (approved):

hadley (approved):

PR #1423, opened 2023-06-09
Refactor `renv_package_checking()`

### LINKED ISSUE #1298: improve handling of renv state directories in R CMD check
From Professor Ripley:

> (On macoS as their directories are reported as zero size):
> 
> du -h /Users/ripley/Library/Caches/org.R-project.R/R
>    0B   /Users/ripley/Library/Caches/org.R-project.R/R/dipsaus/file_locks
>    0B   /Users/ripley/Library/Caches/org.R-project.R/R/dipsaus
>    0B   /Users/ripley/Library/Caches/org.R-project.R/R/R.cache/styler/1.9.1
>    0B   /Users/ripley/Library/Caches/org.R-project.R/R/R.cache/styler
>    0B
> /Users/ripley/Library/Caches/org.R-project.R/R/renv/sandbox/R-4.4-r84396/aarch64-apple-darwin22.4.0/d88109f6
>    0B
> /Users/ripley/Library/Caches/org.R-project.R/R/renv/sandbox/R-4.4-r84396/aarch64-apple-darwin22.4.0
>    0B
> /Users/ripley/Library/Caches/org.R-project.R/R/renv/sandbox/R-4.4-r84396
>    0B   /Users/ripley/Library/Caches/org.R-project.R/R/renv/sandbox
>    0B   /Users/ripley/Library/Caches/org.R-project.R/R/renv
> 
> The CRAN policy says
> 
> "packages may store user-specific data, configuration and cache files in
> their respective user directories obtained from tools::R_user_dir(),
> provided that by default sizes are kept as small as possible and the
> contents are actively managed (including removing outdated material)."
> 
> so these empty dirs should have have been removed.  For renv this seems
> to be from some user package, so if this could be encouraged in
> documentarian that would be good.
> 

I think we should try and do a couple things:

1. Have `renv` try and detect if R CMD check is running, and set `RENV_PATHS_ROOT` to an appropriate temporary location. (Do we also need to do this for other cases; like running examples?)
2. Have `renv` try to remove empty directories in `RENV_PATHS_ROOT` at some opportune time. Not sure when that is; maybe on process exit? Do we always do that, or only in certain scenarios?

hadley:
To verify everything looks ok, I created https://github.com/hadley/renvdep, which errors with `renv:::renv_paths_root()` in both an example and a test. That confirms the root path uses the temp directory, which I think should be enough to close #1298.

### DIFF
--- a/R/migrate.R
+++ b/R/migrate.R
@@ -63,10 +63,6 @@ migrate <- function(
   project <- renv_project_resolve(project)
   renv_project_lock(project = project)
 
-  # for jetpack
-  if (renv_package_checking())
-    renv_patch_repos()
-
   project <- renv_path_normalize(project, mustWork = TRUE)
   if (file.exists(file.path(project, "packrat/packrat.lock"))) {
     packrat <- match.arg(packrat, several.ok = TRUE)
--- a/R/package.R
+++ b/R/package.R
@@ -396,11 +396,6 @@ renv_package_checking <- function() {
 
 renv_package_checking_impl <- function() {
 
-  # check for devtools
-  calls <- sys.calls()
-  if (identical(calls[[1L]], quote(devtools::test())))
-    return(TRUE)
-
   # otherwise, check other things
   is_testing() ||
     "CheckExEnv" %in% search()
--- a/R/package.R
+++ b/R/package.R
@@ -398,7 +398,7 @@ renv_package_checking_impl <- function() {
 
   # otherwise, check other things
   is_testing() ||
-    "CheckExEnv" %in% search()
+    "CheckExEnv" %in% search() ||
     renv_envvar_exists("_R_CHECK_PACKAGE_NAME_") ||
     renv_envvar_exists("_R_CHECK_SIZE_OF_TARBALL_")
 }
--- a/R/package.R
+++ b/R/package.R
@@ -391,12 +391,6 @@ renv_package_built <- function(path) {
 }
 
 renv_package_checking <- function() {
-  memo(renv_package_checking_impl())
-}
-
-renv_package_checking_impl <- function() {
-
-  # otherwise, check other things
   is_testing() ||
     "CheckExEnv" %in% search() ||
     renv_envvar_exists("_R_CHECK_PACKAGE_NAME_") ||
--- a/R/paths.R
+++ b/R/paths.R
@@ -186,7 +186,7 @@ renv_paths_root <- function(...) {
 # nocov start
 renv_paths_root_default <- function() {
 
-  `_renv_root` <<- `_renv_root` %||% {
+  (`_renv_root` <<- `_renv_root` %||% {
 
     # use tempdir for cache when running tests
     # this check is necessary here to support packages which might use renv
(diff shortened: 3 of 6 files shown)


R/migrate.R
```diff
@@ -63,10 +63,6 @@ migrate <- function(
   project <- renv_project_resolve(project)
   renv_project_lock(project = project)
 
-  # for jetpack
-  if (renv_package_checking())
-    renv_patch_repos()
```
hadley, line 68: This doesn't do anything because in `renv_patch_repos()`:

https://github.com/rstudio/renv/blob/8bff10a3cca3ad1eeecf43ae3ee5f0b6f9dbe44b/R/patch.R#L168-L171

R/package.R
```diff
@@ -391,19 +391,8 @@ renv_package_built <- function(path) {
...
-
-  # check for devtools
-  calls <- sys.calls()
-  if (identical(calls[[1L]], quote(devtools::test())))
-    return(TRUE)
-
-  # otherwise, check other things
   is_testing() ||
-    "CheckExEnv" %in% search()
+    "CheckExEnv" %in% search() ||
```
hadley, line 395: OOPS

R/package.R
```diff
@@ -391,19 +391,8 @@ renv_package_built <- function(path) {
 }
 
 renv_package_checking <- function() {
-  memo(renv_package_checking_impl())
```
hadley, line 394: This is fast enough and called in few enough places that I don't think there's an need to memoise, and removing the cache eliminates a possible failure mode.

R/paths.R
```diff
@@ -186,7 +186,7 @@ renv_paths_root <- function(...) {
 # nocov start
 renv_paths_root_default <- function() {
 
-  `_renv_root` <<- `_renv_root` %||% {
+  (`_renv_root` <<- `_renv_root` %||% {
```
hadley, line 189: So it prints when you call `renv_paths_root_default()`

R/migrate.R
kevinushey: I don't understand; I think the conditionals do match up here? This was to workaround a bug in the jetpack package which caused `renv` CRAN submissions to fail, so you probably want to be careful about modifying this code (it could break revdeps)

R/package.R
kevinushey: 😭

R/restore.R
```diff
@@ -53,9 +53,6 @@ restore <- function(project  = NULL,
 
   renv_activate_prompt("restore", library, prompt, project)
 
-  if (renv_package_checking())
```
kevinushey, line 56: This might be needed for jetpack revdeps.

kevinushey (approved):

R/migrate.R
hadley: Oh doh, I was just confused. I'll just remove the conditionals instead.

hadley: Oh actually, I don't think these are needed any more since `renv_patch_repos()` is called from `renv_patch_init()` which is called from `renv_zzz_load()` which is called `.onLoad()`.

kevinushey:
LGTM; we'll just want to keep jetpack in mind when we run revdep checks later.

PR #1424, opened 2023-06-09
use abort for system errors

kevinushey:
This draws inspiration from `rlang::abort()`, and seeks to enhance how `renv` handles errors in a few locations.

In particular, this gives us a way for callers to catch and silence errors from `renv_system_exec()`, or to allow those messages to be emitted if required.

### DIFF
--- a/R/abort.R
+++ b/R/abort.R
@@ -1,0 +1,29 @@
+
+abort <- function(message, ..., details = NULL, class = NULL) {
+
+  # create condition object
+  cnd <- if (is.character(message)) {
+    cnd <- list(message = cnd, details = details, ...)
+    class(cnd) <- c(class, "error", "condition")
+  } else if (inherits(message, "condition")) {
+    message
+  } else {
+    stop("internal error: abort called with unexpected message")
+  }
+
+  # signal the condition, giving calling handlers a chance to run first
+  signalCondition(cnd)
+
+  # if we get here, the condition wasn't handled -- handle printing of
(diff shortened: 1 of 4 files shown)


R/abort.R
```diff
@@ -0,0 +1,29 @@
...
+  # the error ourselves, and then stop with a fallback condition
+  if (length(cnd$details))
+    writeLines(c(cnd$details, ""))
+
+  # create the fallback, but 'dodge' the existing error handlers
+  fallback <- cnd
+  class(fallback) <- "condition"
+
+  # now throw the error
+  stop(fallback)
```
hadley, line 27: I worry about the implications of `stop`ping with a condition that doesn't have an error class.

I don't quite get what you mean by avoiding existing error handlers, because if we're at this point of the code then we know there wasn't a handler on the stack?

kevinushey: This is how `rlang` works (although it keeps an `rlang_error` class), e.g.

https://github.com/r-lib/rlang/blob/main/R/cnd-abort.R#L852-L858

As I understand it, this is necessary to avoid running active calling handlers twice, e.g. in

```
withCallingHandlers(
  abort("ouch"),
  error = function(e) print("I saw an error")
)
```

since that calling handle would've seen the error from the initial `signalCondition()`, without altering control flow.

I believe the error handler installed by `options(error)` gets to handle the error regardless of its class.

hadley: Ah, that makes sense.

kevinushey: Done!

PR #1426, opened 2023-06-10
use watchdog to help remove orphaned locks

### LINKED ISSUE #1261: improve handling of orphaned locks in available packages index
`renv` uses a helper `index()` function for the available packages index, basically allowing the retrieved copy of available packages to be shared across time and across processes. This helps avoid some painful slowness when retrieving available packages is slow, especially in a session that is restarted often.

A simple reproducible example:

```
local({
  renv:::renv_scope_lock("test")
  tools:::pskill(Sys.getpid(), 9L)
})
```

You'll end up with an orphaned lock which cannot be recovered. The best option I can think of (currently) would be to spawn a separate watchdog process that waits until the R session exits, and then deletes the lockfile if it exists and was owned by that process.

Then, each process could generate a unique key (UUID?) on startup, use that key when taking ownership of a lock, and also let the watchdog know about that key so that if the watchdog sees the process exits, it could check if the lock exists with that key and remove it if so.

kevinushey:
Closes #1261.

To give a short summary, `renv` processes might need to maintain file locks when mutating shared resources; e.g. an index of available R packages. However, those file locks could become orphaned if a process were to die or be killed while holding that lock.

This PR attempts to alleviate that by giving each process using `renv` a watchdog process, which tries to ensure those locks are cleaned up if the associated process unexpectedly dies.

I'm definitely open to alternatives though; if we were able to take advantage of POSIX file locking here, that would probably simplify things a great deal -- however, I'm not sure if we have access to those with only base R facilities.

It's also possible there are other ways forward I haven't thought of...


R/watchdog.R
```diff
@@ -0,0 +1,128 @@
...
+  if (getRversion() < "4.0.0")
+    return(FALSE)
+
+  # skip if disabled
+  enabled <- Sys.getenv("RENV_WATCHDOG_ENABLED", unset = "TRUE")
+  if (!truthy(enabled))
+    return(FALSE)
+
+  # skip during R CMD check (but not when running tests)
+  checking <- renv_envvar_exists("_R_CHECK_PACKAGE_NAME_")
```
hadley, line 35: Use `renv_package_checking()`?

R/watchdog.R
```diff
@@ -0,0 +1,128 @@
...
+
+renv_watchdog_start <- function() {
+
+  # create socket for reading initialization data
+  repeat {
+    port <- sample(49152:65536, size = 1L)
+    socket <- tryCatch(serverSocket(port), error = identity)
+    if (!inherits(socket, "error"))
+      break
+  }
```
hadley, line 64: For safety, I think you should terminate this after some number of tries, because if `serverSocket(port)` systematically errors for some reason, you'll get an infinite loop.

R/watchdog.R
```diff
@@ -0,0 +1,128 @@
...
+    wait = FALSE
+  )
+
+  # wait for connection from watchdog server
+  conn <- socketAccept(socket, blocking = TRUE)
+  defer(close(conn))
+
+  # read initialization data
+  `_renv_watchdog_metadata` <<- unserialize(conn)
+
```
hadley, line 97:
```suggestion
  invisible()
  
```

Just to signal that this is a function called for its side effects?

R/watchdog.R
```diff
@@ -0,0 +1,128 @@
...
+  # establish connection
+  port <- renv_watchdog_port()
+  conn <- socketConnection(port = renv_watchdog_port())
+  on.exit(close(conn), add = TRUE)
+
+  # write data
+  serialize(
+    list(method = method, data = data),
+    connection = conn
+  )
```
hadley, line 111: Does this need a `tryCatch()` in case the process somehow died in the mean time?

hadley (approved):

R/watchdog.R
kevinushey: That also checks if `is_testing()` so it doesn't seem like the right call (unless we want to pull that out)

https://github.com/rstudio/renv/blob/a04863eda8b8a871f09e8c51d541f56e691a2566/R/package.R#L405

R/watchdog.R
kevinushey: Good idea; I'll do that.

gaborcsardi:
Do I understand it correctly that the watchdog process is a long running background process, started by the `system(..., wait = FALSE)` call?

kevinushey:
Right; that's correct.

gaborcsardi:
That is tricky, because those processes run in the same process session as the main R process, so they get the same signals as the main R process. E.g. a SIGINT (SIGTERM, etc.) from the terminal will be delivered to the watchdog process as well, interrupting/killing it, possibly before doing any cleanup, no?

kevinushey:
Is that true for processes with wait = FALSE too? I thought they were moved to a separate process group or something... but I should probably verify. I'm mainly worried about things like people killing the main session in RStudio though; I'm not sure if those signals get sent to the whole group or just the one process.

gaborcsardi:
Yes, afair wait=FALSE is the same.

kevinushey:
One thing I still need to figure out: if the watchdog process times out while waiting for a connection, then future attempts to connect and send messages seem to get lost -- like R leaves the connection available for the socket server to try and write to, but drops that data.

gaborcsardi:
I think this is super cool, and frankly I am amazed that it is possible with base R only. 

IDK if the common process group is an issue for you, I _think_ an interrupt from the terminal will be fine because R will catch it, and while it will kill the watchdog process (before it can clean up), the main R process will clean up in the `on.exit()` handlers.

OTOH, if the process (group) gets a SIGTERM (or other unhandled signal), then the lock dir is not cleaned up. E.g. run this in a new terminal window:
```r
do <- function() {
  renv:::renv_scope_lock("/tmp/lock")
  message(paste("lock exists:", file.exists("/tmp/lock")))
  message("watchdog is ", renv:::renv_watchdog_pid())
  Sys.sleep(1000)
}

message("I am ", Sys.getpid())
unlink("/tmp/lock", recursive = TRUE)
message(paste("lock exists:", file.exists("/tmp/lock")))

do()
```
and then close the whole terminal window. The lock dir is not cleaned up:
```
❯ cd ~/works/renv;  R -q -f test.R
> # pkgload::load_all()
>
> do <- function() {
+   renv:::renv_scope_lock("/tmp/lock")
+   message(paste("lock exists:", file.exists("/tmp/lock")))
+   message("watchdog is ", renv:::renv_watchdog_pid())
+   Sys.sleep(1000)
+ }
>
> message("I am ", Sys.getpid())
I am 38902
> unlink("/tmp/lock", recursive = TRUE)
> message(paste("lock exists:", file.exists("/tmp/lock")))
lock exists: FALSE
>
> do()
lock exists: TRUE
watchdog is 38916
```
Close the whole window, and then:
```
❯ ls -ld /tmp/lock
drwxr-xr-x  2 gaborcsardi  wheel  64 Jun 12 14:41 /tmp/lock/
```

If the RStudio build pane's STOP button only kills the R process, and not the whole process group, then it is not an issue there. (OTOH it would probably make sense to kill the whole process group.)

In general, I think a purely filesystem-based solution, that does not have a background process, could be simpler. Managing background processes is hard enough with proper tools, it is really hard with base R. There are a lot parts, synchronization, etc, and a lot of things can go wrong. A filesystem-based solution could use that `link()` (or `symlink()`) are atomic on Unix. Windows would be different, but I think it is probably easier, as Windows is pretty eager with exclusive locks.

kevinushey:
> OTOH, if the process (group) gets a SIGTERM (or other unhandled signal), then the lock dir is not cleaned up. E.g. run this in a new terminal window:

I think you are right 😞 . It's good enough to handle cases like `kill <pid>` from the command line, but terminal interrupts won't be handled, and we probably want to do that.

> In general, I think a purely filesystem-based solution, that does not have a background process, could be simpler.

What do you think the right way to handle "orphaned" locks in this case is? Would we just want a similar scheme where we somewhat regularly bump the write time of lockfiles, and check for "orphaned" lockfiles as those being older than some threshold?

gaborcsardi:
Yeah, I think one way to do this is to use `file.symlink()` to create the lock file, and it would link to a file with a random name, that we create before, and that file or the file name has the pid. Then we can use `psnice()` to check if that process is still running. (Plus optionally some tricks for pid reuse.)

It is still a bit tricky of course when multiple processes are trying to delete the lock file at the same time, but I think there is a way to make it work. Let me try to write it down in a doc.

kevinushey:
> Is there any chance that the different actors (lock-requestors) are on different hosts / within different containers?

No; the watchdog process is just a regular old child process of the "main" session.

kevinushey:
> (Plus optionally some tricks for pid reuse.)

This is the main thing I'd like to know about... I guess we could use command-line tools like `pgrep` or `lsof` to check that if a process exists, it really is an R process, but I'm curious if there's something a bit easier, or cross-platform.

gaborcsardi:
The "standard" id for processes is the (pid, start time) double. The latter is a bit hard to get accurately in base R, but 1-2s precision is enough in this case I think. Then we can get it for the current process as 
```
❯ Sys.time() - as.difftime(proc.time()[["elapsed"]], units = "secs")
[1] "2023-06-12 20:19:30 CEST"
```

And for looking it up for another process we can run this on Unix:
```
~ # ps -o pid,etime
PID   ELAPSED
    1  2:36
   17  0:01
```
This should work with every ps, even with busybox.

gaborcsardi:
Windows would need to have a different implementation, I am afraid. But I expect that to be simpler. Windows is pretty keen on dealing out exclusive locks, and these break when the process terminates.

kevinushey:
I think I still have some concerns about the pid + start time approach, mainly because getting that information on Windows is going to be very slow and expensive, just because of how expensive `system()` is on Windows.

In addition, attempting to check process PIDs etc. is going to fail if the lockfiles exist on a shared NFS, since different machines would have different process tables.

So, I think we're back at square one: the only thing we can try and trust is the lockfile's own metadata e.g. mtime, and so we have to rely on the lock owner to keep that metadata up-to-date to communicate to other processes that the lock is held.

gaborcsardi:
Sure, if you want to handle NFS and such, then indeed you need an active process that keeps the lock alive.

PR #1430, opened 2023-06-12
Show user friendly renv version

### LINKED ISSUE #1425: show package version in load message if renv installed from GitHub
Right now I see:

```
* Project '~/scratch' loaded. [renv 314009b6ab456a413af03cb360239923139e57e3]
```

We should instead do something like:

```
* Project '~/scratch' loaded. [renv 0.17.3-60; sha1: 314009b6]
* Project '~/scratch' loaded. [renv 0.17.3; rstudio/renv@314009b6]
```

hadley:
Fixes #1425


R/metadata.R
```diff
@@ -3,6 +3,16 @@ renv_metadata_version <- function() {
   `_renv_metadata`$sha %||% `_renv_metadata`$version
 }
 
+renv_metadata_version_friendly <- function(metadata = `_renv_metadata`) {
+  ver <- metadata$version
+
+  if (renv_metadata_is_dev(metadata)) {
+    ver <- paste0(ver, "; rstudio/renv@", substr(metadata$sha, 1, 6))
```
kevinushey, line 10:
```suggestion
    ver <- paste0(ver, "; rstudio/renv@", substr(metadata$sha, 1, 7))
```

Just for parity with the way GitHub displays commit hashes?

R/metadata.R
```diff
@@ -3,6 +3,16 @@ renv_metadata_version <- function() {
   `_renv_metadata`$sha %||% `_renv_metadata`$version
 }
 
+renv_metadata_version_friendly <- function(metadata = `_renv_metadata`) {
+  ver <- metadata$version
```
kevinushey, line 7: Is this always the package version, or might it sometimes be the sha? See: https://github.com/rstudio/packrat/pull/713/files?w=1#diff-485d0d5662ab5285bfe374325ff46ad3bff5a6ef9ea4a38914744270e9a567f5R19

Maybe we need to record both the package version, as well as the commit hash, here?

https://github.com/rstudio/renv/blob/4c88614a182d4dc1e1e9e5a309dc1bed01c5714e/inst/resources/vendor/renv.R#L10-L14

hadley: Yeah, I think so.

hadley: I took a slightly different approach to ensure that the vendored metadata always stays in sync with the regular metadata

kevinushey: Good idea!

kevinushey (approved):

Issue #934, opened 2022-02-09
Bioconductor packages on github not installing dependencies


macroscian:
If I attempt to load a github repository of a Bioconductor package, e.g.
```
renv::install("mikelove/DESeq2")
```
I get a series of unavailable (bioconducor) packages - I can install these one-by-one, but is there a way I can force it do recognise that these dependencies do exist? I can do an `install("bioc::DESeq2")` and then install the github on top of that, but my _real_ problem is that I want to install my own bioconductor-dependent package which is only on github and not in the bioconductor repos, and would like a quick way of creating an environment for that - is there an easy way to do that?

```renv::install("mikelove/DESeq2")
Retrieving 'https://api.github.com/repos/mikelove/DESeq2/tarball/f722b14ad9793f88e06be0d56d8adf8f1a6e1b53' ...
        OK [downloaded 641 Kb in 0.7 secs]
Error: package 'BiocGenerics' is not available
In addition: Warning messages:
1: package 'BiocGenerics' is not available 
2: no MRAN records available from repository URL '/bin/windows/contrib/4.0' 
Traceback (most recent calls last):
15: renv::install("mikelove/DESeq2")
14: retrieve(names(remotes))
13: handler(package, renv_retrieve_impl(package))
12: renv_retrieve_impl(package)
11: renv_retrieve_github(record)
10: renv_retrieve_package(record, url, path)
 9: renv_retrieve_successful(record, path)
 8: retrieve(package)
 7: handler(package, renv_retrieve_impl(package))
 6: renv_retrieve_impl(package)
 5: records[[package]] %||% renv_retrieve_missing_record(package)
 4: renv_retrieve_missing_record(package)
 3: renv_available_packages_latest(package)
 2: stopf("package '%s' is not available", package)
 1: stop(sprintf(fmt, ...), call. = call.)
```

macroscian:
Your suggestion has solved this and so many other niggles I was having with renvs - many thanks.

kevinushey:
Great, I'm glad to hear it!

I'm going to keep this issue open to track https://github.com/rstudio/renv/issues/934#issuecomment-1035342573; in particular, `renv` should ideally be able to handle this without requiring you to set the `repos` option.

PR #1433, opened 2023-06-12
activate bioc if necessary when resolving deps

kevinushey:
Closes https://github.com/rstudio/renv/issues/934.

### DIFF
--- a/R/retrieve.R
+++ b/R/retrieve.R
@@ -1025,8 +1025,11 @@ renv_retrieve_successful <- function(record, path, install = TRUE) {
     renv_retrieve_handle_remotes(record, subdir = subdir)
 
   # ensure its dependencies are retrieved as well
-  if (state$recursive)
+  if (state$recursive) local({
+    repos <- if (is.null(desc$biocViews)) getOption("repos") else renv_bioconductor_repos()
+    renv_scope_options(repos = repos)
     renv_retrieve_successful_recurse(deps)
+  })
 
   # mark package as requiring install if needed
   if (install)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `renv` now activates the Bioconductor repositories when installing a package
+  from a remote (e.g. GitHub) which declares a Bioconductor dependency (via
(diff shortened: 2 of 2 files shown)


R/retrieve.R
```diff
@@ -1025,8 +1025,11 @@ renv_retrieve_successful <- function(record, path, install = TRUE) {
     renv_retrieve_handle_remotes(record, subdir = subdir)
 
   # ensure its dependencies are retrieved as well
-  if (state$recursive)
+  if (state$recursive) local({
```
kevinushey, line 1028: This could also be a place for us to turn on Additional_repositories?

R/retrieve.R
```diff
@@ -1025,8 +1025,11 @@ renv_retrieve_successful <- function(record, path, install = TRUE) {
...
+    repos <- if (is.null(desc$biocViews)) getOption("repos") else renv_bioconductor_repos()
+    renv_scope_options(repos = repos)
     renv_retrieve_successful_recurse(deps)
+  })
```
hadley, line 1032: I'd personally write this as 

```suggestion
  if (state$recursive && !is.null(desc$biocViews)) { 
    local({
      renv_scope_options(repos = renv_bioconductor_repos())
      renv_retrieve_successful_recurse(deps)
    })
  }
```

I think `renv_bioconductor_repos()` probably needs a comment stating that it prepends to the existing repos; otherwise you have to follow quite a few calls to confirm it.

It also took me a bit of reading to reassure myself that if BiocManager isn't already installed, it will get installed in the right place. (It's also a bit unfortunate we'll install that in the middle of retrieving without any user notification, but I don't see how we can avoid that until we move to pak.)

R/retrieve.R
hadley: Yeah, that'd make sense.

hadley (approved):
I think it's probably worth adding a unit test here, just to ensure that we continue to handle this case after we switch to pak.

R/retrieve.R
kevinushey: But `renv_retrieve_successful_recurse(deps)` needs to run regardless of the value of `biocViews`, hence the implementation I chose here.

Issue #1211, opened 2023-04-27
Change the default behavior to `RENV_PATHS_PREFIX_AUTO=TRUE`


colearendt:
By default, `renv` creates caches that are used (in an incompatible fashion) across differing linux OSes

By setting `RENV_PATHS_PREFIX_AUTO=TRUE`, this can be remedied for different linux operating systems. We should do this by default so that starting projects on different operating systems "just works" on `renv` (modulo package rebuilding).

There may be ways to mitigate a one-time breakage with existing project options, one-time symlinks, etc.

kevinushey:
This would be worth including in a 1.0.0 release, since it'd be a potentially breaking change.

kevinushey:
Punted due to time constraints + danger in making such a breaking change here.

kevinushey:
@HenrikBengtsson that is an excellent idea -- I'll do that!

kevinushey:
This has been implemented. Thanks again for the suggestion @HenrikBengtsson!

PR #1435, opened 2023-06-12
use platform prefix by default

kevinushey:
Closes https://github.com/rstudio/renv/issues/1211.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,13 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* **Breaking Change**: `renv` now includes platform prefix in library paths
+  by default. This is primarily useful on Linux platforms, which can be used
+  to disambiguate library paths for different Linux distributions from one
+  another. To revert to the old behavior, you can set
+  `RENV_PATHS_PREFIX_AUTO = FALSE` in your project `.Renviron` file.
+  See the documentation in `?paths` for more details. (#1211)
+
 * `renv` no longer attempts to query package repositories when checking
   if a project is synchronzied on startup. (#812)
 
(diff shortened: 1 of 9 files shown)


NEWS.md
```diff
@@ -1,6 +1,13 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* **Breaking Change**: `renv` now includes platform prefix in library paths
```
hadley, line 4:
```suggestion
* **Breaking Change**: `renv` now includes platform prefix in cache paths
```
?

hadley (approved):
Isn't this going to break the cache for every existing linux user? Is that ok?

NEWS.md
kevinushey: In fact, the change effects both:

```
> renv:::renv_paths_library()
[1] "/home/kevin/.cache/R/renv/library/renv-3022e2f8/linux-ubuntu-jammy/R-4.1/aarch64-unknown-linux-gnu"
> renv:::renv_paths_cache()
[1] "/tmp/Rtmpoiba2M/renv-tempfile-2ac82649a830d/cache/v5/linux-ubuntu-jammy/R-4.1/aarch64-unknown-linux-gnu"
```

kevinushey:
> Isn't this going to break the cache for every existing linux user? Is that ok?

Yeah, you're right... do you have an opinion on what we should do here? I suppose we could drop a symlink from the "old" cache location to the "new" cache location on startup, to ensure the previous cache location remains visible?

NEWS.md
hadley: Oh so it's not just the cache that will break, but every existing project will need to re-install all packages? I think that's a pretty strong argument for adding some help to automate the transition.

hadley:
Yeah, that was my feeling too.

Issue #841, opened 2021-10-18
Warn/error when snapshotting the development version of a package


daattali:
I have an existing project that I wanted to run on another machine. I called `renv::init()` on the machine A, committed to git, and pulled from machine B. On machine B I ran `renv::restore()` and kept getting errors. It seems that any package that was installed on machine A that was installed from source locally is being considered as if it was from CRAN. For example the latest {shinycssloaders} on CRAN is version 1.0.0 but locally I'm working on a new version so on my machine I had version 1.0.0.9000, and this is what the renv.lock file had:

```
    "shinycssloaders": {
      "Package": "shinycssloaders",
      "Version": "1.0.0.9000",
      "Source": "Repository",
      "Repository": "CRAN",
      "Hash": "08d8f6984ea3ad3a3eeea9f139e68f79"
    },
```

I understand that local sources can be problematic and I don't expect it to work magically, but I do think it should at least warn me and make it clear that this package is NOT from CRAN. When I ran `init()` it should have clearly warned me that there are packages that are from local source.

kevinushey:
The associated logic comes from here:

https://github.com/rstudio/renv/blob/master/R/snapshot.R#L684-L697

What do you think the right solution here is?

daattali:
I don't know if there is currently a way for locally installed packages to work in renv when restored on a different machine.

If yes: I would expect "Source" or "Repository" to be "local" instead of CRAN, and when trying to restore if I don't have the source available on this machine then I should get an explicit error saying package X was installed from local source but is not available.

If not: when doing a snapshot I should get an error of at least a message letting me know package X has been installed from a local source and therefore may not be restoreable for others.

hadley:
Seems like a warning on snapshot would be a nice feature here.

daattali:
I would say that an error would be even better than a warning in this case, because a warning is easy to miss, and something like this seems urgent enough to warrant a fatal error. Perhaps you can have some flag to allow the snapshot to continue even when this error happens.

kevinushey:
Adding a bit of extra context, I think the confusion in `renv` arises because:

- A package called 'shinycssloaders' is indeed available from CRAN,
- However, that particular version of 'shinycssloaders' is not available on CRAN (since it's the development version of the package)

Normally, `renv` tries to determine a package's source based only on the contents of the DESCRIPTION file, but falls back to assuming packages which are available on CRAN can be assumed to be retrieved from CRAN. That's what this fallback is doing:

[...]

PR #1437, opened 2023-06-12
try to infer dev packages when installed from sources

kevinushey:
Closes https://github.com/rstudio/renv/issues/841.


R/snapshot.R
```diff
@@ -791,6 +795,71 @@ renv_snapshot_description_source <- function(dcf) {
...
+
+  # ok, this is a package installed from sources that "looks" like
+  # the development version of a package; try to guess its remote
+  guess <- function(pattern, field) {
+    urls <- strsplit(dcf[[field]] %||% "", "\\s*,\\s*")[[1L]]
+    for (url in urls) {
+      matches <- regmatches(url, regexec(pattern, url, perl = TRUE))[[1L]]
+      if (length(matches) == 3L) {
+        remote <- paste(matches[[2L]], matches[[3L]], sep = "/")
+        return(renv_remotes_resolve(remote))
```
kevinushey, line 842: Do we need to also validate that the version of the package on the remote is ... equal to? "close to"? the locally-installed package?

hadley: I don't think so; I think the goal is just to provide a heuristic that's a bit better than the existing heuristic of using CRAN.

R/snapshot.R
```diff
@@ -791,6 +795,76 @@ renv_snapshot_description_source <- function(dcf) {
...
+  # if the inferred package version appears to be less than the source one,
+  # don't use it
+  if (renv_version_lt(inferred[["Version"]], dcf[["Version"]]))
+    return(dcf)
+
+  # notify the user if we inferred a different package remote
+  if (!identical(dcf, inferred)) {
+    package <- dcf[["Package"]]
+    remote <- renv_record_format_remote(inferred)
+    writef("* '%s' inferred to have been installed from remote '%s'.", package, remote)
```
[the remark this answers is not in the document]
hadley: I think messaging is probably fine for now.

R/snapshot.R
```diff
@@ -791,6 +795,76 @@ renv_snapshot_description_source <- function(dcf) {
...
+renv_snapshot_description_infer_impl <- function(dcf) {
+
+  # if this package appears to have a declared remote, use it
+  for (field in c("RemoteType", "Repository", "biocViews"))
+    if (!is.null(dcf[[field]]))
+      return(dcf)
+
+  # only continue if this appears to be a dev version of a package
+  package <- dcf[["Package"]]
+  version <- unclass(numeric_version(dcf[["Version"]]))[[1]]
```
hadley, line 833: Are you sure you need this block? It would be slightly safer to compare to the CRAN version since not everyone uses the `.9000` convention (including renv 😉), but you're already checking that the inferred version is newer than the source version above.

R/snapshot.R
```diff
@@ -791,6 +795,76 @@ renv_snapshot_description_source <- function(dcf) {
 
 }
 
+renv_snapshot_description_infer <- function(dcf) {
+
+  inferred <- tryCatch(
```
hadley, line 800: Out of interest, why put this here instead of around `renv_remotes_resolve()` below?

R/snapshot.R
```diff
@@ -791,6 +795,76 @@ renv_snapshot_description_source <- function(dcf) {
...
+    renv_snapshot_description_infer_impl(dcf),
+    error = function(err) {
+      warningf("Error inferring package remote: %s", conditionMessage(err))
```
hadley, line 803:
```suggestion
      warningf("Failed to infer remote for %s which was installed from source:\n%s", dcf$Package, conditionMessage(err))
```
?

tests/testthat/test-snapshot.R
```diff
@@ -492,3 +492,26 @@ test_that("autosnapshot works as expected", {
...
+  writeLines(desc, con = descfile)
+
+  remote <- local({
+    renv_scope_options(renv.verbose = FALSE)
+    renv_snapshot_description(path = descfile)
+  })
+
+  expect_equal(remote$RemoteType, "github")
+
+  expect_snapshot(invisible(renv_snapshot_description(path = descfile)))
```
hadley, line 515: I don't know if this is any better, but it's the convention we use elsewhere.

```suggestion
  expect_snapshot(. <- renv_snapshot_description(path = descfile))
```

hadley (approved):

R/snapshot.R
kevinushey: Mainly just to capture any kind of potential failure that could happen.

It also supports my crusade against excessive indentation, or having overly-large bodies in tryCatch blocks.

R/snapshot.R
kevinushey: For some reason, I was thinking we'd want to avoid needing to query CRAN here ... but that's probably fine. I'll make that change.

kevinushey:
@hadley I made a few changes to how we detect "dev" packages; let me know if they look sensible. https://github.com/rstudio/renv/pull/1437/files?w=1#diff-5d849f3782ddfdd6b0053b841b4be729f4fdaa33c9262caf6757aa185e5f5658R851

NEWS.md
```diff
@@ -1,6 +1,10 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `renv` now attempts to infer remote dependencies for GitHub packages that
+  appear to be installed from sources; that is, for packages without a
+  recognizable remote source encoded in its DESCRIPTION file. (#841)
```
jennybc, line 6:
```suggestion
* `renv` now attempts to infer remote dependencies for GitHub packages that
  appear to be installed from sources; that is, for packages without a
  recognizable remote source encoded in the DESCRIPTION file. (#841)
```

At first, I got confused about whose DESCRIPTION file we were talking about. It's not the DESCRIPTION of the dependency, but rather the DESCRIPTION of the renv project.

---

Just to review the basic facts in my use case:

The renv project is a Shiny app I'm deplying to Connect, which has no DESCRPTION or explicit version info of any kind. It also makes no explicit use of renv (but I know that the deployment machinery does).

The app makes use of several packages I have installed locally from source, such as gargle, which does declare GitHub as its source repo in the URL field of its DESCRIPTION file.

jennybc (approved):
LGTM

The only lingering thing I'm wondering about is this: should this only happen in a "pure inference" case? As in, if someone is explicitly using renv, then I think they *should* declare a remote source using proper techniques.

I wonder if the helpful guessing should only happen in a scenario where the person doesn't even know they are using renv (i.e. the deployment machinery is using it on their behalf).

NEWS.md
jennybc: Ah, ok, true. But I don't think many people realize that the installed DESCRIPTION gets modified in that way when you install from GitHub. I.e. info gets inserted that is not in the DESCRIPTION you see in the package source.

I think why I got confused is that I could also have solved my problem by using a DESCRIPTION file for my app, yeah? I could have put `Remotes: r-lib/gargle` there, which is what this PR is inferring.

kevinushey: I don't think so; this PR is explicitly looking at the installed package's DESCRIPTION file, finding that it doesn't provide any metadata about its origin (ie: no fields like `RemoteType`), and so tries to infer that package should be installed from GitHub by its version number + presence of a GitHub URL in BugReports or URL.

hadley: IOTW renv always uses what is _actually_ installed, where as the `DESCRIPTION` declares what should be installed.

kevinushey:
There's been a bit of back and forth in https://github.com/rstudio/renv/pull/1437#discussion_r1229013107; are we happy with the PR in its current form?

hadley:
Yeah, I think so.

PR #1438, opened 2023-06-12
provide exported functions for interacting with lockfiles

kevinushey:
TBH I'm still on the fence on whether we should provide all these methods, or a single function `lockfile` that returns the "API" object as we did before.


inst/resources/activate.R
```diff
@@ -416,7 +416,7 @@ local({
       on.exit(do.call(base::options, saved), add = TRUE)
     }
   
-    catf("* Downloading from GitHub ... ", appendLF = FALSE)
+    catf("* Downloading version %s from GitHub ... ", version, appendLF = FALSE)
```
hadley, line 419: Spurious diff?

R/lockfile.R
```diff
@@ -243,5 +243,19 @@ renv_lockfile_records <- function(lockfile) {
   invisible(x)
 }
 
+renv_lockfile_resolve <- function(lockfile) {
```
hadley, line 246: Not used anywhere?

R/lockfiles.R
```diff
@@ -96,7 +96,80 @@
...
+
+#' @param remotes A named \R list, mapping package names to the remote
+#'   specifications to be recorded in the lockfile.
+#'
+#' @param repos A named vector, mapping \R repository names to their URLs.
+#'
+#' @rdname lockfile
+#' @export
+lockfile_modify <- function(lockfile,
+                            ...,
```
hadley, line 157: Not used?

R/lockfiles.R
```diff
@@ -96,7 +96,80 @@
 #' Note that the `Name` field may be empty. In that case, a project-local Python
 #' environment will be used instead (when not directly using a system copy of Python).
 #'
+#' @inheritParams snapshot
```
hadley, line 99: I'd suggest adding a "for expert use only" caution somewhere.

hadley (approved):
I definitely prefer the individual functions over providing a list of functions as an API.

R/lockfile.R
kevinushey: It does get used here:

https://github.com/rstudio/renv/blob/02a54315df2408efb1ad7216e8a3438eb4ab0e58/R/repair.R#L51
https://github.com/rstudio/renv/blob/02a54315df2408efb1ad7216e8a3438eb4ab0e58/R/restore.R#L59

but in each case that's after it's already been resolved so we can probably just remove those.

R/lockfiles.R
kevinushey: Not used internally, but I thought this might be useful for end users (e.g. a way to explicitly set the repositories in a lockfile)

kevinushey:
> Maybe remove `lockfile-api.R` as part of this PR?

I did indeed do that 🙂

R/lockfiles.R
hadley: Oh I meant the `...`

kevinushey: Oh sorry. That's there intentionally so we can:

1. Force some arguments to be named,
2. Add new arguments (in any order) in a backwards compatible way.

kevinushey: (but maybe I should add `renv_dots_check()` to warn about unused arguments)

hadley: Yeah, I think if you use `...` for this purpose, you need to error if there's anything in them.

kevinushey: Ok; I made that change on main.

PR #1442, opened 2023-06-13
Use scope instead of envir

hadley:


R/scope.R
```diff
@@ -309,60 +309,60 @@ renv_scope_bioconductor <- function(project = NULL,
...
   suppressMessages(eval(call, envir = parent.frame()))
 
-  defer(suppressMessages(untrace(substitute(what))), envir = envir)
+  defer(suppressMessages(untrace(substitute(what))), scope = scope)
 
 }
 
 
-renv_scope_binding <- function(symbol, replacement, frame = renv_envir_self(), envir = parent.frame()) {
+renv_scope_binding <- function(symbol, replacement, frame = renv_envir_self(), scope = parent.frame()) {
```
kevinushey, line 332: Should we rename `frame` to `envir` here?

kevinushey (approved):

PR #1443, opened 2023-06-13
fixup for repair

kevinushey:
https://github.com/rstudio/renv/actions/runs/5259691290/jobs/9505595748


hadley (approved):
Looks like you also need to update `_pkgdown.yml`

PR #1444, opened 2023-06-13
refine detection of PPM URLs

kevinushey:
Closes https://github.com/rstudio/renv/issues/445.


R/ppm.R
```diff
@@ -33,35 +33,34 @@ renv_ppm_transform_impl <- function(url) {
...
-  # any local knowledge of available CRAN repositories)
+  # check if this is an 'ignored' URL; that is, a repository which we
+  # know is not a PPM URL
   mirrors <- catch(getCRANmirrors(local.only = TRUE))
-  urls <- c(urls, mirrors$URL)
-
-  # also ignore some RStudio URLs
-  rstudio <- c(
+  ignored <- c(
+    getOption("renv.ppm.ignoredUrls", default = character()),
```
hadley, line 55: Need to document `renv.ppm.ignoredUrls` somewhere?

aronatkins:
CC @toph-allen - we may want this change.

tests/testthat/test-ppm.R
```diff
@@ -79,3 +79,34 @@ test_that("renv correctly detects RHEL as CentOS for RSPM", {
...
+    renv_condition_signal("bail")
+  ))
+
+  # should exit early
+  status <- tryCatch(
+    renv_ppm_transform_impl("http://foo/bar"),
+    bail = identity
+  )
+
+  expect_identical(status, "http://foo/bar")
```
hadley, line 102:
```suggestion
  status <- expect_no_condition(renv_ppm_transform_impl("http://foo/bar"), class = "bail")
  expect_equal(status, "http://foo/bar")
```
?

tests/testthat/test-ppm.R
```diff
@@ -79,3 +79,34 @@ test_that("renv correctly detects RHEL as CentOS for RSPM", {
...
+
+  # should emit a condition now
+  status <- tryCatch(
+    renv_ppm_transform_impl("http://foo/bar/baz"),
+    bail = identity
+  )
+
+  expect_s3_class(status, "bail")
```
hadley, line 110:
```suggestion
  expect_condition(renv_ppm_transform_impl("http://foo/bar/baz"), class = "bail"))
```

R/url.R
```diff
@@ -0,0 +1,33 @@
+
+renv_url_parse <- function(url) {
```
hadley, line 2: It would make me feel happier if this had a few tests.

hadley (approved):

R/ppm.R
kevinushey: Or should I just add it to config.yml? Or should it be a project-specific setting?

hadley: Seems like project specific might be most useful?

R/url.R
kevinushey: Added a couple -- should I try to chase down any edge cases?

R/ppm.R
kevinushey: I'll do this in a separate PR (I think the other PPM stuff I added should also be project specific)

PR #1448, opened 2023-06-14
Use `the` instead of `_renv_*` and `__self__`

hadley:
Apart from the mechanical changes, the main change is in vendoring, where `renv$initialize()` now calls `renv_zzz_load()` and `renv_zzz_attach()` after setting up the metadata.


inst/resources/vendor/renv.R
```diff
@@ -27,6 +21,13 @@ renv$initialize <- function() {
   script <- system.file("vendor/renv.R", package = .packageName)
   sys.source(script, envir = renv)
 
+  # set up metadata environment
+  renv$the$metadata <- as.environment(..metadata..)
+
+  # run our load / attach hooks so internal state is initialized
+  renv$renv_zzz_load()
+  renv$renv_zzz_attach()
```
hadley, line 29: I think we should consider not running `renv_zzz_attach()` here, since renv is _not_ attached (and it looks like the code currently run there is not relevant to the embedded scenario).

kevinushey (approved):

inst/resources/vendor/renv.R
kevinushey: I think you're right.

PR #1449, opened 2023-06-14
Update activate template on load

hadley:
It's not obvious that you should re-document to recreate this file, whereas re-loading is something you do all the time. I've now also only message if there's actually a change.


R/zzz.R
```diff
@@ -80,18 +80,9 @@ renv_zzz_attach <- function() {
...
-    if (identical(call[[1]], document)) {
-      documenting <- TRUE
-      break
-    }
-  }
-
+  # check if we're in pkgload::load_all()
   # if so, then create some files
-  if (documenting) {
+  if (renv_envvar_exists("DEVTOOLS_LOAD")) {
```
kevinushey, line 85: If only I read the documentation 😭

kevinushey (approved):

Issue #353, opened 2020-02-27
Warn on missing dependencies / Auto-restore


holub008:
Thank you for `renv`!

I created a [new project](https://github.com/holub008/robopico/tree/master/train) and deployed it on a clean machine with only base R installed. When I opened an R shell in the project root, it was great that the `renv` package auto-installed and the session started with correct library paths (as scripted in `renv/activate.R`). 

Given this, I was surprised that the project dependencies themselves weren't auto-installed. It took me a minute of digging around to realize this. Does an auto-hydrate feature play well with `renv`'s philosophy? If not, a warning or user directive may be valuable for first time users. e.g. "Missing dependencies detected! Run `renv::hydrate()` to install.", based on the outcome of `renv::status()`.

kevinushey:
To be clear, `renv::hydrate()` is used to initialize a project, copying the same packages that are currently available in the default R libraries. For an existing project, it's more likely that you'd want to call `renv::restore()` to restore the packages as defined in the `renv.lock` lockfile.

We _could_ auto-restore, but I'd be worried that this could be annoying as `renv::restore()` may be slow if packages aren't already in the cache, and it could also be slow if many packages need to be installed (especially if they must be compiled from sources).

I agree that a compromise where we notify the user when the two are out-of-sync would be best.

holub008:
Thanks for the clarification.

Agreed auto-restore could be slow, and moreover may be an unexpected action by `renv`, given the user's only action was starting a session.

I'll submit a PR for out of sync warning on session start, in case you're interested in adding.

vspinu:
I think this should be a drop-in option. We use a Makefile driven workflow, so each time a team member clones a project we have to remember to first start the session and do `renv:restore()` manually before any other `make xyz`.

vspinu:
To be more precise, I am talking about the auto-restore on bootstrap only, not for follow-up sessions.

holub008:
@vspinu Maybe I don't understand your setup, but why couldn't `renv:restore` be part of an earlier make target?

To your point, though, I agree that having a user config for auto-restore would be nice. So far in my usage (primarily isolated experiments/analyses), I have never check out a fresh `renv` project and *not* wanted dependencies installed for me.

It's worth noting that various dependency management tools handle this differently. For example, pipenv does not auto-install on first run, but maven does (or at least has configs to).

vspinu:
> Maybe I don't understand your setup, but why couldn't renv:restore be part of an earlier make target?

[...]

PR #1451, opened 2023-06-14
Do more bootstrapping in a hook

hadley:
* Remove extraneous newline
* Make restarting with code a bit easier

I tested this interactively in RStudio, and it seems to have better behaviour. Part of #353


R/bootstrap.R
```diff
@@ -903,3 +903,32 @@ renv_bootstrap_version_is_dev <- function(version) {
     length(components) != 3
   }
 }
+
+renv_bootstrap_bootstrap <- function(version, library) {
```
kevinushey, line 907:
```suggestion
renv_bootstrap_run <- function(version, library) {
```

Or something like that?

kevinushey (approved):

PR #1459, opened 2023-06-15
Simplify `dependency()` reporting

hadley:


tests/testthat/_snaps/dependencies.md
```diff
@@ -0,0 +1,17 @@
...
+    Code
+      dependencies(errors = "fatal")
+    Output
+      Finding R package dependencies ... Done!
+      WARNING: One or more problems were discovered while enumerating dependencies.
+      
+      renv-test-ecd07ec4e426
+      ----------------------
+      
+      ERROR 1: directory "renv-test-ecd07ec4e426" contains 28 files; consider ignoring this directory
```
hadley, line 12: I didn't look into this, but it's suspicious that the test writes 26 files and this reports 28.

hadley: The extra files are `dependencies.R` and `renv`, which are not suspicious 😄

R/dependencies.R
```diff
@@ -191,9 +191,8 @@ dependencies <- function(
 renv_dependencies_impl <- function(
   path = getwd(),
   root = NULL,
-  ...,
```
kevinushey, line 194: Can we keep `...` to force following arguments to be named?

R/dependencies.R
```diff
@@ -191,9 +191,8 @@ dependencies <- function(
...
   field = NULL,
-  progress = TRUE,
+  progress = FALSE,
```
kevinushey, line 195: Doesn't this mean we no longer print progress reports for regular usages?

hadley: No, because the default for `dependency()` is still `TRUE`

kevinushey: Ohh got it!

R/dependencies.R
kevinushey: You may be over-estimating my discipline 😅

R/dependencies.R
```diff
@@ -223,12 +222,6 @@ renv_dependencies_impl <- function(
   renv_dependencies_begin(root = root)
   defer(renv_dependencies_end())
 
-  dots <- list(...)
```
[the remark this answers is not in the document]
kevinushey: I think we can drop this one; I don't think it's documented?

kevinushey (approved):

R/dependencies.R
hadley: Ok, I'll replace with `renv_dots_check`.

R/dependencies.R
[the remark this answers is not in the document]
kevinushey: Oh oops. Should we just document it then?

PR #1460, opened 2023-06-15
Better handling of header in snapshots

hadley:


R/utils.R
```diff
@@ -331,7 +331,7 @@ header <- function(label,
                    suffix = "-",
                    n = min(getOption("width"), 78))
 {
-  n <- n - nchar(label) - nchar(prefix) - 2L
+  n <- max(n - nchar(label) - nchar(prefix) - 2L, 3)
```
kevinushey, line 334: This ensures there will always be at least three `---` following the header text?

kevinushey (approved):
LGTM!

R/utils.R
hadley: Yeah, I don't think it makes much difference for real life, but guarantees that there's something unique for the transformer to latch on to.

tests/testthat/helper-snapshot.R
```diff
@@ -45,6 +45,9 @@ strip_dirs <- function(x) {
   x <- gsub(renv_path_aliased(getwd()), "<wd>", x, fixed = TRUE)
   x <- gsub(renv_path_aliased(tempdir()), "<tempdir>", x, fixed = TRUE)
 
+  # Standardise the dashes produced by header()
+  x <- gsub("-{3,}", "---", x, fixed = TRUE)
```
hadley, line 49: I meant this to be `perl = TRUE` 😬

PR #1461, opened 2023-06-15
Rework dependency prompting and init caching

hadley:
* Instead of `renv_dependency_scope()` we now have `renv_dependency_confirm()`
* Only `init()` caches dependency computation, and it's now explicit

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -1655,53 +1655,27 @@ renv_dependencies_report <- function(errors) {
     stopf(fmt)
   }
 
-  renv_condition_signal("renv.dependencies.error", problems)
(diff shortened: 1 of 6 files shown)


R/dependencies.R
```diff
@@ -1655,53 +1655,27 @@ renv_dependencies_report <- function(errors) {
     stopf(fmt)
   }
 
-  renv_condition_signal("renv.dependencies.error", problems)
+  renv_condition_signal("renv.dependencies.problem", problems)
```
hadley, line 1658: I think "problem" is better than "error" because it doesn't imply that it's an error condition.

kevinushey (approved):
LGTM!

Issue #1413, opened 2023-06-08
Changes to `renv.pretty.print.emitter`


juliasilge:
The vetiver package vendors all of renv, and I recently updated to the dev version of renv as outlined in #1253. I have this function that is used to report to users how their current packages are different from their model's packages (for a specific subset of packages relevant to making predictions):

```r
renv_report_actions <- function(current, model) {
    withr::local_options(
        list(renv.pretty.print.emitter = function(text, ...) {cli::cli_inform(text)})
    )
    diff <- renv$renv_lockfile_diff_packages(current, model)

    if (renv$empty(diff))
        return(invisible(NULL))

    lhs <- renv$renv_records(current)
    rhs <- renv$renv_records(model)
    renv$renv_pretty_print_records_pair(
        lhs[names(lhs) %in% names(diff)],
        rhs[names(rhs) %in% names(diff)],
        "The following package(s) do not match your model:",
        "Consider installing the same versions that your model was trained with."
    )

}
```

Everything in here works up to using `renv_pretty_print_records_pair()` (`diff`, `lhs`, and `rhs` are all correct) but then nothing is printed or returned.

The new version of renv doesn't have `renv.pretty.print.emitter` anymore and I am struggling with how to get similar output to what I had before, or actually any output at all now from `renv_pretty_print_records_pair()`. What would be an approach that gets me something similar to what I am doing currently?

kevinushey:
Thank you! Sorry for the trouble in the interim.

PR #1465, opened 2023-06-15
support optional pretty print emitter

kevinushey:
Closes https://github.com/rstudio/renv/issues/1413.

@hadley @juliasilge: this PR brings back the previous behavior; do you think this is sufficient or should we consider an alternate mechanism for customizing how messages are printed?

### DIFF
--- a/R/pretty.R
+++ b/R/pretty.R
@@ -23,9 +23,15 @@ renv_pretty_print <- function(values,
   }
 
   msg$push("")
-  text <- paste(as.character(msg$data()), collapse = "\n")
 
-  # NOTE: Used by vetiver, so is part of the API
+  text <- paste(as.character(msg$data()), collapse = "\n")
+  renv_pretty_print_impl(text)
+
+}
+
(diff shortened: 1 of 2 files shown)


R/pretty.R
```diff
@@ -25,7 +25,13 @@ renv_pretty_print <- function(values,
   msg$push("")
   text <- paste(as.character(msg$data()), collapse = "\n")
 
-  writef(text)
+  # NOTE: Used by vetiver, so is part of the API
+  # https://github.com/rstudio/renv/issues/1413
+  emitter <- getOption("renv.pretty.print.emitter", default = writef)
+  emitter(text)
+
+  invisible(NULL)
```
hadley, line 33: You've used it three times, so probably worth a helper function?

hadley (approved):
I don't have strong feelings either way. I'm also happy to take a look at vetiver and see if there's another way we can work around this, but these changes are already pretty minimal.

PR #1476, opened 2023-06-16
respect type = source with PPM

kevinushey:
Closes https://github.com/rstudio/renv/issues/927.
Supersedes https://github.com/rstudio/renv/pull/1420.


R/repos.R
```diff
@@ -1,20 +1,4 @@
 
-renv_repos_encode <- function(x) {
```
kevinushey, line 2: These were used with a very old version of the lockfile.

PR #1478, opened 2023-06-16
Detect R dependencies from Jupyter notebooks

### LINKED ISSUE #929: .Rprofile crashes R Jupyter notebook
Hi all,

I am using a Jupyter notebook with an R kernel and we want to use `renv` with a particular project. After initializing the env, we find that the presence of .Rprofile in the folder crashes the R kernel in the notebook. If we remove .Rprofile from the folder and activate manually, everything works. This also works if we activate the `renv` manually from a different folder (e.g. using renv::activate("path to folder")), because the .Rprofile file is not there.

Any ideas?

Thank you so much for your time and help

hadley:
Fixes #929


R/dependencies.R
```diff
@@ -908,6 +909,28 @@ renv_dependencies_discover_chunks_ranges <- function(path, contents, patterns) {
...
 }
 
+renv_dependencies_discover_ipynb <- function(path) {
+
+  json <- renv_json_read(path)
+  if (json$metadata$kernelspec$language != "R")
+    return()
+
+  deps <- stack()
+  if (json$metadata$kernelspec$name == "ir")
```
kevinushey, line 919: Same as above?

kevinushey (approved):
LGTM; is it worth being paranoid in case of missing fields?

Issue #890, opened 2021-12-11
simplify bootstrap code


kevinushey:
Right now, `renv` injects a pretty large subset of code into `renv/activate.R` to try and bootstrap itself. We should try to minimize some of this if possible.

In the simplest scenario, we could try to load `renv` from the user library, and then use that to load the associated project.

hadley:
Are you imaging that `template-activate.R` becomes something very roughly like this:

```R
if (requireNamespace("renv", quietly = TRUE)) {
  templib <- tempfile()
  dir.create(templib)

  install.packages("renv", templib)
  if (packageVersion("renv", templib) < "0.18") {
    stop("Failed to find new enough renv")
  }
  loadNamespace("renv", lib.loc = templib)
}

renv:::renv_bootstrap("${VERSION}")
```

That code assumes installing renv from CRAN (or other preconfigured `repos`), but maybe it would make sense to provide a renv only repo? (That could either be used exclusively or as fallback).

kevinushey:
Yeah, exactly. The other main things we might need to support would be:

- Support `RENV_BOOTSTRAP_TARBALL` (https://github.com/rstudio/renv/blob/8b145e8856f3a9e782dd51662b3247662067aa11/R/bootstrap.R#L308-L338)

- Support the various ways that repositories might be set. (I sort of regret allowing so many disparate ways for these to be provided now...)

- Think about whether we still need to support custom download headers in bootstrap.

We might also need to think about supporting the repositories as defined in the `renv.lock` lockfile?

hadley:
So at least for now we probably need to preserve `renv_bootstrap_repos()`, and `renv_bootstrap_download_tarball()`. 

I'd suggest doing this in two steps: we first create `renv_bootstrap()` in `bootstrap2.R` and then move in all the `renv_bootstrap_*` functions that are no longer needed in `bootstrap.R`. Then we can reconsider all the `renv_bootstrap()` functions and if they can be simplified by using other renv tools that were previously out of scope.

We'll also need to solve some meta-bootstrapping problem, in that we'll want to use the new activate script before renv is available on CRAN. I think that is probably most easily solved with a custom repo? Maybe we could use r-universe.dev?

kevinushey:
That sounds like the right approach to me.

PR #1480, opened 2023-06-16
tweaks to bootstrap output display

kevinushey:
Also removes `renv_metadata_is_dev()` as we should be consistently setting both `version` and `sha` in the `renv` metadata. We still need the bootstrap equivalents; I'd like to get rid of that requirement but that's hard to unravel.

### DIFF
--- a/R/bootstrap.R
+++ b/R/bootstrap.R
@@ -39,9 +39,9 @@ bootstrap <- function(version, library) {
   # load failed; inform user we're about to bootstrap
   friendly <- version
   if (nchar(friendly) >= 40)
-    friendly <- sprintf("rstudio/renv@%s", substring(version, 1L, 7L))
+    friendly <- sprintf("[rstudio/renv@%s]", substring(version, 1L, 7L))
 
-  section <- header(paste("Bootstrapping renv", friendly))
+  section <- header(sprintf("Bootstrapping renv %s", friendly))
   catf(section)
 
   # attempt to download renv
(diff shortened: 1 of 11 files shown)


R/bootstrap.R
```diff
@@ -20,27 +16,56 @@ catf <- function(fmt, ..., appendLF = TRUE) {
...
     error = function(err) {
+      catf("FAILED")
       stop("failed to download:\n", conditionMessage(err))
     }
   )
+  catf("OK")
   on.exit(unlink(tarball), add = TRUE)
 
   # now attempt to install
+  catf("- Installing renv  ... ", appendLF = FALSE)
```
hadley, line 60:
```suggestion
  catf("- Installing renv ... ", appendLF = FALSE)
```

hadley: Or is this to line up with downloading?

hadley (approved):
Looks good!

Maybe it's worth doing https://github.com/rstudio/renv/issues/890 just to avoid all the code gymnastics that we currently need?

R/bootstrap.R
kevinushey: Yeah, for nice vertical alignment.

kevinushey:
> Maybe it's worth doing #890 just to avoid all the code gymnastics that we currently need?

I was originally considering that, but it ended up feeling kind of tricky to untangle the ways we're using versions vs. shas around `activate`. I definitely think we should do this before the bootstrap code gets any more unwieldy though!

PR #1484, opened 2023-06-18
install optimizations

kevinushey:
This PR improves the performance of `install()` by ensuring we only snapshot the packages which are actually part of the dependency graph of the packages the user requested be installed.

Closes https://github.com/rstudio/renv/issues/1491.


R/activate.R
```diff
@@ -78,17 +78,15 @@ renv_activate_impl <- function(project,
...
-    setwd(project)
+  if (renv_rstudio_available())
+    renv_rstudio_initialize(project)
 
-    # try to load the project
-    load(project)
-  }
+  # try to load the project
+  setwd(project)
+  load(project)
```
kevinushey, line 89: I needed to change this for interactive testing, as otherwise projects weren't being loaded after `activate` was called.

R/bioconductor.R
```diff
@@ -120,3 +120,35 @@ renv_bioconductor_repos_biocinstaller <- function(version) {
...
+}
+
+renv_bioconductor_augment <- function(records, project) {
+
+  # NOTE: BiocVersion is available on Bioconductor, so we need to ensure
+  # the Bioconductor repositories are activated here
+  sources <- records %>% filter(is.list) %>% extract_chr("Source")
+  if ("Bioconductor" %in% sources) {
+    renv_scope_bioconductor(project = project)
+    packages <- c("BiocManager", "BiocVersion")
```
kevinushey, line 141: Or "BiocInstaller"?

R/lockfile.R
```diff
@@ -193,7 +193,13 @@ renv_lockfile_create_impl <- function(project, libpaths, type, packages, exclude
 
   lockfile <- renv_lockfile_init(project)
 
-  records <- renv_snapshot_libpaths(libpaths = libpaths, project = project)
+  packages <- packages %||% renv_snapshot_dependencies(project, type)
+
+  records <- renv_snapshot_packages(
```
kevinushey, line 198: This is really the meat of the PR; instead of snapshotting the whole library paths and then filtering, we just snapshot the specific packages we need.

R/snapshot.R
```diff
@@ -950,37 +950,78 @@ renv_snapshot_report_actions <- function(actions, old, new) {
 
 renv_snapshot_dependencies <- function(project, type = NULL, dev = FALSE) {
 
-  if (!is.null(the$init_dependencies)) {
```
kevinushey, line 953: This didn't feel right here, since `the$init_dependencies` is populated using `renv_dependencies_impl()`, which doesn't take into account the snapshot type. Maybe we need to make a change on the `init()` side to use `renv_snapshot_dependencies()`?

I think that might be the right change since someone could theoretically do something like:

```
renv::init(settings = list(snapshot.type = "explicit"))
```

but then they'd end up with dependencies from all project files rather than just the DESCRIPTION.

R/activate.R
hadley: Would it be better to do `restart && !renv_tests_running()`?

R/dependencies.R
```diff
@@ -513,37 +513,48 @@ renv_dependencies_discover_description <- function(path,
...
       renv_dependencies_state(key = "root") %||%
       renv_restore_state(key = "root") %||%
       renv_project_resolve()
 
+    # get fields from settings
     fields <- settings$package.dependency.fields(project = project)
 
     # if this is the DESCRIPTION file for the active project, include
     # the dependencies for the active profile (if any) and Suggested fields.
+    # collect profile-specific dependencies as well
```
hadley, line 529: IMO this lines duplicates the info in the previous line

R/filebacked.R
```diff
@@ -98,3 +98,10 @@ filebacked <- function(scope, path, callback, ...) {
   result
 
 }
+
+renv_filebacked_invalidate <- function(path) {
+  renv_scope_options(warn = -1L)
+  eapply(the$filebacked_cache, function(scope) {
+    rm(list = path, envir = scope)
+  })
```
hadley, line 106:
```suggestion
  eapply(the$filebacked_cache, function(envir) {
    rm(list = path, envir = envir)
  })
```

Because it's an environment in an object, not an environment that refers to a frame on the call stack?

R/install.R
```diff
@@ -436,8 +442,8 @@ renv_install_package_cache <- function(record, cache, linker) {
   after <- Sys.time()
 
   type <- case(
-    identical(linker, renv_file_copy) ~ "copied from cache",
-    identical(linker, renv_file_link) ~ "linked from cache"
+    identical(linker, renv_file_copy) ~ "copied cache",
```
hadley, line 445: IMO it's worth keeping the from to more clearly distinguish to copying to the cache.

R/snapshot.R
hadley: That proposal makes sense to me. I think it's important to not recompute here so that the user is only questioned about missing dependencies once when `init`ing

tests/testthat/_snaps/snapshot.md
```diff
@@ -3,13 +3,6 @@
     Code
       snapshot()
     Output
-      The following package(s) are missing their DESCRIPTION files:
```
hadley, line 6: This is a regression (or at least requires some changes to the test), because the point of these tests (at least as I understand them) is to ensure various problems are reported to the user.

tests/testthat/_snaps/snapshot.md
```diff
@@ -186,7 +173,7 @@
       
       # Installing packages ---
       
-      - Installing egg ...                            OK [built from source and cached]
+      - Installing egg ...                            OK [built source and cached]
```
hadley, line 176: IMO removing the from makes this is a little less clear.

hadley (approved):
Looks good! I had noticed that `install()` was surprisingly slow in one of the tests and was going to see what was wrong.

R/filebacked.R
kevinushey: Good idea. I had reused the name 'scope' here because that's what it's called in `filebacked()`, e.g.

https://github.com/rstudio/renv/blob/35230cd61833a75461e72b06b17f7c01db123be0/R/filebacked.R#L82

but I think that should be called 'context' or 'id' or something separate.

R/activate.R
kevinushey: Makes sense; I made that change.

R/dependencies.R
kevinushey: You're right, I removed it.

R/dependencies.R
```diff
@@ -1646,23 +1657,6 @@ renv_dependencies_report <- function(errors) {
 
 }
 
-renv_dependencies_confirm <- function(action, path, ...) {
-
-  withCallingHandlers(
```
kevinushey, line 1651: This was inlined into `renv_snapshot_dependencies()`, and that's now used consistently in each function that needs to compute a project's package dependencies.

R/filebacked.R
```diff
@@ -33,20 +33,20 @@ renv_filebacked_set <- function(scope, path, value) {
...
-renv_filebacked_get <- function(scope, path) {
+renv_filebacked_get <- function(context, path) {
 
   # validate the path
   if (!renv_path_absolute(path))
     stopf("internal error: '%s' is not an absolute path", path)
 
-  # get scoped sub-environment
-  envir <- renv_filebacked_envir(scope)
+  # get contextd sub-environment
```
kevinushey, line 48:
```suggestion
  # get context sub-environment
```

R/hydrate.R
```diff
@@ -187,12 +187,7 @@ renv_hydrate_filter_impl <- function(package, path, library, update) {
 }
 
 renv_hydrate_packages <- function(project) {
-
-  deps <- the$init_dependencies %||%
-    renv_dependencies_impl(project, errors = "ignored", dev = TRUE)
-
-  unique(deps$Package)
-
+  renv_snapshot_dependencies(project, dev = TRUE)
```
kevinushey, line 190: We need to make sure `errors = "ignored"` is passed along here.

R/snapshot.R
```diff
@@ -948,39 +957,99 @@ renv_snapshot_report_actions <- function(actions, old, new) {
...
+      library = library,
+      paths   = filter(paths, startswith, prefix = library))
+  })
+
+  # now, snapshot the remaining packages
+  records <- map(paths, renv_snapshot_description)
+
 }
 
 renv_snapshot_filter <- function(project, records, type, packages, exclude) {
```
kevinushey, line 1055: Now that we're collecting only the records we know we require (respecting the snapshot type) I wonder if this method is still really needed?

R/snapshot.R
```diff
@@ -1085,59 +1098,6 @@ renv_snapshot_filter_report_missing <- function(missing, type) {
...
-
-  # report if dependency discovery took a long time
-  limit <- getOption("renv.snapshot.filter.timelimit", default = 10L)
-  diff <- difftime(end, start, units = "secs")
-
-  if (diff > limit) {
-
-    time <- difftime(end, start, units = "auto")
-    lines <- c(
-      "NOTE: Dependency discovery took %s during snapshot.",
```
kevinushey, line 1103: We'll probably want to revive this in some way.

tests/testthat/helper-scope.R
```diff
@@ -58,12 +58,6 @@ renv_tests_scope_repos <- function(scope = parent.frame()) {
 
 }
 
-renv_tests_scope_system_cache <- function(scope = parent.frame()) {
```
hadley, line 61: The point of this was to use the system cache for installing bioconductor packages to speed the tests up.

kevinushey: Yeah, but the Bioconductor tests are failing now and I don't understand why. 😞

PR #1488, opened 2023-06-19
Tweak messaging on load

hadley:


R/project.R
```diff
@@ -270,24 +270,24 @@ renv_project_synchronized_check <- function(project = NULL, lockfile = NULL) {
...
+      "- One or more packages recorded in the lockfile are not installed.",
+      "- Use `renv::status()` for more details."
     )
     writef(msg)
     return(FALSE)
   }
 
   # otherwise, use status to detect if we're synchronized
-  info <- quietly({
+  info <- local({
```
hadley, line 281: This is really the guts of the PR — previously this didn't create a new scope, so the message about the project being out of sync was never shown.

R/load.R
```diff
@@ -729,15 +730,13 @@ renv_load_quiet <- function() {
...
 
-  if (!renv_load_quiet()) {
-    renv_load_report_project(project)
-    renv_load_report_python(project)
-  }
+  renv_load_check(project)
+  renv_load_report_project(project)
+  renv_load_report_python(project)
 
   if (interactive()) {
```
hadley, line 741: Is it really the case we should only do this reports interactively? If so, should `renv_snapshot_auto_update()` also only be done interactively? The current behaviour seems inconsistent to me.

R/load.R
```diff
@@ -91,9 +91,10 @@ load <- function(project = NULL, quiet = FALSE) {
   if (switch)
     return(renv_load_switch(project))
 
-  if (quiet)
+  if (quiet || renv_session_quiet())
     renv_scope_options(renv.verbose = FALSE)
 
+  writef(header("Loading renv [%s]", renv_metadata_version_friendly()))
```
hadley, line 97: IMO a header makes sense here because it gets bullet updates.

R/load.R
```diff
@@ -757,11 +756,11 @@ renv_load_report_project <- function(project) {
   version <- renv_metadata_version_friendly()
 
   if (length(profile)) {
-    fmt <- "* (%s) Project '%s' loaded. [renv %s]"
-    writef(fmt, profile, renv_path_aliased(project), version)
+    fmt <- "- Project '%s' loaded with %s profile."
```
kevinushey, line 759: Do we want to make a decision around whether we use `*` prefixes or `-` prefixes for messages like these? Or maybe we should start wrapping these into internal calls to cli-like functions like `inform()` so we can easily adapt in the future?

kevinushey (approved):

R/load.R
kevinushey: The intention was for automatic snapshots to only happen in interactive sessions, so I'm happy to make things more consistent here.

R/load.R
hadley: I thought we were moving towards `-` everywhere?

kevinushey: Did we make that decision explicitly? I'm happy to do it but can't recall if we consciously chose to.

hadley: I don't think we discussed it but I think I assumed based on your recent PRs 😄

PR #1494, opened 2023-06-21
vendor tweaks

kevinushey:


R/vendor.R
```diff
@@ -110,20 +111,31 @@ renv_vendor_create <- function(project, sources, header) {
...
-renv_vendor_loader <- function(project, header) {
+renv_vendor_loader <- function(project, remote, header) {
 
   source <- system.file("resources/vendor/renv.R", package = "renv")
   template <- readLines(source, warn = FALSE)
 
   # replace '..imports..' with the imports we use
   imports <- renv_vendor_imports()
 
-  metadata <- as.list(the$metadata)
```
kevinushey, line 121: This didn't seem correct, since the version of `renv` being embedded might not actually match the currently-loaded version of `renv`.

PR #1497, opened 2023-06-21
Experiment with new `status()` display

hadley:


R/status.R
```diff
@@ -58,7 +111,7 @@ renv_status_impl <- function(project, libpaths, lockpath, sources, cache) {
 
   # check to see if we've initialized this project
   if (!renv_project_initialized(project)) {
-    writef("* This project has not yet been initialized.")
+    writef("This project has not yet been initialized.")
```
hadley, line 114: I wonder if these should be errors.

R/status.R
```diff
@@ -155,134 +214,45 @@ renv_status_check_unknown_sources <- function(project, lockfile) {
...
+    writef("The following package(s) are out of sync:")
+    writef()
+    print(problems, row.names = FALSE, right = FALSE)
   }
 
-  # */not recorded/not used ----------------------------------------------------
-  # No action; it's okay if some auxiliary packages are installed.
-
   # other changes, i.e. different version/source -------------------------------
   actions <- renv_lockfile_diff_packages(lockfile, library)
```
hadley, line 246: Probably worth pulling this out into a separate function.

NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `status()` gets new, more compact, display when packages have some 
+  inconsitent combination of being installed, used, and recorded.
```
kevinushey, line 5:
```suggestion
  inconsistent combination of being installed, used, and recorded.
```

R/status.R
```diff
@@ -3,8 +3,61 @@ the$status_running <- FALSE
 
 #' Report differences between lockfile and project library
 #'
-#' Report differences between the project's lockfile and the current state of
-#' the project's library (if any).
+#' `status()` reports on problems cauased by inconsistencies across the project
+#' lockfile, library, and [dependencies()]. In general, you should strive to
+#' ensure that `status()` reports no problems, as this maximises your chances
+#' of succesfully `restore()`ing the project in the future or on another
```
kevinushey, line 9:
```suggestion
#' of successfully `restore()`ing the project in the future or on another
```

R/status.R
```diff
@@ -3,8 +3,61 @@ the$status_running <- FALSE
...
+#' machine.  See the headings below for specific advice on resolving any
+#' problems revealed by `status()`.
+#'
+#' # Missing packages
+#'
+#' First, ensure that all packages used by the project are installed. This is
+#' important do to first because if any packages are missing we can’t tell for
```
kevinushey, line 16:
```suggestion
#' important to do first because if any packages are missing we can’t tell for
```

kevinushey:
```suggestion
#' important do to first because if any packages are missing we can't tell for
```

R/status.R
```diff
@@ -3,8 +3,61 @@ the$status_running <- FALSE
...
+#' sure that a package isn’t used; it might be a dependency that we don’t know
```
kevinushey, line 17:
```suggestion
#' sure that a package isn't used; it might be a dependency that we don’t know
```

R/status.R
kevinushey: I don't think so; IMHO `status()` shouldn't ever give an error, it should just give advice for getting the project into a "good" state.

R/status.R
```diff
@@ -74,18 +127,23 @@ renv_status_impl <- function(project, libpaths, lockpath, sources, cache) {
...
 
   # get library records
   library <- renv_snapshot_libpaths(libpaths = libpaths, project = project)
 
+  # get all dependencies, including transitive
+  dependencies <- renv_snapshot_dependencies(project, dev = FALSE)
+  packages <- sort(union(dependencies, "renv"))
+  paths <- renv_package_dependencies(packages, project = project)
+  packages <- as.character(names(paths))
+  # projects will implicitly depend on BiocManager & BiocVersion if any
```
kevinushey, line 141: This might no longer be necessary; we now implicitly add BiocManager and BiocVersion as dependencies to any Bioconductor packages we find.

kevinushey: https://github.com/rstudio/renv/blob/main/R/dependencies.R#L556-L562

R/status.R
```diff
@@ -112,7 +169,9 @@ renv_status_impl <- function(project, libpaths, lockpath, sources, cache) {
     renv_status_check_cache(project)
 
   if (synchronized)
-    writef("* The project is already synchronized with the lockfile.")
+    writef("No problems found.")
+  else
+    writef(c("", "See ?status() for advice on resolving the problems."))
```
kevinushey, line 174: For some reason I don't like the word 'problems' in this context: maybe instead we can use something like "reconciling this state" or something similar? That is, I don't view the lockfile + library being out of sync as a "problem" but just a thing that happens over the lifetime of a project that needs your decision-making.

R/status.R
hadley: My reasoning is that it might make sense for it to error when called on something that's not an renv project; you can obviously fix that by calling `renv::init()` but it might be more likely to reflect a genuine error? i.e. you're in the wrong directory.

kevinushey: Ah, that's a good point...

What if the user (1) hasn't loaded renv but (2) the working directory is in an renv project? (What if it's a sub-directory of an renv project?)

R/status.R
hadley: Yup, confirmed that it's not needed any more (the test we added for this case still passes).

R/status.R
hadley: What if we error if there's neither a project library nor a lockfile, but otherwise give a helpful message (i.e. call `snapshot()` if no lockfile; call `restore()` if no library)?

R/status.R
```diff
@@ -155,141 +209,55 @@ renv_status_check_unknown_sources <- function(project, lockfile) {
...
-        "Use `renv::snapshot()` if you'd like to remove these packages from the lockfile."
-    )
-
-    ok <- FALSE
+    issues <- status[!pkg_ok, , drop = FALSE]
+    issues$installed <- ifelse(issues$installed, "y", "n")
+    issues$recorded <- ifelse(issues$recorded, "y", "n")
+    issues$used <- ifelse(issues$used, "y", if (missing) "?" else "n")
 
+    writef("The following package(s) are inconsistent:")
```
kevinushey, line 238:
```suggestion
    writef("The following package(s) are in an inconsistent state:")
```

NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 # renv 0.18.0  (UNRELEASED)
 
+* `status()` gets new, more compact, display when packages have some
```
kevinushey, line 4:
```suggestion
* `renv::status()` gets new, more compact, display when packages have some 
```

I try to stick to using `renv::` as a prefix because we often encourage users to prefix usages of `renv` in a similar way; e.g. to avoid confusion about `base::load()` versus `renv::load()`.

R/status.R
```diff
@@ -1,10 +1,65 @@
...
+#' revealed by `status()`.
+#'
+#' # Missing packages
+#'
+#' First, ensure that all packages used by the project are installed. This is
+#' important to do first because if any packages are missing we can't tell for
+#' sure that a package isn't used; it might be a dependency that we don’t know
+#' about.
+#'
+#' There are four possibilities for an uninstalled package:
```
kevinushey, line 21: What should users do when multiple packages are in an inconsistent state? E.g. you have some packages which are used and recorded; and others which are recorded but not used.

(I think we want to recommend using `renv::restore()` first; perhaps `renv::restore(packages = <...>)`, and then using `renv::snapshot()` to remove unused packages?)

R/status.R
```diff
@@ -1,10 +1,65 @@
...
+#' or `renv::snapshot()`:
+#'
+#' * Call `renv::snapshot()` if your project code is working. This implies that
+#'   the library is correct and you need to update your lockfile.
+#' * Call `renv::restore()` if your project code isn't working. This probably
+#'   implies that you have the wrong package versions installed and you need
+#'   to restore from known good state in the lockfile.
+#'
+#' If you're not sure which case applies, it's generally safer to call
+#' `renv::snapshot()`
```
kevinushey, line 62: We might also recommend using `git` + `renv::history()` or `renv::revert()` for finding and using historical versions of lockfiles.

R/status.R
kevinushey: I think that makes sense.

kevinushey (approved):

R/status.R
kevinushey: Maybe pedantic but the package itself isn't inconsistent; it's the way it appears to be used / recorded in the project that is?

R/status.R
hadley: I've left these all as messages, not errors, but tweaked them all to be more immediately actionable.

kevinushey:
LGTM; I'll let you merge when ready!

Issue #1485, opened 2023-06-18
Should init() infer, or prompt, what the default snapshot type should be?


kevinushey:
If a user calls `renv::init()` on a project containing a DESCRIPTION file, should we:

- Assume they want to use "explicit" snapshots in the project, and set that project setting for them?
- Or, prompt them to do so?
- Or, do nothing?

hadley:
I like the idea of prompting, or at least messaging something like:

```
* You project has a `DESCRIPTION` but renv will capture all dependenices
* Use `settings$snapshot.type("explict")` to only capture dependencies defined in the `DESCRIPTION`
```

kevinushey:
Do we want different behavior for R packages versus "regular" projects that have a DESCRIPTION file?

PR #1504, opened 2023-06-26
prompt user on init if project has DESCRIPTION file

kevinushey:
Closes https://github.com/rstudio/renv/issues/1485.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,10 +1,13 @@
 
 # renv 0.18.0  (UNRELEASED)
 
-* `load()` makes the loading message a little more prominent, and we fixed a 
+* `renv::init()` now prompts the user to select a snapshot type, for projects
+  containing a top-level DESCRIPTION file. (#1485)
+
+* `renv::load()` makes the loading message a little more prominent, and we fixed a 
   bug that prevented out-of-sync repos from being reported.
 
-* `dependencies()` now discovers R dependencies inside jupyter notebooks (#929).
+* `renv::dependencies()` now discovers R dependencies inside jupyter notebooks (#929).
 
 * `renv::install(type = "source")` now ensures source repositories are used
   in projects using [PPM](https://packagemanager.posit.co/). (#927)
@@ -12,7 +15,7 @@
 * `renv::restore()` now emits an error if called in a project that
   does not contain a lockfile. (#1474)
 
-* If `config$user.profile()` is `TRUE`, the packages it uses are now included
+* If `renv::config$user.profile()` is `TRUE`, the packages it uses are now included
   in the development dependencies, which means that they will be installed 
   by `install()` but not recorded in the snapshot.
 
@@ -20,7 +23,7 @@
   appear to be installed from sources; that is, for packages without a
   recognizable remote source encoded in its DESCRIPTION file. (#841)
 
-* `load()` gives a more informative message if a lockfile is present but 
+* `renv::load()` gives a more informative message if a lockfile is present but 
   no packages are installed (#353).
 
 * `renv` now provides a small family of functions for interacting with `renv`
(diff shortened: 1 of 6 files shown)


kevinushey:
Here's an example of the output.

```
> renv::init()
This project contains a DESCRIPTION file.
DESCRIPTION files are often used to explicitly declare a project's R dependencies.
How would you like to declare R package dependencies in this project?

1: I will record my project's package dependencies in the DESCRIPTION file.     (explicit mode)
2: I will let renv automatically determine the project's package dependencies.  (implicit mode)

Selection: 1
- Using 'explicit' snapshot type. Please see `?renv::snapshot` for more details.
```

One other question to ask: should we perform this check if we're calling `renv::init()` in a project that has already been initialized? (In other words, if `settings$snapshot.type()` has already been set?)

R/init.R
```diff
@@ -307,3 +310,34 @@ renv_init_repos <- function() {
...
+
+    title = c(
+      "This project contains a DESCRIPTION file.",
+      "DESCRIPTION files are often used to explicitly declare a project's R dependencies.",
+      "How would you like to declare R package dependencies in this project?"
+    ),
+
+    choices = c(
+      explicit = "I will record my project's package dependencies in the DESCRIPTION file.     (explicit mode)",
+      implicit = "I will let renv automatically determine the project's package dependencies.  (implicit mode)"
```
hadley, line 332:
```suggestion
      explicit = "Take project dependencies only from the DESCRIPTION (explicit mode)",
      implicit = "Use all files to determine the dependencies (implicit mode)"
```

That's probably a bit too short, but in particularly the "I will" reads weirdly to me.

kevinushey: Thanks! This seems like an improvement to me.

The main thing I was trying to decide was whether we need to be clear that it's the user's responsibility to update the DESCRIPTION file, but maybe that's going to be obvious enough for anyone who has taken the time to actually create one?

hadley: Yeah, I think it's obvious enough from the context + message.

NEWS.md
```diff
@@ -1,26 +1,29 @@
...
   
 * `renv::restore()` now emits an error if called in a project that
   does not contain a lockfile. (#1474)
 
-* If `config$user.profile()` is `TRUE`, the packages it uses are now included
+* If `renv::config$user.profile()` is `TRUE`, the packages it uses are now included
   in the development dependencies, which means that they will be installed 
   by `install()` but not recorded in the snapshot.
 
 * `renv` now attempts to infer remote dependencies for GitHub packages that
```
hadley, line 22: If you're touching this, you could do `` `renv` `` to `renv` too

tests/testthat/_snaps/init.md
```diff
@@ -0,0 +1,30 @@
+# init() prompts the user for the snapshot type
+
+    Code
+      init()
+    Output
+      This project contains a DESCRIPTION file.
```
hadley, line 6: I really like this new wording 😄

tests/testthat/_snaps/init.md
```diff
@@ -0,0 +1,30 @@
...
+      Which files should renv use for dependency discovery in this project?
+      
+      1: Use only the DESCRIPTION file. (explicit mode)
+      2: Use all files in this project. (implicit mode)
```
hadley, line 10: The holy grail: perfect alignment without extra spaces 😆

tests/testthat/test-init.R
```diff
@@ -217,3 +217,13 @@ test_that("init() uses PPM by default", {
   expect_equal(repos, "https://packagemanager.posit.co/cran/latest")
 
 })
+
+test_that("init() prompts the user for the snapshot type", {
+  skip_on_cran()
+
+  project <- renv_tests_scope("bread")
+  writeLines("Depends: bread", con = "DESCRIPTION")
```
hadley, line 225: Might be worth also installing bread just to keep the snapshot focussed?

tests/testthat/_snaps/init.md
kevinushey: I tried very hard to make it happen 😄

PR #1509, opened 2023-06-28
replace default RStudio mirrors with PPM

kevinushey:
Since some R installations (from the default Linux OS package manager) might set a default R repository for the user.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -303,17 +303,21 @@ renv_init_repos <- function() {
   }
 
   # otherwise, check for some common 'default' CRAN settings
-  cran <- sub("/*$", "", repos[["CRAN"]] %||% "@CRAN@")
-  defaults <- c(
-    "@CRAN@",
-    "https://cloud.R-project.org",
-    "https://cran.rstudio.com",
-    "https://cran.rstudio.org"
-  )
+  cran <- repos[["CRAN"]]
+  if (is.character(cran) && length(cran) == 1L) {
+    cran <- sub("/*$", "", cran)
+    defaults <- c(
+      "@CRAN@",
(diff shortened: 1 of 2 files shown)


R/init.R
```diff
@@ -302,9 +302,19 @@ renv_init_repos <- function() {
     return(repos)
   }
 
-  # if no repository was set, use PPM
-  if (identical(repos, list(CRAN = "@CRAN@")))
-    return(config$ppm.url())
+  # otherwise, check for some common 'default' CRAN settings
+  cran <- sub("/*$", "", repos[["CRAN"]] %||% "@CRAN@")
```
kevinushey, line 306: Or should we only do this step if `CRAN` was already set?

R/init.R
```diff
@@ -302,9 +302,23 @@ renv_init_repos <- function() {
...
+      "@CRAN@",
+      "https://cloud.R-project.org",
+      "https://cran.rstudio.com",
+      "https://cran.rstudio.org"
+    )
+
+    if (tolower(cran) %in% tolower(defaults)) {
+      repos[["CRAN"]] <- config$ppm.url()
+      return(repos)
+    }
```
hadley, line 319:
```suggestion
  std <- sub("/*$", "", repos)
  defaults <- c(
    "@CRAN@",
    "https://cloud.R-project.org",
    "https://cran.rstudio.com",
    "https://cran.rstudio.org"
  )

  repos[tolower(cran) %in% tolower(std)] <- config$ppm.url()
```

Not sure if this is any better, but maybe it's a bit easier to reason about?

hadley (approved):

PR #1512, opened 2023-06-28
tweak reporting of auto snapshot errors

kevinushey:
Closes https://github.com/rstudio/renv/issues/1508.

### DIFF
--- a/R/snapshot-auto.R
+++ b/R/snapshot-auto.R
@@ -1,6 +1,7 @@
 
 the$library_info <- NULL
 
+the$snapshot_failed <- FALSE
 the$snapshot_running <- FALSE
 the$snapshot_suppressed <- FALSE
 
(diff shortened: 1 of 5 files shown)


R/snapshot-auto.R
```diff
@@ -113,6 +114,21 @@ renv_snapshot_auto_update <- function(project = renv_project_get() ) {
 
 renv_snapshot_task <- function() {
 
+  # treat warnings as errors in this scope
+  renv_scope_options(warn = 2L)
+
+  # attempt automatic snapshot, but disable on failure
+  withCallingHandlers(
+    renv_snapshot_task_impl(),
+    error = function(cnd) {
```
hadley, line 123:
```suggestion
  # attempt automatic snapshot, but disable on failure
  tryCatch(
    renv_snapshot_task_impl(),
    warning = function(cnd) {
```

?

Probably still also want to capture errors

R/snapshot-auto.R
```diff
@@ -1,6 +1,7 @@
 
 the$library_info <- NULL
 
+the$snapshot_failed <- FALSE
```
hadley, line 4: Am I failing to see where this is actually used?

hadley (approved):
This seems a bit fiddly to get right, so maybe worth a test? There's a model at https://github.com/rstudio/renv/blob/main/tests/testthat/test-snapshot.R#L470-L487

R/snapshot-auto.R
kevinushey: You aren't 🙈  I'll fix this.

R/snapshot-auto.R
kevinushey: I set `options(warn = 2L)` above so that warnings are converted into errors (since I figure I want to treat both warnings and errors the same way here)

hadley: Yeah, I just think you're better of doing that directly, rather than relying on R to convert warnings to errors (which travels through a fairly complex code path).

Issue #1511, opened 2023-06-28
Error in if (startswith(loaded, version)) { : argument is of length zero


kevinushey:
This can happen if the activate script has a recorded RemoteSha, but the currently-installed version of `renv` lacks it.

kevinushey:
If you re-run `renv::activate()` in the affected project, does the issue go away?

kevinushey:
I guess this can happen if `renv` is updated to 1.0.0 without the activate script being updated. :-/

LDalby:
I just ran into the same issue as @dchiu911, but I was able to fix it via `renv::activate()` as suggested by @kevinushey 
I was upgrading a project from renv version 0.16.0 to 1.0.0

kevinushey:
I think this should be fixed via https://github.com/rstudio/renv/pull/1576, so at least we won't see this issue in future `renv` releases.

PR #1513, opened 2023-06-28
tweak display of bootstrap versions

kevinushey:
Closes https://github.com/rstudio/renv/issues/1511.

I'm starting to think we really should be recording more information into the `activate.R` script, though... but this is probably good enough until we eventually implement https://github.com/rstudio/renv/issues/890.


kevinushey:
I went kind of overboard and tried to resolve the conflation between version + sha in the version argument. Now:

- version always refers to renv's package version;
- it will have a "sha" attribute set if it's a development version

It's a bit wonky but I think helps in the end -- now, we know the `renv` version associated with a particular commit hash when loading `renv` as well.

R/activate.R
```diff
@@ -129,17 +139,15 @@ renv_activate_version_lockfile <- function(project) {
...
-    lockfile$Packages[["renv"]]$Version %||%
-    lockfile[["renv"]]$Version
+  records <- renv_lockfile_records(lockfile)
+  renv_metadata_version_create(records[["renv"]])
 
 }
 
-# TODO: can we unravel this, and just record the metadata structure
-# into the activate script?
 renv_activate_version_default <- function(project) {
```
hadley, line 149: Maybe rename to `renv_activate_version_metadata()`?

R/bootstrap.R
```diff
@@ -145,20 +141,19 @@ renv_bootstrap_repos_lockfile <- function() {
...
+    function() renv_bootstrap_download_tarball(version),
+    if (!is.null(sha))
+      function() renv_bootstrap_download_github(sha)
     else c(
-      renv_bootstrap_download_cran_latest,
-      renv_bootstrap_download_cran_archive
+      function() renv_bootstrap_download_cran_latest(version),
+      function() renv_bootstrap_download_cran_archive(version)
     )
   )
```
hadley, line 153: I wonder if it's better to move the `if` up a level:

```R
methods <- if (is.null(sha)) {
  c(
   function() renv_bootstrap_download_tarball(version),
   function() renv_bootstrap_download_github(sha)
  )
} else {
  c(
    function() renv_bootstrap_download_tarball(version),
    function() renv_bootstrap_download_cran_latest(version),
    function() renv_bootstrap_download_cran_archive(version)
  )
}
```

Or possibly even check `renv_envvar_exists("RENV_BOOTSTRAP_TARBALL")` first, because I feel like if that's set and the file doesn't exist, you probably don't want to fall back to the default.

R/infrastructure.R
```diff
@@ -108,7 +109,14 @@ renv_infrastructure_write_activate <- function(project = NULL,
     return(FALSE)
 
   template <- renv_file_read(source)
-  new <- renv_template_replace(template, list(VERSION = version))
+  new <- renv_template_replace(
+    text = template,
+    replacements = list(
+      version = stringify(c(version)),
+      sha = stringify(c(sha))
```
hadley, line 116: Why `c()`?

hadley (approved):

R/infrastructure.R
hadley: Maybe `as.character()` would be clearer?

kevinushey: Good idea. Even the documentation agrees with you. `?c` has:

>      'c' is sometimes used for its side effect of removing attributes
>      except names, for example to turn an 'array' into a vector.
>      'as.vector' is a more intuitive way to do this, but also drops
>      names.

Issue #961, opened 2022-03-15
renv_python_discover should return python earliest in PATH, not latests


timbalam:
Hi devs,

The issue I have encountered occurs when first starting an renv project that uses a python virtualenv, in a system with two installations of the same version of python available in PATH. When running for the first time running, renv attempts to create a virtualenv, but finds an unexpected python executable. I expect that the python executable in a folder found earliest on PATH (left-most) would be used, however the python executable found latest on PATH is found.

I believe the cause of this behaviour is in the code for discovering paths for installed python binaries renv_python_discover. In the function, the folders on PATH are traversed (from latest to earliest/rightmost to leftmost) and any found binaries are pushed to a 'stack'. Finally, the $data() method is called on the stack to get the list of pushed binaries. This method returns a list of pushed elements in the order they were pushed, meaning that the python executables that occur in folders earlier on PATH are at the front of the list. The function then returns the first executable with the appropriate version from the list.
 
 To fix this behaviour, some possible options are:
 - don't reverse the order of the folders on PATH in the loop where the found executables are pushed
 - change the behaviour of the $data() method of the stack to reverse the internal list so that elements pushed last are at the front of the list
 - reverse the output of the $data() method
 
 Sorry for not providing a reproducible example! 
 
 Thanks, 
 Tim

kevinushey:
Thanks for the bug report + in-depth investigation! I agree with this solution:

> don't reverse the order of the folders on PATH in the loop where the found executables are pushed

I'll push a change to that effect. Thanks!

kevinushey:
Should now be fixed via https://github.com/rstudio/renv/commit/2538da36d9dfa88fc6e1cf88005a31067cb5bd82.

PR #1529, opened 2023-07-08
Discover python with reticulate, fix PATH order

mxposed:
I encountered 2 problems with `renv_python_discover` function:

1. It ignores `RETICULATE_PYTHON` environment variable, which is confusing, because after virtualenv is setup, this variable controls the python executable
2. It goes over `PATH` components in forward order, but puts pythons discovered from there into a stack, so the actual python picked from `PATH` is from the last, not the first encountered directory.

This pull request corrects both things. I did not test this

### DIFF
--- a/R/python.R
+++ b/R/python.R
@@ -278,6 +278,10 @@ renv_python_discover <- function() {
     "~/opt/python",
     file.path(renv_pyenv_root(), "versions")
   )
+  reticulate <- Sys.getenv("RETICULATE_PYTHON")
+  if (length(reticulate)) {
+    roots <- c(roots, reticulate)
+  }
 
   for (root in roots) {
     versions <- sort(list.files(root, full.names = TRUE), decreasing = TRUE)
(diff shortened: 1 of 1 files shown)


R/python.R
```diff
@@ -278,6 +278,10 @@ renv_python_discover <- function() {
     "~/opt/python",
     file.path(renv_pyenv_root(), "versions")
   )
+  reticulate <- Sys.getenv("RETICULATE_PYTHON")
+  if (length(reticulate)) {
```
kevinushey, line 282:
```suggestion
  if (nzchar(reticulate)) {
```

Since `Sys.getenv()` will return a zero-length string if the associated environment variable doesn't exist.

mxposed:
Apologies, I have found the issue https://github.com/rstudio/renv/issues/961 and that it's been fixed (I happened to have renv 0.15.5).

I'll close this now, but please let me know if you think using `RETICULATE_PYTHON` in this context makes sense (it does for me), I'll create a separate pull request for that.

Issue #1530, opened 2023-07-10
Possible bug: renv does not respect renv.lock when activating


MichalLauer:
Hello,
I have noticed that when creating a new project with only renv.lock file, the packages are not installed even tho `{renv}` says that the library is synchronized. Steps for reproduction:

1) Create a blank project with renv.lock file
```
ProjectFolder
│
└─project.Rproj
└─renv.lock
```

2) Run `renv::activate()`

The expected behavior would be installing all packages from renv.lock, but this does not happen, and the library in *renv/library/R-4.3/x86_64-w64-mingw32* contains only renv folder. Here is a screenshot from R console.

![image](https://github.com/rstudio/renv/assets/20439350/96922629-11d9-4664-9abd-2f935efb7ed2)

After running `renv::status()`, I can see that renv thinks that all packages are installed, but it is not true.

![image](https://github.com/rstudio/renv/assets/20439350/5d2fca6b-ab96-462e-b65f-82ac685e5bc4)
(note they are not in use because I am using a new clean project, not my main project)

I also tried running `renv::restore()` or `renv::install()` but no luck either

![image](https://github.com/rstudio/renv/assets/20439350/dc22690b-2533-4a31-a384-4414f0005891)

<details>

<summary>Here is my current (also long) renv.lock file</summary>
```
{
  "R": {
    "Version": "4.3.0",
    "Repositories": [
      {
        "Name": "CRAN",
        "URL": "https://cran.rstudio.com"
      }
    ]
  },
  "Packages": {
    "MASS": {
      "Package": "MASS",
      "Version": "7.3-60",
      "Source": "Repository",
      "Repository": "CRAN",
      "Requirements": [
        "R",
        "grDevices",
        "graphics",
        "methods",
        "stats",
        "utils"
      ],
      "Hash": "a56a6365b3fa73293ea8d084be0d9bb0"
    },
    "Matrix": {
      "Package": "Matrix",
      "Version": "1.5-4.1",
      "Source": "Repository",
      "Repository": "CRAN",
      "Requirements": [
        "R",
        "graphics",
        "grid",
        "lattice",
        "methods",
        "stats",
        "utils"
      ],
      "Hash": "38082d362d317745fb932e13956dccbb"
    },
    "R6": {
      "Package": "R6",
      "Version": "2.5.1",
      "Source": "Repository",
      "Repository": "CRAN",
      "Requirements": [
        "R"
      ],
[...]

PR #1534, opened 2023-07-10
perform 'full' session restart after activate

kevinushey:
Closes https://github.com/rstudio/renv/issues/1530.

### DIFF
--- a/R/restart.R
+++ b/R/restart.R
@@ -59,10 +59,10 @@ renv_restart_request_rstudio <- function(project, reason, ...) {
     restart <- getOption("renv.restart.function", default = function() {
       .rs.api.executeCommand("restartR", quiet = TRUE)
     })
-    return(restart())
+    return(invisible(restart()))
   }
 
   # otherwise, explicitly open the new project
-  tools$.rs.api.openProject(project, newSession = FALSE)
+  invisible(tools$.rs.api.openProject(project, newSession = FALSE))
 
 }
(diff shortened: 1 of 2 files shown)


R/restart.R
```diff
@@ -49,12 +49,17 @@ renv_restart_request_rstudio <- function(project, reason, ...) {
...
+  # the 'restart' option here as the version RStudio uses
+  # tries to preserve session state that we need to change.
+  #
+  # https://github.com/rstudio/renv/issues/1530
   projdir <- tools$.rs.getProjectDirectory() %||% ""
   if (renv_file_same(projdir, project)) {
-    restart <- getOption("restart")
-    if (is.function(restart))
-      return(restart())
+    restart <- getOption("renv.restart.function", default = function() {
```
hadley, line 59: Is this a new option?

kevinushey: Yeah. It's really just an escape hatch in case someone needs it for some reason.

R/restart.R
```diff
@@ -49,12 +49,17 @@ renv_restart_request_rstudio <- function(project, reason, ...) {
...
+      .rs.api.executeCommand("restartR", quiet = TRUE)
+    })
+    return(restart())
```
hadley, line 62: I wonder if this function should return invisibly because it's called primarily for its side effects.

hadley (approved):

R/restart.R
kevinushey: It should; thanks for catching that!

Issue #1536, opened 2023-07-11
[renv@1.0.0]: prompted for action after project activation in RStudio


Gilquin:
After activating a project within RStudio and executing a single line in the console, `renv` always prompts the following:
```r
What do you want to do?

1: Snapshot, just using the currently installed packages.
2: Install the packages, then snapshot.
3: Cancel, and resolve the situation on your own.
 
Selection: 
```
This doesn't occur when the project is activated within an R session launched from the terminal.

```bash
R version 4.3.0 (2023-04-21)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 22.04.2 LTS
RStudio: 2023.06.0+421
```

kevinushey:
It looks like this happens when an automatic snapshot is triggered when a package recorded in the lockfile isn't currently installed. As a workaround, you can disable automatic snapshots:

```
RENV_CONFIG_AUTO_SNAPSHOT = FALSE
```

in your `~/.Renviron`.

PR #1539, opened 2023-07-11
allow cancel of auto snapshot

kevinushey:
Closes https://github.com/rstudio/renv/issues/1536. In essence, if an automatic snapshot would have prompted the user for installation of missing packages, we just cancel the automatic snapshot.

This might warrant some more consideration of when it's safe to perform an automatic snapshot.

### DIFF
--- a/R/snapshot-auto.R
+++ b/R/snapshot-auto.R
@@ -25,7 +25,7 @@ renv_snapshot_auto <- function(project) {
       error = function(err) FALSE
     ),
 
-    cancel = function() { FALSE }
+    cancel = function() FALSE
 
   )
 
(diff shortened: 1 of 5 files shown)


R/snapshot-auto.R
```diff
@@ -1,23 +1,35 @@
...
-    renv_snapshot_auto_impl(project),
-    error = function(err) FALSE
+  updated <- withCallingHandlers(
+
+    tryCatch(
+      renv_snapshot_auto_impl(project),
+      error = function(err) FALSE
+    ),
+
+    cancel = function() { FALSE }
```
hadley, line 28:
```suggestion
    cancel = function() FALSE
```

hadley (approved):

PR #1545, opened 2023-07-12
include system library in restore

### LINKED ISSUE #1544: [renv 1.0.0] restore(prompt = FALSE) does not respect prompt parameter in profile with default snapshot settings
I've found an issue with our [hydrate -> restore -> snapshot workflow](https://carpentries.github.io/sandpaper/articles/building-with-renv.html) where `renv::restore(prompt = FALSE)` will fail in a profile if it's run non-interactively (see https://github.com/carpentries/sandpaper/issues/490), but I've also found that it will still prompt if it's run interactively. 

I've created [a minimum working example](https://github.com/zkamvar/sandpaper-issue-490#readme) (less complex, but derived from the WE for #1177) that demonstrates this workflow and shows that, while new packages are provisioned for the restore step, it never moves on to the snapshot step, so new packages are not recorded in the lockfile: 

<details><summary>Non-interactive results</summary>

Excerpt from [the output of the MWE](https://github.com/zkamvar/sandpaper-issue-490/blob/e6dc4bc4eaa0c03fbed5f7358b49f5ed215a1e01/out.txt):

```
The following packages are used in this project, but not available locally:

- reprex

renv will attempt to download and install these packages.

- Hydrated 0 packages in 0.028 milliseconds.
- Resolving missing dependencies ... 

- Installing ps ...                             OK [linked from cache in 0.00038s]
- Installing processx ...                       OK [linked from cache in 0.00027s]
- Installing callr ...                          OK [linked from cache in 0.00027s]
- Installing clipr ...                          OK [linked from cache in 0.00025s]
- Installing rstudioapi ...                     OK [linked from cache in 0.00027s]
- Installing withr ...                          OK [linked from cache in 0.00025s]
- Installing reprex ...                         OK [linked from cache in 0.00032s]

The following required packages are not installed:

- boot
- class
- cluster
- codetools
- foreign
- KernSmooth
- lattice
- MASS
- Matrix
- mgcv
- nlme
- nnet
- rpart
- spatial
- survival

Packages must first be installed before renv can snapshot them.
Use `renv::dependencies()` to see where this package is used in your project.

Not interactive. Will:
Snapshot, just using the currently installed packages.
- Operation canceled.
```

</details> 

In addition, when I run this interactively, I find the prompt appearing:

```r
r$> renv::restore(library = renv_lib, lockfile = renv_lock, prompt = FALSE)
The following required packages are not installed:

- boot
- class
- cluster
- codetools
- foreign
- KernSmooth
- lattice
- MASS
- Matrix
- mgcv
- nlme
- nnet
- rpart
- spatial
- survival

Packages must first be installed before renv can snapshot them.
Use `renv::dependencies()` to see where this package is used in your project.

What do you want to do?

1: Snapshot, just using the currently installed packages.
2: Install the packages, then snapshot.
3: Cancel, and resolve the situation on your own.

Selection:
```

kevinushey:
Closes #1544.

@zkamvar, are you able to test if this PR does indeed resolve your issue?

### DIFF
--- a/R/lockfile.R
+++ b/R/lockfile.R
@@ -212,7 +212,10 @@ renv_lockfile_create_impl <- function(project, type, libpaths, packages, exclude
   # warn if some required packages are missing
   ignored <- c(renv_project_ignored_packages(project), renv_packages_base(), exclude)
   missing <- setdiff(packages, c(names(records), ignored))
-  if (!the$status_running && !the$init_running)
+
+  # TODO: we likely need a better way to distinguish between top-level snapshot
+  # calls, versus snapshot calls that renv is using internally.
+  if (!the$status_running && !the$init_running && !the$restore_running)
     renv_snapshot_report_missing(missing, type)
 
   records <- renv_snapshot_fixup(records)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 
 # renv 1.1.0  (UNRELEASED)
 
+* `renv::restore()` no longer erroneously prompts when determining the packages
+  which need to be installed. (#1544)
+
 * `renv::update()` now ensures the `prompt` parameter is properly respected
   during package installation. (#1540)
 
@@ -8,6 +11,7 @@
   paths after restarting the session within RStudio. (#1530)
 
 * Use correct spelling of IRkernel package (#1528).
+
 
 # renv 1.0.0
 
--- a/R/activate.R
+++ b/R/activate.R
@@ -180,7 +180,7 @@ renv_activate_prompt <- function(action, library, prompt, project) {
 renv_activate_prompt_impl <- function(action, project = NULL) {
   title <- c(
     sprintf(
-      "It looks like you've called renv::%s() in a project that hasn't been activated yet",
+      "It looks like you've called renv::%s() in a project that hasn't been activated yet.",
       action
     ),
     "How would you like to proceed?"
--- a/R/libpaths.R
+++ b/R/libpaths.R
@@ -186,6 +186,16 @@ renv_libpaths_restore <- function() {
   renv_libpaths_set(libpaths)
 }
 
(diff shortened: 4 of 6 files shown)


zkamvar:
It looks like it fixes that particular issue, [the `renv.lock` now contains the {reprex} package as promised](https://github.com/zkamvar/sandpaper-issue-490/commit/bb7094fa221cec5b201f4d488cf191436f6b3a2e#diff-85932349faa3693a873c64d3f2d30b2bdae4e113c3026961db68ffeb34e03a5d) meaning that `renv::restore()` did not fail!

R/lockfile.R
```diff
@@ -212,7 +212,10 @@ renv_lockfile_create_impl <- function(project, type, libpaths, packages, exclude
   # warn if some required packages are missing
   ignored <- c(renv_project_ignored_packages(project), renv_packages_base(), exclude)
   missing <- setdiff(packages, c(names(records), ignored))
-  if (!the$status_running && !the$init_running)
+
+  # TODO: we likely need a better way to distinguish between top-level snapshot
+  # calls, versus snapshot calls that renv is using internally.
+  if (!the$status_running && !the$init_running && !the$restore_running)
     renv_snapshot_report_missing(missing, type)
```
zkamvar, line 219: I'm curious about this, is this what causes snapshot to require input?

kevinushey: That's right. But that prompt should only really be presented to the user when performing a top-level snapshot; that is, the user ran `renv::snapshot()` themselves. I think we need a cleaner way of handling this, but at least this is a good enough bandaid for now.

zkamvar: Thank you so much for the explanation and your quick response on these issues! I really appreciate the work you do

kevinushey: I'm happy I can help; sorry for the trouble in the interim!

Issue #1336, opened 2023-05-25
renv::install with explicit dependencies = Imports still installs Suggests 


Fuco1:
If I'm trying to install "active project", suggested packages get installed even when I explicitly only request `Imports` to be installed.  The following code seems to be responsible:

```R
  # if this is the DESCRIPTION file for the active project, include
  # Suggests since they're often needed as well. such packages will be
  # considered development dependencies, except for package projects
  type <- "unknown"
  project <-
    project %||%
    renv_dependencies_state(key = "root") %||%
    renv_project_resolve()

  if (renv_path_same(file.path(project, "DESCRIPTION"), path)) {

    # collect profile-specific dependencies as well
    profile <- renv_profile_get()
    field <- if (length(profile))
      sprintf("Config/renv/profiles/%s/dependencies", profile)

    fields <- c(fields, "Suggests", field)
    type <- renv_description_type(desc = dcf)

  }
```

Basically, now if I want to really only install `Imports` packages, I have to use some renv internals to generate a list of the dependencies and then filter out the "dev" dependencies (also I have to set `Type` in description to some bogus value so the type is not "package").

This behaviour is very unintuitive and is not documented anywhere in `install`.  I think when user explicitly passes the dependencies parameter, nothing should happen to it and it should be taken as-is.

Fuco1:
Sorry, I should've been more precise.  In a `renv` project, call `renv::install("PATH_TO_PROJECT", dependencies = "Imports")`.  I'm not sure if path needs to be absolute or you can pass `"."` for current directory.  I always use absolute path.

I don't know if this is normal, but I'm trying to install the project "to itself" as a dependency so that my analysis script can require itself (instead of using `devtools::load_all()`. (the `R/` contains all the logic and then I have a top-level file `run.R` which loads the package with `library` and does the analysis)

hadley:
And presumably you have a `DESCRIPTION` file?

hadley:
If so, this is a duplicate of #1019

kevinushey:
I think this issue might be slightly different -- it sounds like the `dependencies` argument is being ignored in this scenario? The other issue relates to what dependencies are installed by default?

hadley:
@kevinushey good catch, I'll add a note to that issue.

Fuco1:
@kevinushey The argument is not ignored, but `Suggests` is appended to it as you see on this line `fields <- c(fields, "Suggests", field)`.

[...]

PR #1556, opened 2023-07-14
tweak handling of dependencies

kevinushey:
Closes https://github.com/rstudio/renv/issues/1336.

It seems like 'dependencies' can be used for two different meanings:

1. Controlling which subset of packages get installed from the package DESCRIPTION file;
2. Controlling which (custom) fields are used when installing packages and their dependencies.

For (1), if you specify `renv::install(dependencies = "Imports")`, I think we should only take the Imports of the top-level DESCRIPTION file, but still use `c("Depends", "Imports", "LinkingTo")` for the dependencies of those packages.

For (2), I think we should propagate the "custom" dependency fields. So, for example, if you use:

```
renv::install("toast", dependencies = "Config/Needs/protein")
```

And we have this dependency graph:

```
toast -> bread -> [Config/Needs/protein: egg]
```

then we should install `egg`. Is that reasonable? Or should the `dependencies` argument only apply when scanning the top-level DESCRIPTION file?

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -540,24 +540,34 @@ renv_dependencies_discover_description <- function(path,
       renv_restore_state(key = "root") %||%
       renv_project_resolve()
 
-    # get fields from settings
+    # if this appears to be the DESCRIPTION associated with the active project,
+    # and an explicit set of dependencies was provided in install, then use those
+    primary <- renv_path_same(file.path(project, "DESCRIPTION"), path)
+
+    # by default, respect fields defined in settings
     fields <- settings$package.dependency.fields(project = project)
 
-    # if this is the DESCRIPTION file for the active project, include
-    # the dependencies for the active profile (if any) and Suggested fields.
-    # collect profile-specific dependencies as well
-    if (renv_path_same(file.path(project, "DESCRIPTION"), path)) {
-      fmt <- "Config/renv/profiles/%s/dependencies"
-      profile <- renv_profile_get()
-      fields <- c(fields, "Suggests", sprintf(fmt, profile))
+    # check for overrides if this is the project's own DESCRIPTION file
+    if (primary) {
+      default <- the$install_dependency_fields %||% c(fields, "Suggests")
+      profile <- sprintf("Config/renv/profiles/%s/dependencies", renv_profile_get())
+      fields <- c(default, profile)
     }
 
+    # return computed fields
     fields
 
   }
 
   # make sure dependency fields are expanded
   fields <- renv_description_dependency_fields_expand(fields)
+
+  # propagate 'extra' dependency fields if requested in install
+  if (!is.null(the$install_dependency_fields)) {
+    default <- c("Depends", "Imports", "Suggests", "LinkingTo")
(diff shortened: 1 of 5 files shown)


R/dependencies.R
```diff
@@ -540,25 +540,35 @@ renv_dependencies_discover_description <- function(path,
       renv_restore_state(key = "root") %||%
       renv_project_resolve()
 
-    # get fields from settings
+    # if this appears to be the DESCRIPTION associated with the active project,
```
hadley, line 543: IMO it's time to refactor this big `{}` block because it's getting hard to see the overall flow of this function. Maybe we need `renv_dependencies_discover_description_fields()`?

tests/testthat/test-install.R
```diff
@@ -597,3 +609,21 @@ test_that("install() reports failure when a 'bad' binary is installed", {
...
+    Type: Project
+    Depends: coffee
+    Imports: bread
+    Suggests: muffin
+  ")
+
+  writeLines(contents, con = "DESCRIPTION")
+  install(dependencies = "Imports")
+
+  expect_true(renv_package_installed("bread"))
```
hadley, line 628:
```suggestion
  expect_true(renv_package_installed("bread"))
  expect_false(renv_package_installed("coffee"))
  expect_false(renv_package_installed("muffin"))
```

R/dependencies.R
```diff
@@ -540,25 +540,35 @@ renv_dependencies_discover_description <- function(path,
...
+    # return computed fields
     fields
 
   }
 
   # make sure dependency fields are expanded
   fields <- renv_description_dependency_fields_expand(fields)
 
+  # propagate 'extra' dependency fields if requested in install
+  if (!is.null(the$install_dependency_fields)) {
```
hadley, line 566: This only affects non-primary description fields, right?

Also the `dependencies` argument of `install` is documented to accept "strong" etc so I think you need another `renv_description_dependency_fields_expand(fields)` somewhere.

hadley (approved):

R/dependencies.R
kevinushey: Yeah, exactly. The idea being if you wanted to install all `Config/Needs/blah` packages, we'd want to install all discovered dependencies even from recursive dependencies.

kevinushey: Hmm, possibly... part of the reason I had it outside that block was because of this:

https://github.com/rstudio/renv/blob/75c75dc51071a1a916ace5ec6f00ccfac6cd4b51/R/retrieve.R#L1033

In effect, I wanted to make sure that the 'extra' fields were passed along even if the caller supplied `fields = <...>` in some call to `renv_dependencies_discover_description()`.

kevinushey: Also here: https://github.com/rstudio/renv/blob/75c75dc51071a1a916ace5ec6f00ccfac6cd4b51/R/package.R#L288-L292

hadley: Overall, my feeling is that it's probably worth refactoring this function (and testing just the fields outputs), but I don't think that needs to happen in this PR.

Issue #1557, opened 2023-07-16
Provide option for profiling renv startup


kevinushey:
On occasion, R can be slow to start up due to work that `renv` is doing. It would be helpful if we could define something like:

```
RENV_STARTUP_DIAGNOSTICS = TRUE
```

and use `Rprof()` to provide a summary on what `renv` was doing while the project `.Rprofile` was sourced.

black-snow:
Would be nice. I receive
> Warning message:
renv took longer than expected (20 seconds) to activate the sandbox.

with 1.0.0 constantly and it's unclear why :)

PR #1566, opened 2023-07-19
add option for startup diagnostics

kevinushey:
Closes https://github.com/rstudio/renv/issues/1557.

### DIFF
--- a/inst/resources/activate.R
+++ b/inst/resources/activate.R
@@ -8,6 +8,21 @@ local({
   # the project directory
   project <- getwd()
 
+  # use start-up diagnostics if enabled
+  diagnostics <- Sys.getenv("RENV_STARTUP_DIAGNOSTICS", unset = "FALSE")
+  if (diagnostics) {
+    start <- Sys.time()
+    profile <- tempfile("renv-startup-", fileext = ".Rprof")
+    utils::Rprof(profile)
+    on.exit({
+      utils::Rprof(NULL)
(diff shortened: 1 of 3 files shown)


inst/resources/activate.R
```diff
@@ -8,6 +8,21 @@ local({
...
+  if (diagnostics) {
+    start <- Sys.time()
+    profile <- tempfile("renv-startup-", fileext = ".Rprof")
+    utils::Rprof(profile)
+    on.exit({
+      utils::Rprof(NULL)
+      elapsed <- signif(difftime(Sys.time(), start, units = "auto"), digits = 2L)
+      writeLines(sprintf("- renv took %s to run the autoloader.", format(elapsed)))
+      writeLines(sprintf("- Profile: %s", profile))
+      print(utils::summaryRprof(profile))
```
hadley, line 22:
```suggestion
      print(utils::summaryRprof(profile))
      unlink(profile)
```
?

And maybe also `the$startup_profile <- utils::summaryRprof(profile)` so that if needed we can reach inside and interactively inspect the profile?

hadley (approved):

inst/resources/activate.R
kevinushey: I wanted to keep the profile around on disk just in case it's worth sharing the raw data in some scenarios. Since it lives in the temporary directory, it'll be cleaned up when the user quits anyhow.

Issue #1565, opened 2023-07-19
automatically activate sandbox if a sandbox library is on libpaths


kevinushey:
This allows `renv` to do the right thing when used in an `R --vanilla` session where the sandbox library was propagated. See https://github.com/rstudio/rsconnect/issues/918#issuecomment-1642605748 for motivation.

PR #1567, opened 2023-07-19
automatically activate sandbox on load if appropriate

kevinushey:
Closes https://github.com/rstudio/renv/issues/1565.


aronatkins (approved):
Confirmed that I am able to deploy from an renv project with this adjustment.

Issue #1559, opened 2023-07-17
installed-from-source warning for non-project package


aronatkins:
Using renv-1.0.0 from CRAN and an installation from source of rsconnect into the default R library:

``` r
td <- tempfile()
dir.create(td)
fn <- file.path(td, "index.Rmd")
writeLines(c("hello, markdown"), fn)
renv:::renv_snapshot_preflight_check_sources(td, .libPaths()[[1L]])
#> The following package(s) were installed from sources, but may be available from the following remotes:
#> 
#> - rsconnect  [rstudio/rsconnect]
#> 
#> Would you like to use these remote sources for these packages?
#> See 'snapshot.inference' in ?renv::config for more details.
#> Otherwise, these packages will be recorded with an unknown source.
#> [1] TRUE
```

<sup>Created on 2023-07-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

aronatkins:
This does not happen when rsconnect is installed from GitHub using remotes; it only happens when the package is installed from source. All from-source packages are reported as having problems, even when they are unrelated to the project.

aronatkins:
When renv and rsconnect are both installed from source (to show this is not a problem only with rsconnect):

``` r
td <- tempfile()
dir.create(td)
fn <- file.path(td, "index.Rmd")
writeLines(c("hello, markdown"), fn)
renv:::renv_snapshot_preflight_check_sources(td, .libPaths()[[1L]])
#> The following package(s) were installed from sources, but may be available from the following remotes:
#> 
#> - renv       [rstudio/renv]
#> - rsconnect  [rstudio/rsconnect]
#> 
#> Would you like to use these remote sources for these packages?
#> See 'snapshot.inference' in ?renv::config for more details.
#> Otherwise, these packages will be recorded with an unknown source.
#> [1] TRUE
```

<sup>Created on 2023-07-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

aronatkins:
Note: The formatting of this message has changed in the latest version of renv; it is much more compact:

``` r
td <- tempfile()
dir.create(td)
fn <- file.path(td, "index.Rmd")
writeLines(c("hello, markdown"), fn)
renv:::renv_snapshot_preflight_check_sources(td, .libPaths()[[1L]])
#> The following package(s) were installed from sources, but may be available from the following remotes:
#> - renv       [rstudio/renv]
#> - rsconnect  [rstudio/rsconnect]
[...]

PR #1569, opened 2023-07-19
tweak how missing packages / unknown sources are reported

kevinushey:
Closes https://github.com/rstudio/renv/issues/1559.

This PR tweaks when and how we report packages with unknown sources. In particular, we now only report those sources when creating a lockfile via a top-level call to `snapshot()`.

Honestly though, I'm starting to think it's not worth it for us to try and "fix" the remote source for packages installed from local sources like this.

### DIFF
--- a/R/activate.R
+++ b/R/activate.R
@@ -78,10 +78,9 @@ renv_activate_impl <- function(project,
     renv_imbue_self(project)
 
   # restart session if requested
-  if (restart && !is_testing())
-    return(renv_restart_request(project, reason = "renv activated"))
-
-  if (renv_rstudio_available())
+  if (restart)
+    renv_restart_request(project, reason = "renv activated")
+  else if (renv_rstudio_available())
     renv_rstudio_initialize(project)
 
   # try to load the project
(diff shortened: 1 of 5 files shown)


R/activate.R
```diff
@@ -78,10 +78,9 @@ renv_activate_impl <- function(project,
     renv_imbue_self(project)
 
   # restart session if requested
-  if (restart && !is_testing())
-    return(renv_restart_request(project, reason = "renv activated"))
-
-  if (renv_rstudio_available())
+  if (restart)
```
kevinushey, line 81: This change was needed as otherwise we don't run `load()` below, and so things like `the$project_path` are not set in tests.

hadley:
I'm fine with backing out of figuring out the exact remote reference — it can be quite slow (for obvious reasons) and since the dialog defaults to yes, it can add a decent amount of overhead. Even though I know what's going on behind the scenes, I also find the dialog moderately confusing as there's not enough information to suggest which option I should pick.

kevinushey:
What do you think about moving the remote inference into `renv::repair()`? So users could run that by hand if `renv` complained about a package with an unknown source that appeared to be from GitHub.

hadley:
Oh yeah, I like that!

kevinushey:
I filed https://github.com/rstudio/renv/issues/1574 so we can do that change in a separate PR.

Issue #1546, opened 2023-07-12
Error in eval(quote({ : object '..version..' not found


schoulten:
I'm experiencing this error locally and in one of my [GH Actions](https://github.com/schoulten/macro_report/actions/runs/5533292982) workflows. I'm not sure if it's due to the latest `renv` update (1.0.0) because I noticed it started happening after this commit ([#11a22a9](https://github.com/r-lib/actions/commit/11a22a908006c25fe054c4ef0ac0436b1de3edbe)).

I tried to create a reproducible example of what's happening in the workflow above. I believe the problem is installing a newer version of `renv` and activating the project without restarting the session (if I restart, the error doesn't occur). However, the [setup-renv](https://github.com/r-lib/actions/tree/v2/setup-renv) action seems to install a new version of renv and activate the project without restarting the session. And so I'm stuck with this error.

Here's the reproducible example:

```
# console
install.packages("remotes")
remotes::install_version("renv", "0.16.0")
```
Create an R file with any code, for example:

```
library(ggplot2)
print(head(ggplot2::economics_long))
```

And then continue with:

```
# console
renv::init()
renv::status()
renv::snapshot() # y
install.packages("remotes")
remotes::install_version("renv", "1.0.0")
renv::activate()
.rs.restartR()
```

After the last command, to restart the session, this error will be displayed:
```
# Error in eval(quote({ : object '..version..' not found
```

I apologize if the issue is not with `renv` but with the action.

kevinushey:
The problem here occurs because you still have an older version of renv loaded, but a newer version of renv is installed. So you end up with renv 0.17.3 trying to write an activate script, but instead of the activate script for renv 0.17.3 being available, the version for 1.0.0 is available instead, and because of a small change that was made this error occurs.

I'll need to think about how to make `renv` more robust in these scenarios, but the simple fix here is to call `renv::activate()` and then restart the R session.

kevinushey:
It's also possible that this recent change in the action will have resolved the issue:

https://github.com/r-lib/actions/commit/11a22a908006c25fe054c4ef0ac0436b1de3edbe

sgvignali:
This happens also with `renv::upgrade()`. In my case form version 0.17.3 in Windows.
The command also raises several warnings:

```r
renv::upgrade()
A new version of the renv package will be installed:

# CRAN ===============================
- renv   [0.17.3 -> 1.0.0]

This project will use the newly-installed version of renv.

Do you want to proceed? [y/N]: y
[...]

PR #1576, opened 2023-07-20
invoke activate in separate process

kevinushey:
Closes https://github.com/rstudio/renv/issues/1546.


hadley (approved):
I can't see anything obviously wrong, but I'm also not very familiar with this code.

PR #1586, opened 2023-07-24
Default to the project library, if it exists

hadley:


tests/testthat/test-snapshot.R
```diff
@@ -63,8 +63,8 @@ test_that("multiple libraries can be used when snapshotting", {
   lockfile <- snapshot(lockfile = NULL, library = libs, type = "all")
   records <- renv_lockfile_records(lockfile)
 
-  expect_length(records, 2L)
-  expect_setequal(names(records), c("bread", "toast"))
+  expect_length(records, 17) # 2 + 15 base/recommended packages
```
kevinushey, line 66: Maybe we should change `type` above and just supply an explicit dependency list?

NEWS.md
```diff
@@ -1,6 +1,9 @@
 
 # renv 1.1.0  (UNRELEASED)
 
+* If `library` is unsupplied, renv functions default to using the project
+  library, if `project` is supplied and the project library exists.
```
kevinushey, line 5: This reads pretty awkwardly to me. Can we give a more concrete example of the consequences of this change, or where users might see a change in behavior?

R/roxygen.R
```diff
@@ -9,7 +9,8 @@
 #' @param lockfile Path to a lockfile. When `NULL` (the default), the `renv.lock` located in the root of the current project will be used.
 #'
 #' @param library The \R library to be used. When `NULL`, the active project
-#'  library will be used instead.
+#'  library will be used instead. If no project is active, will fall back
+#'  to the system library.
```
kevinushey, line 13: I wonder if this could just be something like "When `NULL`, the current library paths will be used."? IIRC almost all places in `renv` just use the current library paths, with the assumption that `load()` would have set up the default library paths.

R/libpaths.R
```diff
@@ -191,11 +191,17 @@ renv_libpaths_restore <- function() {
...
-  if (is.null(library))
-    return(renv_libpaths_all())
+renv_libpaths_resolve <- function(library = NULL, project = NULL) {
+  if (!is.null(library)) {
+    return(unique(c(library, .Library)))
+  }
 
-  unique(c(library, .Library))
+  if (!is.null(project)) {
+    proj_lib <- renv_paths_library(project = project)
```
kevinushey, line 200: I'm not sure if this is the desired behavior. Suppose I do something like:

```
setwd("~/test/project")
renv::install("rlang")
```

And `~/test/project` does happen to have an `renv` library, but I haven't actually activated or loaded that project. IMHO, we should install into the current library paths, not the renv project library. That is, I think loading a project explicitly should be a pre-requisite to having `renv::install()` use the project library.

kevinushey (commented):
I'm not sure if I agree with the change. In general, I think `renv` APIs should almost always use the current library paths, and rely on `load()` to set up the library paths when a particular project has been loaded.

R/libpaths.R
hadley: But this PR doesn't change that code — this only affects that behaviour when you explicitly supply a `project`. In that case, I think it's reasonable to say that you want to operate on that project, regardless of what `project` is currently loaded in the current session (if any).

i.e. if you do `renv::install("rlang", project = "another-project")` don't you want to install into use `another-project`s library?

But maybe only installing to that project's library if it exists is wrong — if you say you want to install into a given project, and that project doesn't have a library, it seems like it would be better to error?

kevinushey: I think my concern is that `project` is usually going to be non-NULL internally, since we fall back to the current working directory when no project is active. For example:

```
> renv:::renv_project_resolve(NULL)
[1] "/Users/kevin"
```

And we use that pretty commonly when resolving the "active" project, e.g. we do that here:

https://github.com/rstudio/renv/pull/1586/files#diff-76ebee0a54b80584fcf1eff0cdf55ef6f17e47a6d80cb807b5939e8e50cac32cR128

I think my fears would be abated if we distinguished between the case where `project` was explicitly supplied by the user, versus we just used the default / current value.

kevinushey: Could we add some tests (with multiple `renv` projects) just to confirm the behavior we want here as well? I guess the matrix here is (A x B) / (loaded x unloaded) / (explicit project / implicit project).

PR #1587, opened 2023-07-24
Force GH auth in test setup

hadley:
I don't know what this does but it resolves my rate limit problem 😭


kevinushey:
Seems okay to me, but perhaps something about when / why / how `gitcreds::gitcreds_get()` is getting invoked? Maybe we should just run that in test setup or something?

Issue #1582, opened 2023-07-24
R session encounters fatal error when trying to use dev renv version 


ChaitaC:
**Note:** This issue can only be recreated when using the rstudio/renv version. 
when I install `remotes::install_cran("renv")` everything works as expected and session doesn't encounter any issue 

**Steps to reproduce:**
1. Navigate IDE 
2. run the following in the console to open an renv project
```
remotes::install_github("rstudio/renv")
remotes::install_version("rsconnect", "1.0.1")
usethis::create_project("~/Shared/workspace/renv-test")
```
3. Once a new session opens run the following commands to create your content.
```
writeLines(con = "app.R", c(
  "library(shiny)",
  "ui <- fluidPage('hello')",
  "server <- function(input, output) {}",
  "shinyApp(ui, server)"
))
writeLines(con = "deploy.R", c("library(rsconnect)"))
renv::init(repos = "https://cloud.r-project.org")
```

**Result:** R session encounters a fatal error

PR #1588, opened 2023-07-24
avoid multiple calls to restart in same session

kevinushey:
Some extra insurance against https://github.com/rstudio/renv/issues/1582.

### DIFF
--- a/R/restart.R
+++ b/R/restart.R
@@ -1,4 +1,7 @@
 
+# whether or not we're already trying to restart the session
+the$restarting <- FALSE
+
 renv_restart_request <- function(project = NULL, reason = "", ...) {
 
   project <- renv_project_resolve(project)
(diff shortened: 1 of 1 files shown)


R/restart.R
```diff
@@ -59,10 +62,22 @@ renv_restart_request_rstudio <- function(project, reason, ...) {
...
+
+}
+
+renv_restart_invoke <- function(callback) {
+
+  # avoid multiple attempts to restart in a single call, just in case
+  if (!the$restarting) {
+    the$restarting <- TRUE
+    callback()
+  }
```
hadley, line 81:
```suggestion
  }
  
  the$restarting <- FALSE
```

Just in case it fails?

hadley (approved):

R/restart.R
kevinushey: The problem here is that the restart will happen "eventually", so we don't know directly in R code if the restart was (or will be) successful at that point.

hadley: Yeah, I wondered about that. I just worry about getting stuck in some bad state if the restart fails.

Issue #1578, opened 2023-07-21
`renv::install()` fails when CRAN-like repositories are in `renv.lock`


ezraporter:
My institution uses a [`drat`](https://eddelbuettel.github.io/drat/) repository in our GitHub Enterprise to distribute packages internally. Since `renv 0.17.0` `renv::install()` seems to fail for **any** package (even CRAN packages) when our `drat` repo is in my `renv.lock`.

I did some digging and it seems like the regression may have occurred at https://github.com/rstudio/renv/commit/62c73d9c74117ceafec37c8e4a5c343aed149792 where a warning was upgraded to an error in [`available-packages.R`](https://github.com/rstudio/renv/blob/62c73d9c74117ceafec37c8e4a5c343aed149792/R/available-packages.R#L58).

## reprex

I created a `drat` repo at https://ezraporter.github.io/drat to show the issue:

`renv.lock`:
``` yaml
...
    "Repositories": [
      {
        "Name": "CRAN",
        "URL": "https://cran.rstudio.com"
      },
      {
        "Name": "OTHERAN",
        "URL": "https://ezraporter.github.io/drat"
      }
    ]
  ...
```

```r
# Fails
renv::install("dplyr")
```
<details>
<summary>Show error message</summary>

```r
Error: package 'dplyr' is not available
Traceback (most recent calls last):
15: renv::install("dplyr")
14: retrieve(packages)
13: handler(package, renv_retrieve_impl(package))
12: renv_retrieve_impl(package)
11: withCallingHandlers(renv_available_packages_latest(package), 
        error = function(err) stopf("package '%s' is not available", 
            package))
10: renv_available_packages_latest(package)
 9: map(errors$data(), warning)
 8: lapply(x, f, ...)
 7: FUN(X[[i]], ...)
 6: withRestarts({
        .Internal(.signalCondition(cond, message, call))
        .Internal(.dfltWarn(message, call))
    }, muffleWarning = function() NULL)
 5: withOneRestart(expr, restarts[[1L]])
 4: doWithOneRestart(return(expr), restart)
 3: <condition-handler>(...)
 2: stopf("package '%s' is not available", package)
 1: stop(sprintf(fmt, ...), call. = call.)
```
</details>

```r
# Works but issues some warning messages due to the drat repo not serving binaries
utils::install.packages("dplyr")

#> ...
#> Warning: unable to access index for repository https://ezraporter.github.io/drat/bin/macosx/big-sur-arm64/contrib/4.2:
[...]

PR #1589, opened 2023-07-25
tolerate errors when querying available packages

kevinushey:
Closes https://github.com/rstudio/renv/issues/1578.

### DIFF
--- a/tests/testthat/test-available-packages.R
+++ b/tests/testthat/test-available-packages.R
@@ -196,6 +196,5 @@ test_that("available_packages() tolerates missing repositories", {
   )
 
   expect_false(is.null(dbs[["CRAN"]]))
-  expect_true(is.null(dbs[["DRAT"]]))
 
 })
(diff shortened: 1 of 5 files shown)


tests/testthat/test-available-packages.R
```diff
@@ -194,3 +182,20 @@ test_that("we can query the R universe", {
   expect_identical(lhs, rhs)
 
 })
+
+test_that("available_packages() tolerates missing repositories", {
+  skip_on_cran()
+  skip_if_not(grepl("\\bbinary\\b", .Platform$pkgType))
+
+  dbs <- available_packages(
```
hadley, line 190: Add snapshot test?

hadley (approved):

tests/testthat/test-available-packages.R
kevinushey: What should the snapshot capture? (since the set of available packages from CRAN will change over time)

kevinushey: It's also possible that drat could start providing binaries in the future (or at least provide the infrastructure so `available.packages()` doesn't complain) so I think a snapshot test doesn't quite fit.

kevinushey: That is a very good point! I'll update in a separate PR.

Issue #1558, opened 2023-07-17
Warning messages on load after upgrading to renv 1.0.0


mgriffiths2:
I was using renv 0.17.3 with no problem, then upgraded to 1.0.0 using renv::upgrade()

Now whenever I open my project, or run renv::load() I get a list of warning messages:

```
- Project '<project name>' loaded. [renv 1.0.0]
Warning messages:
1: In is.na(value) :
  is.na() applied to non-(list or vector) of type 'language'
2: In if (is.na(value)) default else as.logical(value) :
  the condition has length > 1
3: In is.na(value) :
  is.na() applied to non-(list or vector) of type 'language'
4: In if (is.na(value)) default else as.logical(value) :
  the condition has length > 1
5: In is.na(value) :
  is.na() applied to non-(list or vector) of type 'language'
6: In if (is.na(value)) default else as.logical(value) :
  the condition has length > 1
7: In is.na(value) :
  is.na() applied to non-(list or vector) of type 'language'
8: In if (is.na(value)) default else as.logical(value) :
  the condition has length > 1
9: In is.na(value) :
  is.na() applied to non-(list or vector) of type 'language'
10: In if (is.na(value)) default else as.logical(value) :
  the condition has length > 1
```

It still seems to work fine, but it would be nice to get rid of the warnings.

kevinushey:
Do you get more detail if you use:

```
options(error = renv:::renv_error_handler_call())
options(warn = 2L)
```

Perhaps by putting this in your project's `.Rprofile` before `source("renv/activate.R")` is run?

kevinushey:
A reproducible example would also help; e.g. if you could share your project sources publicly (e.g. on GitHub).

mgriffiths2:
Unfortunately I'm not able to share the whole project. Would the renv lockfile help? 

Here is what I get when I load with the options you suggested above:

`Error in is.na(value) : 
  (converted from warning) (converted from warning) is.na() applied to non-(list or vector) of type 'language'
Traceback (most recent calls last):
35: source("renv/activate.R")
34: withVisible(eval(ei, envir))
33: eval(ei, envir)
32: eval(ei, envir)
31: local(...) at activate.R#2
30: eval.parent(substitute(eval(quote(expr), envir)))
29: eval(expr, p)
28: eval(expr, p)
27: eval(quote(...), new.env())
26: eval(quote(...), new.env())
25: if (renv_bootstrap_load(project, libpath, version))
[...]

PR #1590, opened 2023-07-25
handle calls in chunk options

kevinushey:
Closes https://github.com/rstudio/renv/issues/1558.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -1710,7 +1710,7 @@ renv_dependencies_report <- function(errors) {
 renv_dependencies_eval <- function(expr) {
 
   # create environment with small subset of symbols
-  syms <- c("list", "c")
+  syms <- c("list", "c", "!", "+", "-", "*", "/", "||", "&&")
   vals <- mget(syms, envir = baseenv())
   envir <- list2env(vals, parent = emptyenv())
 
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -1712,8 +1712,9 @@ renv_dependencies_eval <- function(expr) {
   # create environment with small subset of "safe" symbols, that
   # are commonly used for chunk expressions
   syms <- c(
-    "list", "c",
-    "::", ":::", "$", "@", "[", "[[",
+    "list", "c", "T", "F",
+    "{", "(", "[", "[[",
+    "::", ":::", "$", "@",
     ":",
     "+", "-", "*", "/",
     "<", ">", "<=", ">=", "==", "!=",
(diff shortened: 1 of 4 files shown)


R/dependencies.R
```diff
@@ -1710,7 +1710,7 @@ renv_dependencies_report <- function(errors) {
 renv_dependencies_eval <- function(expr) {
 
   # create environment with small subset of symbols
-  syms <- c("list", "c")
+  syms <- c("list", "c", "!", "+", "-", "*", "/", "||", "&&")
```
kevinushey, line 1713: Do you think there's anything else we should include for a subset of commonly-used things in chunk options?

kevinushey:
FWIW you mentioned:

> It's probably worth breaking up truthy() into two functions — one can be used for env vars (90% of the existing usage) and the other for parsing knitr options.

but I think we can do that in a separate PR?

R/truthy.R
```diff
@@ -1,5 +1,13 @@
 
-truthy <- function(value, default) {
+truthy <- function(value, default = FALSE) {
```
hadley, line 2: I think you should make the code parsing stuff a separate function.

R/dependencies.R
```diff
@@ -1709,8 +1709,18 @@ renv_dependencies_report <- function(errors) {
 
 renv_dependencies_eval <- function(expr) {
 
-  # create environment with small subset of symbols
-  syms <- c("list", "c")
+  # create environment with small subset of "safe" symbols, that
+  # are commonly used for chunk expressions
+  syms <- c(
+    "list", "c",
+    "::", ":::", "$", "@", "[", "[[",
```
hadley, line 1716: Not sure it's worth `::` and `:::` if you don't otherwise support a bigger set of base functions.

R/dependencies.R
```diff
@@ -1709,8 +1709,18 @@ renv_dependencies_report <- function(errors) {
 
 renv_dependencies_eval <- function(expr) {
 
-  # create environment with small subset of symbols
-  syms <- c("list", "c")
+  # create environment with small subset of "safe" symbols, that
```
hadley, line 1712: In general, I'm not sure it's worth bothering to expand this since most calls will be like the example in the linked issue, i.e. `!use_cache`, where `use_cache` is computed in the doc.

hadley (approved):

R/truthy.R
hadley: Oh sorry, I didn't see your overall comment.

PR #1591, opened 2023-07-25
ensure renv unloaded on devtools::load_all()

kevinushey:
Fixes an issue where the R session could crash on restart if `devtools::load_all()` were called multiple times.


hadley (approved):
Confirmed that this no longer crashes RStudio for me, after multiple devtools re-loads.

Issue #1583, opened 2023-07-24
`init()` needs `load = FALSE`


hadley:
So you can set up renv in another directory. I'm thinking particularly of the rsconnect tests where I want this.

Unless there's some other way of doing this that I'm missing?

kevinushey:
That seems reasonable to me. Would that basically just control whether this code is run? https://github.com/rstudio/renv/blob/ae783a9879162eb60a28aff9886eef6e72ea33bd/R/init.R#L148-L149

kevinushey:
Might also need some tweaks around things like https://github.com/rstudio/renv/blob/ae783a9879162eb60a28aff9886eef6e72ea33bd/R/init.R#L89-L91.

kevinushey:
Actually, this reminds me of `renv::scaffold()`, which is basically "init without loading". Does that do what you need here?

hadley:
That's useful but I still need to do a few steps by hand:

``` r
path <- tempfile()
dir.create(path)

writeLines("library(MASS); library(foreign)", file.path(path, "deps.R"))
renv::scaffold(path)
#> - Lockfile written to "/private/tmp/RtmpEA53xn/file149ae7dbe29e1/renv.lock".
#> - renv infrastructure has been generated for project "/private/tmp/RtmpEA53xn/file149ae7dbe29e1".
renv::snapshot(path)
#> The following package(s) will be updated in the lockfile:
#> 
#> # CRAN -----------------------------------------------------------------------
#> - foreign   [* -> 0.8-84]
#> - MASS      [* -> 7.3-60]
#> 
#> - Lockfile written to "/private/tmp/RtmpEA53xn/file149ae7dbe29e1/renv.lock".
renv::hydrate(project = path, prompt = FALSE, library = renv:::renv_paths_library(project = path))
#> The following packages were discovered:
#> 
#> # ~/Library/R/arm64/4.2/library ----------------------------------------------
#> - foreign   0.8-84
#> - MASS      7.3-60
#> 
#> They will be linked into the project library.
#> 
#> - Linking packages into the project library ... Done!
#> - Hydrated 2 packages in 75 milliseconds.
```

<sup>Created on 2023-07-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

(the hydrate call will get simpler if I can figure out #1586)

PR #1593, opened 2023-07-25
init gains load argument

kevinushey:
Closes https://github.com/rstudio/renv/issues/1583.

### DIFF
--- a/R/init.R
+++ b/R/init.R
@@ -56,6 +56,8 @@ the$init_running <- FALSE
 #'   to `TRUE` to use the default version of Bioconductor recommended by the
 #'   BiocManager package.
 #'
+#' @param load Boolean; should the project be loaded after it is initialized?
+#'
 #' @param restart Boolean; attempt to restart the \R session after initializing
 #'   the project? A session restart will be attempted if the `"restart"` \R
 #'   option is set by the frontend embedding \R.
@@ -71,6 +73,7 @@ init <- function(project = NULL,
                  force        = FALSE,
                  repos        = NULL,
                  bioconductor = NULL,
+                 load         = TRUE,
                  restart      = interactive())
 {
   renv_consent_check()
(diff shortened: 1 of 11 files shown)


tests/testthat/helper-scope.R
```diff
@@ -38,7 +38,7 @@ renv_tests_scope <- function(packages = character(),
   renv_scope_libpaths(library, scope = scope)
 
   # return path to project directory
-  invisible(project)
+  invisible(renv_path_normalize(project))
```
kevinushey, line 41: To avoid troubles with this:

```
> c(tempdir(), normalizePath(tempdir()))
[1] "/var/folders/9n/3nxsgkrj0rlfx196kzdttpww0000gn/T//RtmpbpCVk1"       
[2] "/private/var/folders/9n/3nxsgkrj0rlfx196kzdttpww0000gn/T/RtmpbpCVk1"
```

R/init.R
```diff
@@ -82,6 +85,13 @@ init <- function(project = NULL,
   project <- renv_path_normalize(project %||% getwd())
   renv_project_lock(project = project)
 
+  # if we're not loading this project, make sure we reset
+  # the working directory and repos when we're done
+  if (!load) {
+    renv_scope_options(repos = getOption("repos"))
+    renv_scope_wd(dir = getwd())
```
hadley, line 92: We shouldn't need to reset the wd, right? Because that's already conditioned behind the `load`?

I think we can also just fail to set the `repos` option since `repos` is already explicitly passed to `snapshot()`? Is there a bug in `init(bioconductor = TRUE)`? Since it just sets the global repos options, which isn't consulted by snapshot, that repo doesn't end up in the lockfile.

tests/testthat/test-init.R
```diff
@@ -229,3 +229,18 @@ test_that("init() prompts the user for the snapshot type", {
...
+test_that("a project can be initialized without loading it", {
+  skip_on_cran()
+
+  project <- renv_tests_scope()
+  init()
+  expect_equal(renv_project_get(), project)
+
+  other <- renv_scope_tempfile("renv-project-")
+  ensure_directory(other)
+  init(project = other, load = FALSE, restart = FALSE)
```
hadley, line 242: Maybe worth explicitly checking that `load = FALSE` ensures that wd, repos, and libpaths are unchanged?

R/init.R
kevinushey: Nice catch. Let me see if I can fix that up.

hadley (approved):
Nice!

PR #1595, opened 2023-07-25
Make upgrade quieter

hadley:
And use better test for version

### DIFF
--- a/R/upgrade.R
+++ b/R/upgrade.R
@@ -1,10 +1,13 @@
 
 #' Upgrade renv
 #'
+#' @description
 #' Upgrade the version of renv associated with a project, including using
-#' a development version from GitHub. If you want to update all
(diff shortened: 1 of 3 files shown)


R/upgrade.R
```diff
@@ -54,14 +54,15 @@ renv_upgrade_impl <- function(project, version, reload, prompt) {
 
   reload <- reload %||% renv_project_loaded(project)
 
-  old <- renv_snapshot_description(package = "renv")
+  lockfile <- renv_lockfile_load(project)
```
hadley, line 57: I don't know if this is correct or not, but `renv_snapshot_description()` returns a rather different data structure so the comparison below was always `FALSE`.

kevinushey (approved):
LGTM!

PR #1597, opened 2023-07-25
avoid leaking flush_console definition

kevinushey:
Since right now, in an `renv` project:

```
$ R -s -e 'ls()'
[1] "flush_console"
```


hadley (approved):
Woops

Issue #1571, opened 2023-07-20
Running renv::restore() in parallel will fail


Kjir:
I have some Airflow tasks where I launch a Kubernetes Pod and in that Pod I use `renv` to install and run code from a package.
Since the task is, essentially, "copy data from A to B for date X", there can be multiple instance of this task running in parallel for different days.

If the packages are not in the cache, `renv` will attempt to install the missing packages in all the parallel jobs. Unfortunately, only the first job to finish installing will actually be sucessful and the other jobs will fail with an error:

```
Error: target file '/renv/cache/v5/linux-ubuntu-jammy/R-4.3/x86_64-pc-linux-gnu/kofcasts/1.2.1/51a5c034ff08245557a03533120c02e9/kofcasts' already exists
 - Installing kofcasts ...                       Traceback (most recent calls last):
 12: renv::install(package_name)
 11: renv_install_impl(records)
 10: renv_install_staged(records)
  9: renv_install_default(records)
  8: handler(package, renv_install_package(record))
  7: renv_install_package(record)
  6: renv_cache_synchronize(record, linkable = linkable)
  5: renv_cache_synchronize_impl(cache, record, linkable, path)
  4: renv_cache_copy(path, cache, overwrite = TRUE)
  3: renv_file_copy(source, target, overwrite = overwrite)
  2: renv_file_copy_dir(source, target)
  1: stop(status)
 Execution halted
```

Does this really need to be an error condition?
Ideally it would acknowledge that the desired package has been installed in the cache in the meantime and move on.

kevinushey:
I believe https://github.com/rstudio/renv/pull/1598 will help resolve the issue you're seeing. Could you try testing with:

```
renv::install("rstudio/renv#1598")
```

and test if things appear better on your side?

Kjir:
I tried and reproduced the error running this in parallel in two terminals:

```
$ docker run --rm -it -v /tmp/rcache:/renv -e RENV_PATHS_CACHE=/renv rocker/verse R --vanilla -e "options(repos = c(CRAN='https://packagemanager.posit.co/cran/latest')); install.packages('renv'); renv::install('tsbox')"
```

I then tried to use the fixed version (clearing the cache beforehand `sudo rm -rf /tmp/rcache/*`):

```
[...]

PR #1598, opened 2023-07-25
use lock when synchronizing with cache

kevinushey:
Closes https://github.com/rstudio/renv/issues/1571.


hadley (approved):
How do locks work in renv? Will it block until the lock is available?

kevinushey:
> How do locks work in renv? Will it block until the lock is available?

Yep, that's right.

Issue #1573, opened 2023-07-20
renv 1.0.0 check on startup can lead to long session loading time


michalovadek:
if an `.*ignore` file is missing or misspecified, loading an R project with `renv` that contains many files will take a long time because `renv` 1.0.0 checks all files at startup by default, _without a prior warning or prompt_ (see also discussion here: https://github.com/rstudio/renv/issues/1181#issuecomment-1640508295)

How I replicated the issue: I created a new R project in a new directory and created 100 000 dummy .txt files in subdirectory `data`. Opening the project hangs RStudio seemingly until the files are scanned. Adding `data` to `.gitignore` "fixes" it.

I think the user should be prompted whether they want to do the startup check if it involves a high number of files, otherwise it should be skipped. At very minimum they must be informed, as when running `renv::status()`. In a complex project it took me a while to figure out what was causing the slow loading time and at first did not realize this was related to updating renv to 1.0.0.

`sessionInfo()`:

R version 4.2.3 (2023-03-15 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19044)

Matrix products: default

locale:
[1] LC_COLLATE=English_United Kingdom.utf8  LC_CTYPE=English_United Kingdom.utf8   
[3] LC_MONETARY=English_United Kingdom.utf8 LC_NUMERIC=C                           
[5] LC_TIME=English_United Kingdom.utf8    

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base     

loaded via a namespace (and not attached):
[1] compiler_4.2.3 tools_4.2.3    renv_1.0.0

kevinushey:
We do display the following on `renv::init()`:

```
> renv::init()
A large number of files (100000 in total) have been discovered.
It may take renv a long time to crawl these files for dependencies.
Consider using .renvignore to ignore irrelevant files.
See `?renv::dependencies` for more information.
Set `options(renv.config.dependencies.limit = Inf)` to disable this warning.
```

We suppress output like this here, though: https://github.com/rstudio/renv/blob/2d62bfa0a4e15275e5e0edc67c2dd48f1838431d/R/load.R#L844

michalovadek:
even if there is a case for suppressing other types of output, I would say here the default option should be to display it whenever there is a chance the function will take a while to run. Since `renv::init()` is not run manually most of the time (I suspect?), the user actually does not know what's causing the hang.

PR #1603, opened 2023-07-26
notify user if directory contains lots of files on load

kevinushey:
Closes https://github.com/rstudio/renv/issues/1573.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -391,6 +391,9 @@ renv_dependencies_find_dir <- function(path, root, depth) {
   # list children
   children <- renv_dependencies_find_dir_children(path, root, depth)
 
+  # notify about number of children
+  renv_condition_signal("renv.dependencies.count", list(path = path, count = length(children)))
+
   # find recursive dependencies
   depth <- depth + 1
   paths <- map(children, renv_dependencies_find_impl, root = root, depth = depth)
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -898,6 +898,10 @@ renv_snapshot_dependencies_impl <- function(project, type = NULL, dev = FALSE) {
     }
   )
 
+  # count the number of files in each directory, so we can report
+  # to the user if we scanned a folder containing many files
+  count <- integer()
+
   packages <- withCallingHandlers(
 
     renv_dependencies_impl(
@@ -919,6 +923,11 @@ renv_snapshot_dependencies_impl <- function(project, type = NULL, dev = FALSE) {
 
     },
 
+    # collect information about folders containing lots of files
+    renv.dependencies.count = function(cnd) {
+      count[[cnd$data$path]] <<- cnd$data$count
+    },
+
     # notify the user if we took a long time to discover dependencies
     renv.dependencies.elapsed_time = function(cnd) {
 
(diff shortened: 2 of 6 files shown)


R/snapshot.R
```diff
@@ -932,15 +941,26 @@ renv_snapshot_dependencies_impl <- function(project, type = NULL, dev = FALSE) {
...
         "Consider using .renvignore to ignore files, or switching to explicit snapshots.",
         "See `?renv::dependencies` for more information.",
+        if (length(count)) c(
+          "",
+          sprintf("- %s: %s", format(names(count)), nplural("file", count))
+        ),
         ""
       )
 
+      renv_scope_options(renv.verbose = TRUE)
```
hadley, line 963: I wonder if this suggests that the `writef()` should actually be a warning?

R/snapshot.R
```diff
@@ -932,15 +941,26 @@ renv_snapshot_dependencies_impl <- function(project, type = NULL, dev = FALSE) {
       if (elapsed < limit)
         return()
 
+      renv_scope_options(renv.verbose = TRUE)
```
hadley, line 944: Duplicated below

R/dependencies.R
```diff
@@ -391,6 +391,9 @@ renv_dependencies_find_dir <- function(path, root, depth) {
   # list children
   children <- renv_dependencies_find_dir_children(path, root, depth)
 
+  # notify about number of children
+  renv_condition_signal("renv.dependencies.count", list(path = path, count = length(children)))
```
hadley, line 395: Is this used anywhere?

tests/testthat/test-dependencies.R
```diff
@@ -462,3 +462,20 @@ test_that("we tolerate calls when parsing dependencies", {
...
+  # create data directory with 'lots' of files
+  dir.create("data")
+  setwd("data")
+  files <- sprintf("%03i.R", 1:200)
+  file.create(files)
+  setwd("..")
+
+  # try to collect snapshot dependencies
+  renv_scope_options(renv.dependencies.elapsed_time_threshold = 0)
+  expect_snapshot(. <- renv_snapshot_dependencies(project))
```
hadley, line 479: If we stick with the `writef()`, I'd suggest adding a `renv_scope_options(renv.verbose = FALSE)` to assert that the warning is displayed even when not verbose.

hadley (approved):

R/snapshot.R
kevinushey: I feel like we need something like `ewritef()`, or some other version of `writef()` that ignores `renv.verbose`.

R/dependencies.R
kevinushey: Yes, see https://github.com/rstudio/renv/pull/1603/files#diff-5d849f3782ddfdd6b0053b841b4be729f4fdaa33c9262caf6757aa185e5f5658R927-R929, which fills in the `count` variable, and https://github.com/rstudio/renv/pull/1603/files#diff-5d849f3782ddfdd6b0053b841b4be729f4fdaa33c9262caf6757aa185e5f5658R946-R948 where we use it.

kevinushey:
I'm going to merge this and consider your comments as follow-up for https://github.com/rstudio/renv/issues/1607.

PR #1604, opened 2023-07-26
use ASCII quotes consistently

kevinushey:


aronatkins (approved):

hadley (approved):

kevinushey:
@hadley anything that needs to happen on your side to avoid inserting fancy quote?

PR #1609, opened 2023-07-27
Split truthy() into two functions

### LINKED ISSUE #1592: split truthy into separate functions
Right now, it also handles calls, specifically for dependency discovery:

https://github.com/rstudio/renv/blob/dd4f2b094e9ab1c79917f2e34b51f9975b5a0eef/R/dependencies.R#L769-L772

We should instead have a separate function for safely evaluating these expressions.

hadley:
* `truthy()` for env vars and options
* `renv_dependenciy_param_is_true()` for knitr chunk options

Fixes #1592

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -769,7 +769,7 @@ renv_dependencies_discover_chunks_ignore <- function(chunk) {
   # if renv.ignore is set, respect it
   ignore <- chunk$params[["renv.ignore"]]
   if (!is.null(ignore))
-    return(truthy(ignore))
+    return(renv_dependencies_param_is_true(ignore))
 
   # skip non-R chunks
   engine <- chunk$params[["engine"]]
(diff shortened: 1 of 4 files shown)


R/dependencies.R
```diff
@@ -1707,6 +1707,22 @@ renv_dependencies_report <- function(errors) {
...
+
+renv_dependencies_param_is_true <- function(x, default = TRUE) {
+  # https://github.com/rstudio/renv/issues/1558
+  if (is.call(x) || is.symbol(x)) {
+    x <- tryCatch(renv_dependencies_eval(x), error = identity)
+    if (inherits(x, "error"))
+      return(default)
+  }
+
+  if (is.logical(x) && length(x) == 1 && !is.na(x)) {
```
kevinushey, line 1719: For a chunk header like:

`````
```{r, renv.ignore=0}
# blah
```
`````

We would end up ignoring the value of `0` as it's not logical?

In other words, I think we should fall back to `truthy()` here instead for whatever value the user has supplied, or just coerce non-logical values to logical.

hadley: Sorry, I didn't realise knitr was so loosely typed here. Even `include = "false"` works 😭

kevinushey: "Be liberal in what you accept", as they say. 😉

hadley: FWIW `truthy()` isn't quite right either because `include = "True"` doesn't work. But we can coerce to logical.

Issue #1608, opened 2023-07-27
snapshot does not report installed-from-source packages when renv.verbose = FALSE


aronatkins:
rsconnect calls `renv::snapshot` with `renv.verbose = FALSE`. From a push-button deploy session, we see:

```
ℹ Capturing R dependencies with renv
Error in renv_snapshot_validate_report(valid, prompt, force) : 
  aborting snapshot due to pre-flight validation failure
Calls: <Anonymous> ... snapshotRenvDependencies -> <Anonymous> -> renv_snapshot_validate_report
Execution halted
```

In https://github.com/rstudio/rsconnect/issues/934, we believe that this may have been caused by installing the dflowR package from source rather than from GitHub.

From the console, this looks like:

```r
options(renv.verbose = FALSE)
deps <- renv::dependencies(getwd())
renv::snapshot(getwd(), packages = deps$Package, prompt = FALSE)
#> Error in renv_snapshot_validate_report(valid, prompt, force) : 
#>   aborting snapshot due to pre-flight validation failure
options(renv.verbose = NULL)
renv::snapshot(getwd(), packages = deps$Package, prompt = FALSE)
#> The following package(s) were installed from an unknown source:
#> - dflowR [0.01.1]
#> renv may be unable to restore these packages in the future.
#> Consider reinstalling these packages from a known source (e.g. CRAN).
#> 
#> Error in renv_snapshot_validate_report(valid, prompt, force) : 
#>   aborting snapshot due to pre-flight validation failure
```

renv should report problems even when `renv.verbose=FALSE`

kevinushey:
I've merged https://github.com/rstudio/renv/pull/1610; @aronatkins can you let me know if all looks well now?

aronatkins:
See https://github.com/rstudio/renv/issues/1607#issuecomment-1656155120

I tested this issue and #1607 together; half way there. Living on a prayer? 😁

aronatkins:
Outside the deploy pane, we do not receive a stack.

```
ℹ Capturing R dependencies with renv
Error in renv_snapshot_validate_report(valid, prompt, force) : 
  aborting snapshot due to pre-flight validation failure
```

aronatkins:
From the RStudio IDE deploy pane (using push-button deploy):

```
ℹ Capturing R dependencies with renv
The following package(s) were installed from an unknown source:
- dflowR [0.01.1]
renv may be unable to restore these packages in the future.
[...]

Issue #1607, opened 2023-07-27
snapshot does not report missing packages when renv.verbose = FALSE


aronatkins:
Use a non-renv project where the R library does not contain a package. In my case, the R library is missing tidyr and there is an R file containing a `library(tidyr)` statement.

```r
options(renv.verbose = FALSE)
deps <- renv::dependencies(getwd())
renv::snapshot(getwd(), packages = deps$Package, prompt = FALSE)
#> What do you want to do? 
#> 
#> 1: Snapshot, just using the currently installed packages.
#> 2: Install the packages, then snapshot.
#> 3: Cancel, and resolve the situation on your own.
#> 
#> Selection: 3
#> - Operation canceled.
options(renv.verbose = NULL)
renv::snapshot(getwd(), packages = deps$Package, prompt = FALSE)
#> The following required packages are not installed:
#> - tidyr
#> Packages must first be installed before renv can snapshot them.
#> Use `renv::dependencies()` to see where this package is used in your project.
#> 
#> What do you want to do? 
#> 
#> 1: Snapshot, just using the currently installed packages.
#> 2: Install the packages, then snapshot.
#> 3: Cancel, and resolve the situation on your own.
#> 
#> Selection: 3
#> - Operation canceled.
```

When run in a non-interactive session, the output is even less helpful:

```
- Operation canceled.
Execution halted
```

kevinushey:
We should audit usages of `writef()` that are used to report error output, and use a different function (`ewritef()`?) that ignores `renv.verbose` (or respects a different variable if people _really_ want that to be quiet)

kevinushey:
Open to alternatives too; e.g. different names; override in `writef()` to force printing; whatever else feels clean.

kevinushey:
`info()` / `caution()`?

kevinushey:
https://github.com/rstudio/renv/pull/1603#pullrequestreview-1549692359

kevinushey:
I've merged https://github.com/rstudio/renv/pull/1610; @aronatkins can you let me know if all looks well now?

aronatkins:
@kevinushey Sadly, no. I'm seeing the following in the deploy pane:

```
ℹ Capturing R dependencies with renv
Traceback (most recent calls last):
9: rsconnect::deployApp(appDir = "~/dev/customer/rsconnect-934/FlowDataPull", 
[...]

PR #1610, opened 2023-07-27
avoid using 'writef' for warning / error style output

kevinushey:
Addresses https://github.com/rstudio/renv/issues/1607.
Addresses https://github.com/rstudio/renv/issues/1608.

Still needs tests.
Still needs NEWS.


R/available-packages.R
```diff
@@ -89,13 +89,10 @@ renv_available_packages_query <- function(type, repos, quiet = FALSE) {
   if (empty(errors))
     return(dbs)
 
-  if (!is_testing())
```
kevinushey, line 92: I wonder if `renv_pretty_print` needs a new name, and if it should just write output by default. We almost always use it for "warning"-style output.

R/pretty.R
```diff
@@ -1,21 +1,17 @@
...
-  msg$push("")
+  lines <- c(
+    if (length(preamble)) paste(preamble, collapse = "\n"),
+    if (length(values))   paste("-", values, collapse = "\n"),
+    if (length(postamble)) paste(postamble, collapse = "\n"),
+    ""
+  )
 
-  text <- paste(as.character(msg$data()), collapse = "\n")
+  text <- paste(as.character(lines), collapse = "\n")
```
hadley, line 14:
```suggestion
  text <- paste(lines, collapse = "\n")
```

R/utils.R
```diff
@@ -121,6 +121,9 @@ ask <- function(question, default = FALSE) {
   if (!interactive())
     return(default)
 
+  # be verbose in this scope, as we're asking the user for input
+  renv_scope_options(renv.verbose = TRUE)
```
hadley, line 125: That only affects the single `writef()` below, right?

tests/testthat/_snaps/load.md
```diff
@@ -32,5 +32,10 @@
       load()
     Output
       - Project '<wd>' loaded. [renv <version>]
+      The following package(s) are out of sync [lockfile -> library]:
```
kevinushey, line 35: We probably need to suppress this kind of output in this scope.

R/abi.R
```diff
@@ -45,7 +45,7 @@ renv_abi_check <- function(packages = NULL,
   reasons <- unique(tbl$reason)
   if ("Rcpp_precious_list" %in% reasons) {
     packages <- sort(unique(tbl$package[tbl$reason == "Rcpp_precious_list"]))
-    renv_pretty_print(
+    notice(
```
aronatkins, line 48: Unrelated: I was experimenting with `renv_abi_check` last week and did not see it reporting issues (Linux container using R-4.3) when using mismatched Rcpp & httpuv. I don't have that environment available at the moment, unfortunately. 🚴 ➕ 🌧️ ➕ 💻 ➡️ 💥 😭

R/cache.R
```diff
@@ -271,7 +271,7 @@ renv_cache_diagnose_corrupt_metadata <- function(paths, problems, verbose) {
 
     # nocov start
     if (verbose) {
-      renv_pretty_print(
+      notice(
```
aronatkins, line 274: Each of these feel like they should not be subject to the verbose control because it is reporting specific instances of corrupt metadata. Is there a reason for the caller to suppress this output? Is that rare enough that they could use `capture.output`?

R/hydrate.R
```diff
@@ -345,7 +345,7 @@ renv_hydrate_resolve_missing <- function(project, library, na) {
       sprintf("[%s]: %s", package, short)
     })
 
-    renv_pretty_print(
+    notice(
```
aronatkins, line 348: Should this be a caution? It tells us why hydration was unsuccessful.

R/install.R
```diff
@@ -721,7 +721,7 @@ renv_install_preflight_requirements <- function(records) {
   fmt <- "Package '%s' requires '%s', but '%s' will be installed"
   text <- sprintf(fmt, format(package), format(requires), format(actual))
   if (renv_verbose()) {
-    renv_pretty_print(
+    notice(
```
aronatkins, line 724: This is announcing an installation problem; should it be a caution not guarded by verbose?

R/install.R
```diff
@@ -781,7 +781,7 @@ renv_install_preflight_permissions <- function(library) {
     postamble <- sprintf(fmt, info$effective_user %||% info$user)
 
     # print it
-    renv_pretty_print(
+    notice(
```
aronatkins, line 784: This message announces a permission problem; should it be a caution?

R/notice.R
```diff
@@ -0,0 +1,34 @@
...
+    if (length(values))   paste("-", values, collapse = "\n"),
+    if (length(postamble)) paste(postamble, collapse = "\n"),
+    ""
+  )
+
+  text <- paste(as.character(lines), collapse = "\n")
+  renv_notice_impl(text, emitter)
+}
+
+renv_notice_impl <- function(text, emitter = NULL) {
```
aronatkins, line 22: vetiver used `renv_pretty_print_impl`; you've changed its name.

R/snapshot.R
```diff
@@ -462,7 +462,7 @@ renv_snapshot_validate_dependencies_available <- function(project, lockfile, lib
 
   })
 
-  renv_pretty_print(
+  notice(
```
aronatkins, line 465: Most of the feedback in this file appears to be reporting a problem with the active environment; should they be `caution`?

R/dependencies.R
```diff
@@ -1695,10 +1695,10 @@ renv_dependencies_report <- function(errors) {
   enumerate(splat, function(file, problem) {
     messages <- paste("Error", problem$message, sep = ": ", collapse = "\n\n")
     text <- c(header(file), messages, "")
-    writef(text)
+    caution(text)
```
hadley, line 1698: Is it worth turning this into a `notice()`? It does have a preamble, postamble, and basically a bulleted interior.

R/notice.R
```diff
@@ -0,0 +1,34 @@
+
+# notice me senpai
+notice <- function(preamble,
```
hadley, line 3: How about `caution_bullets()` or `caution_list()` as name? I think this should use `caution` as a prefix to tie it more closely to `caution()`.

R/notice.R
```diff
@@ -0,0 +1,34 @@
...
+                   values,
+                   postamble = NULL,
+                   emitter = NULL)
+{
+  if (empty(values))
```
hadley, line 8: Maybe this should respect `renv.caution.verbose` or similar?

R/notice.R
```diff
@@ -0,0 +1,34 @@
...
+  # We should think of a cleaner way of exposing this.
+  # https://github.com/rstudio/renv/issues/1413
+  emitter <- emitter %||% {
+    getOption("renv.pretty.print.emitter", default = caution)
+  }
+
+  emitter(text)
+  invisible(NULL)
+
+}
```
hadley, line 34: I think `caution()` should be defined here too.

R/snapshot.R
```diff
@@ -960,8 +958,9 @@ renv_snapshot_dependencies_impl <- function(project, type = NULL, dev = FALSE) {
         ""
       )
 
-      renv_scope_options(renv.verbose = TRUE)
-      writef(lines, renv_difftime_format(elapsed))
+      # force output in this scope
+      renv_scope_caution(TRUE)
```
hadley, line 962: Isn't this the default?

R/load.R
```diff
@@ -842,18 +842,19 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
...
     )
-    writef(msg)
+    caution(msg)
     return(FALSE)
   }
 
   # otherwise, use status to detect if we're synchronized
   info <- local({
     renv_scope_options(renv.verbose = FALSE)
+    renv_scope_caution(FALSE)
```
hadley, line 852: Are you sure this should be set fo `FALSE` here?

hadley (approved):

R/utils.R
kevinushey: It looks like there's 3 writef's below, although two are just there to print a newline. But yeah, that's right.

R/load.R
kevinushey: Yes, because we want to avoid the regular "noisy" output from status in this scope -- instead, we want to ask users to run `renv::status()` themselves to see the full log.

R/snapshot.R
kevinushey: It is, but it's forced TRUE here because we want to display this message even if caution was disabled (e.g. as we do in the "synchronized" check)

PR #1615, opened 2023-07-28
be verbose if we're going to prompt

kevinushey:
Since the output we display might be controlled by `renv.verbose`.

### DIFF
--- a/R/scope.R
+++ b/R/scope.R
@@ -390,11 +390,11 @@ renv_scope_caution <- function(value) {
   )
 }
 
-renv_scope_verbose_if <- function(value) {
+renv_scope_verbose_if <- function(value, scope = parent.frame()) {
   if (value) {
     renv_scope_options(
-      renv.verbose = value,
-      scope = parent.frame()
+      renv.verbose = TRUE,
+      scope = scope
     )
   }
 }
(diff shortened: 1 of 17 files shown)


aronatkins (approved):
Should the documentation for any of the `prompt` arguments or the `renv.verbose` option discuss their interactions?

kevinushey:
Hmm... we don't actually document the use of `renv.verbose` anywhere. Maybe it deserves a mention in a vignette?

R/scope.R
```diff
@@ -389,3 +389,12 @@ renv_scope_caution <- function(value) {
...
 }
+
+renv_scope_verbose_if <- function(value) {
+  if (value) {
+    renv_scope_options(
+      renv.verbose = value,
+      scope = parent.frame()
+    )
+  }
+}
```
hadley, line 400:
```suggestion
renv_scope_verbose_if <- function(value, scope = parent.frame()) {
  if (value) {
    renv_scope_options(
      renv.verbose = TRUE,
      scope = scope
    )
  }
}
```

`value` -> `TRUE` makes the logic a bit clearer IMO and I think it's good practice to always have a` scope` argument.

hadley (approved):
Agree that it would be good to document this somewhere, but it's not clear where.

PR #1624, opened 2023-08-03
Honor R_LIBCURL_SSL_REVOKE_BEST_EFFORT

zeehio:
`R_LIBCURL_SSL_REVOKE_BEST_EFFORT` was introduced in R on R-4.2 at  https://github.com/wch/r-source/commit/f1ec503e986593bced6720a5e9099df58a4162e7.

If that environment variable is set to `TRUE`, then curl relaxes the certificate revocation checks. This is needed on Windows systems using curl with the schannel backend on environments with man in the middle https certificates (typically corporate environments), since those certificate revocation checks otherwise fail.

The `curl` package already disables those checks on windows as discussed at https://github.com/jeroen/curl/issues/306.

To avoid having to declare a custom configuration for each package, I suggest to honor that same environment variable in `renv` as well.

I manually verified (on a Windows system with a MITM certificate authority) that renv was giving certificate revocation check errors and that with this fix the errors are gone. Besides, I did the following command line curl tests on such system:

The command:

    curl.exe -X GET https://cloud.r-project.org/src/contrib/PACKAGES

failed for the reasons described above (certificate revocation check failed).

The proposed solution:

    curl.exe -X GET --ssl-revoke-best-effort  https://cloud.r-project.org/src/contrib/PACKAGES

worked without any issue.

I also considered that a user may have set the `--ssl-revoke-best-effort` parameter manually for renv, and with this patch the user would be passing the parameter twice. Therefore I tested passing that option to curl twice. curl did not care, so no backwards compatibility issue is expected. In other words, this worked:

    curl.exe -X GET --ssl-revoke-best-effort   --ssl-revoke-best-effort  https://cloud.r-project.org/src/contrib/PACKAGES

Thanks for your consideration reviewing this PR and for your time working on renv. It is really useful.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -58,6 +58,8 @@
 
 * Use correct spelling of IRkernel package (#1528).
 
+* Honor `R_LIBCURL_SSL_REVOKE_BEST_EFFORT` when using an external `curl.exe`
+  binary to download files (#1624)
 
 # renv 1.0.0
 


kevinushey (approved):
LGTM -- thank you very much for the PR!

PR #1636, opened 2023-08-10
Explicit snapshot

cstepper:
Hi,

we used to use `renv::snapshot(type = "explicit")` to capture all dependencies stated in the DESCRIPTION file for package development. We put all packages we need for CICD to *Suggests*, as they are not required for a working package API.
During CICD pipelines, we used `renv::restore()` for installing all packages we need.

After the recent update to v1.0.0 we basically ran into the same issue as described here:
https://github.com/rstudio/renv/issues/1019#issuecomment-1644266879

We don't like the idea of putting all CICD required packages into *Imports*. Instead, we propose to expose the option to  record *dev* dependencies in the snapshot (and to consider them in checking the project status with `renv::status()`).

Please have a look. Thanks.

### DIFF
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -91,8 +91,7 @@ the$auto_snapshot_hash <- TRUE
 #'   to the currently active package repositories, as retrieved by
 #'   `getOption("repos")`.
 #'
-#' @param dev Boolean; should development dependencies be captured
-#'   (for type `"explicit"`)?
+#' @inheritParams dependencies
 #'
 #' @param packages A vector of packages to be included in the lockfile. When
 #'   `NULL` (the default), all packages relevant for the type of snapshot being
--- a/R/snapshot.R
+++ b/R/snapshot.R
@@ -124,6 +124,7 @@ snapshot <- function(project  = NULL,
                      lockfile = paths$lockfile(project = project),
                      type     = settings$snapshot.type(project = project),
                      repos    = getOption("repos"),
+                     dev      = FALSE,
                      packages = NULL,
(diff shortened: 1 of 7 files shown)


R/status.R
```diff
@@ -86,6 +86,8 @@ the$status_running <- FALSE
 #'   cache are installed at the expected + proper locations, and validate the
 #'   hashes used for those storage locations.
 #'
+#' @param dev Boolean; should development dependencies be considered?
```
hadley, line 89: You should only need to define this in one place, and then rely on `@inheritParams` to pull it in.

R/snapshot.R
```diff
@@ -91,6 +91,9 @@ the$auto_snapshot_hash <- TRUE
 #'   to the currently active package repositories, as retrieved by
 #'   `getOption("repos")`.
 #'
+#' @param dev Boolean; should development dependencies be captured
```
hadley, line 94: I think it also applies to the type all" Regardless, I'd suggest pointing people to the `?dependencies` for more details.

R/status.R
cstepper: I used `@inheritParams dependencies` for documenting the `dev` argument in both status.R and snapshot.R.
Thanks for the hint

R/snapshot.R
cstepper: I added the link to the `dependencies()` help into the **See Also** section for more details about handling R package dependencies. Feel free to modify if you think this should be more prominent in `shapshot()`.

kevinushey (approved):
LGTM; thanks!

Issue #1617, opened 2023-07-29
Activation considerably slower since 1.0.0 if renv version used in project is not installed globally


lorenzwalthert:
renv version 0.17.3 has a relativley fast `source('renv/activate.R')` (after the env was set up with `restore()` before), even if not the same renv version that is used in a specific project is globally installed. If my project uses renv version 1.0.0, the same command very slow if the same version of renv is not available in the global R package library. For https://github.com/lorenzwalthert/precommit, it took ~ 1s on my labtop if my project used 0.17.3 and around 30s with version 1.0.0 if renv is not installed globally.

**reprex**
```bash
# do this outside RStudio to avoid RENV_PROJECT and other env variables to interfere
# ------------------ setup -------------------- 
git clone https://github.com/lorenzwalthert/precommit
cd precommit
# verify that relevant commit was only renv version change
git show 981cb1b29c5b028906e1a965ce0ab961f3164d0f
git reset 981cb1b29c5b028906e1a965ce0ab961f3164d0f~1 --hard

# -------------  with renv verion 1.0.0 recorded in lockfile -------------- 

Rscript -e "if (rlang::is_installed('renv')) remove.packages('renv')"
Rscript -e "source('renv/activate.R'); renv::restore()"

# after I restored the renv, this is now very slow
Rscript -e "system.time(source('renv/activate.R'))"
Rscript -e "if (rlang::is_installed('renv')) remove.packages('renv')"


# -------------- with renv version 0.17.3 in lockfile  ---------- 
git reset 981cb1b29c5b028906e1a965ce0ab961f3164d0f --hard
Rscript -e "source('renv/activate.R'); renv::restore()"

# after I restored the renv, this is fast
Rscript -e "system.time(source('renv/activate.R'))"
Rscript -e "if (rlang::is_installed('renv')) remove.packages('renv')"
```
I noted that I could disable the sandbox, as a warning message suggest, which is apparently responsible for nearly all the overhead according to the docs and my experiments. It's not clear what side effect disabling the sandbox has for my use case, which is a read-only, non-interactive usage of renv to run an R script as a pre-commit hook (https://pre-commit.com/). Since every hook invokes a shell process which invokes `Rscript` which has to load renv, while the user waits to see all hooks pass after `git commit`, a delay of even a few seconds is already a lot (on top of the slow R Startup). I know this is maybe a bit of special case, but also everytime someone opens RStudio with a renv, a startup time of 20s on top of RStudio + R is not very convenient.

[...]

PR #1640, opened 2023-08-11
watchdog: fix path to renv's Meta/package.rds

kyleam:
renv_watchdog_start_impl() specifies the library so that the subprocess can find renv.  If renv appears to be loaded by devtools, it falls back to setting the library to the result of renv_libpaths_default().

As of f05df6748 (fix bad refactor; tweaks, 2023-08-07), a missing Meta/package.rds is taken as an indication that renv was loaded by devtools.  However, this check incorrectly looks in the _parent_ directory of the renv package.

As a result, f05df6748 resurfaced the problem resolved by 43a3fcdbf (set library path when loading renv in watchdog process (#1618), 2023-08-01).  The launched process fails 1) if renv is not in the renv_libpaths_default() paths or 2) if those paths contain an older renv without the required watchdog functionality.

Point the check to the correct location.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # renv (development version)
 
+* Corrected logic that counteracted the previous release's fix for a watchdog
+  loading issue. (#1640)
 
 # renv 1.0.1
 
--- a/R/watchdog.R
+++ b/R/watchdog.R
@@ -88,9 +88,11 @@ renv_watchdog_start_impl <- function() {
   script <- renv_scope_tempfile("renv-watchdog-", fileext = ".R")
 
   # figure out library path -- need to dodge devtools::load_all()
-  library <- dirname(renv_namespace_path(.packageName))
-  if (!file.exists(file.path(library, "Meta/package.rds")))
-    library <- renv_libpaths_default()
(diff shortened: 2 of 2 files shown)


kevinushey:
Thanks for the PR! I pushed a fix this morning before seeing your PR; it looks like we had the same idea. Can you confirm on your end as well whether my change seems sufficient?

kyleam:
> Can you confirm on your end as well whether my change seems sufficient?

Yes, your code change in abe8125d0 is identical minus choice in variable name ([diff](https://github.com/rstudio/renv/compare/1f00303f1c69bb5966abefae56f24f6879798f2d..abe8125d0d5c01c24f95b32d237aad7cd600512e)).

I'd recommend adding something to the NEWS because, as far as I can tell, f05df674 reintroduced the bug reported in gh-1617 before the 1.0.0 release.

kevinushey:
Thanks, you're right -- I'll update NEWS. I'll also plan to get a patch release of renv to CRAN shortly.

kyleam:
> I'll also plan to get a patch release of renv to CRAN shortly.

Great.  Thanks for the fix and all your work on renv.

PR #1664, opened 2023-08-31
Add closing double quote in a helpfile option

remlapmot:
Thanks for such a great package.

This just adds the missing double quote to the end of the `"custom"` `type` in the `snapshot()` helpfile code. I also reran `devtools::document()`.


kevinushey:
Thank you for catching that!

PR #1716, opened 2023-10-11
package manager doesn't serve rhel9 packages at /centos

edavidaja:


R/ppm.R
```diff
@@ -201,8 +201,9 @@ renv_ppm_platform_rhel <- function(properties) {
   id <- properties$VERSION_ID
   if (is.null(id))
     return(NULL)
+  rhel_version <- ifelse(as.integer(id) < 9, "centos", "rhel")
```
kevinushey, line 204: Nit: we should use `numeric_version()` here instead of `as.integer()` in case the version here happened to be something like `9.10` in some theoretical future.

kevinushey (approved):

R/ppm.R
```diff
@@ -201,8 +201,9 @@ renv_ppm_platform_rhel <- function(properties) {
   id <- properties$VERSION_ID
   if (is.null(id))
     return(NULL)
+  rhel_version <- ifelse(numeric_version(id) < 9, "centos", "rhel")
```
kevinushey, line 204:
```suggestion
  rhel_version <- ifelse(numeric_version(id) < "9", "centos", "rhel")
```

kevinushey (approved):

kevinushey:
Thanks!

PR #1728, opened 2023-10-17
Expand PPM platform support

### LINKED ISSUE #1720: `renv_ppm_platform()` should detect Rocky Linux as Red Hat.
Rocky Linux is "designed to be 100% bug-for-bug compatible with Red Hat Enterprise Linux". However, `renv` (via `renv_ppm_platform()`) does not detect it as RHEL-compatible, so `renv_ppm_transform()` transforms Package Manager URLs (even those using compatible RHEL9 binaries) to source package URLs.

You can currently work around this by setting the `RENV_PPM_PLATFORM` environment variable, but it would be better if it worked out of the box.

toph-allen:
Fixes #1720 

I set out to add support for getting Rocky Linux 9 binaries from Package Manager instances, as it's binary compatible with CentOS/RHEL binaries. This was initially motivated by adding binary support for Connect instances hosted on Rocky. Along the way I went to check out `/etc/os-release` files from other supported distros, and made some other additions/fixes.

Connect's supported distros are listed [here](https://docs.posit.co/connect/admin/), and Package Manager's are available [here](https://packagemanager.posit.co/client/#/repos/cran/setup).

Changes I made:

- Added functions to detect Rocky Linux and AlmaLinux, two CentOS/RHEL-compatible distributions, as `centos8`/`rhel9`.
- Added support for Debian. (It isn't supported by Connect but is by PPM, so it's nice for `renv` to support it.) Its `platform` function is like Ubuntu.
- Fixed support for OpenSUSE. Previously, the `renv_ppm_platform_suse()` function would return something like `opensuse15`, but the minor part of the version needs to be appended, e.g. `opensuse154`. Also, OpenSUSE `/etc/os-release` files don't always contain `suse`  as a standalone word, so the I changed the regex used to identify this distro.
- Added support for SLES, which has different `/etc/os-release` files from OpenSUSE.
- Added tests for a cross-section distros (not all minor versions).

Other changes I'm considering, and would appreciate feedback on whether to include or not:

- Adding support for detecting Amazon Linux 2 as `centos7`. It seems to be mostly compatible ([StackOverflow](https://serverfault.com/questions/798427/what-linux-distribution-is-the-amazon-linux-ami-based-on), [Amazon docs](https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html#epel)), and I think this is the basis of Connect listing it as supported, but I'm not sure.

Code style questions:

- It looks like we previously erred on the side of giving each distro a separate function, even if for now the logic is identical to other functions. I followed this pattern with the new distros. But we could also just call e.g. `renv_ppm_platform_rhel` for AlmaLinux and Rocky.
- We could look for e.g. `\\bsuse\\b` in the `ID_LIKE` field instead of directly looking for the `ID` field.

### DIFF
--- a/R/ppm.R
+++ b/R/ppm.R
@@ -171,7 +171,6 @@ renv_ppm_platform_impl <- function(file = "/etc/os-release") {
       identical(id, "rhel")      ~ renv_ppm_platform_rhel(properties),
       identical(id, "rocky")     ~ renv_ppm_platform_rocky(properties),
       identical(id, "almalinux") ~ renv_ppm_platform_alma(properties),
-      # TODO: Could use `ID_LIKE` and look for \\bsuse\\b here, and get rid of the separate suse and sles calls?
       grepl("suse\\b", id)       ~ renv_ppm_platform_suse(properties),
       identical(id, "sles")      ~ renv_ppm_platform_sles(properties),
       identical(id, "debian")    ~ renv_ppm_platform_debian(properties)
--- a/R/ppm.R
+++ b/R/ppm.R
@@ -166,10 +166,15 @@ renv_ppm_platform_impl <- function(file = "/etc/os-release") {
     id <- properties$ID %||% ""
 
     case(
-      identical(id, "ubuntu") ~ renv_ppm_platform_ubuntu(properties),
-      identical(id, "centos") ~ renv_ppm_platform_centos(properties),
-      identical(id, "rhel")   ~ renv_ppm_platform_rhel(properties),
-      grepl("\\bsuse\\b", id) ~ renv_ppm_platform_suse(properties)
+      identical(id, "ubuntu")    ~ renv_ppm_platform_ubuntu(properties),
+      identical(id, "centos")    ~ renv_ppm_platform_centos(properties),
+      identical(id, "rhel")      ~ renv_ppm_platform_rhel(properties),
+      identical(id, "rocky")     ~ renv_ppm_platform_rocky(properties),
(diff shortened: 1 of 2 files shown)


aronatkins (approved):
Here is the `/etc/os-issue` from an amazonlinux2 container

```bash
docker run -it --rm amazonlinux:2 cat /etc/os-release
```

```bash
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
SUPPORT_END="2025-06-30"
```

aronatkins:
Does this change (and the one in https://github.com/rstudio/renv/pull/1716) need NEWS?

kevinushey (approved):
LGTM!

kevinushey:
> It looks like we previously erred on the side of giving each distro a separate function, even if for now the logic is identical to other functions. I followed this pattern with the new distros. But we could also just call e.g. renv_ppm_platform_rhel for AlmaLinux and Rocky.

I'd prefer keeping these separate just for clarity, even if it means some code is duplicated here.

toph-allen:
> I'd prefer keeping these separate just for clarity, even if it means some code is duplicated here.

Yeah, agreed 👍🏻

PR #1738, opened 2023-10-24
Updated workflow projects link.

Dulani:
When you click on the Workflow: projects link, there is a notice at the top of the page that directs the reader to the second edition.  So, I've updated the link in this vignette to point the reader directly to the second edition.


kevinushey:
Thanks for the PR -- sorry for taking so long to review and merge.

PR #1851, opened 2024-03-13
Add `remove_python` function

warnes:
Add `remove_python` function to disable Python dependency tracking and delele associated directories/files.


kevinushey:
Thanks, and sorry for letting this sit for so long. I'm going to try an alternate mechanism for this.

warnes:
Is there any particular reason for not accepting this PR?

kevinushey:
I want to keep the API surface in renv small, and given that we already had similar behavior available via `use_python(FALSE)`, my preference was to document that behavior, and then notify if other Python-related infrastructure files still existed so the user could manually clean those up if necessary.

PR #1862, opened 2024-03-22
for `settings$ppm.enabled()`, update should be NULL like other settings without config

cderv:
Otherwise this creates an issue in `renv_settings_updated` as it will use a wrong update function

https://github.com/rstudio/renv/blob/5bc089696a96ca372f1f1ecaf0a0a4167891ed36/R/settings.R#L180-L183


kevinushey (approved):
Thanks!

Issue #1877, opened 2024-04-16
Addition of `renv.lock` schema


jrdnbradford:
I see there's been some talk about having a `renv.lock` schema (https://github.com/rstudio/renv/issues/1107). I put together a minimal version of this schema and opened a PR over at https://github.com/lorenzwalthert/precommit/pull/558 that would use [`pre-commit`](https://pre-commit.com/) to validate `renv.lock` files against the schema.

If there's still interest, it may be a better idea to fill out the schema more and put in a PR _here_ instead, just let me know where it should live. Then it could be maintained here, perhaps with a new `renv::validate_lockfile()` function, and we could add a hook at https://github.com/lorenzwalthert/precommit that utilizes the functionality native to {`renv`}

Or I could throw the schema over at https://github.com/SchemaStore/schemastore instead. In any case, I'm happy to do the work and put in a PR with the schema and anything else. Let me know.

jrdnbradford:
To my knowledge, generally in situations like this practitioners avoid creating a version of a schema for each new version of what they’re validating and instead just expand the schema. JSON schema has the flexibility to set fields such as `deprecated` to allow this sort of expansion as the lockfile format changes. There are also ways to conditionally validate, validate against multiple patterns, etc, so we can use multiple ways to validate changing fields if necessary. This isn't to say we should do this for `renv`, but given it's general stability it may be the best way to go.

JSON schema also has a `required` field to define what is absolutely necessary, and an `additionalProperties` field to define whether other fields not specified in the schema are allowed. Both are available at each block level. This would allow toleration of user adornments.

Here's my current thoughts: we define a schema that validates all the fields `renv` knows about. We allow `additionalProperties` and set the `required` fields to only the absolute minimum. This might not sound stringent enough, but you still get 99% of the validation with _far_ fewer false negatives and much less maintenance. 

For those who want more stringent validation, the function we write to validate can have a “bring your own schema” parameter, so to speak. Users can validate against `renv`’s schema, or else use it as a starting point for their own lockfile schema:

```R
lockfile_validate <- function(
    lockfile = NULL, # Use default project lockfile if not provided
    lockfile_schema  = NULL, # Use default schema in inst if not provided
    # Other parameters
)
```
This would make sense for those who have very specific production requirements for their validation. 

Let me know what you think. I'll put in a PR with what I'm thinking this week so it can be discussed and iterated over. It will include a schema, an exported validation function, and tests.

kevinushey:
[...]

PR #1889, opened 2024-05-01
Add JSON lockfile schema and validation function

jrdnbradford:
Here's my first go at issue #1877.

## Why Use [JSON Schema](https://json-schema.org/):

- Provides an easy, declarative way to document `renv.lock`
- Integrates with text editors to provide autocompletion, hover help, and validation
- Allows automated validation in CI, builds, tests, etc for R applications

## General Decisions

- A relatively relaxed schema to prevent false negatives, with the option for users to pass their own schema that fits their personal requirements. Additional fields are allowed.
- Heavy on positive and negative `testthat` tests, so that we can verify updates to the schema
- The function wraps [`jsonvalidate::json_validate()`](https://docs.ropensci.org/jsonvalidate/) and takes/passes many of its parameters there. I have not implemented all of them, but we can add if you’d like.
- Ease of use. Calling without parameters should cover most cases: `renv::lockfile_validate()`

## Questions

- Do I have all the fields `renv` knows about? 
- Do I have `required` set _only_ for the fields that are absolutely necessary?
- I’ve used `description`s from `renv` documentation whenever possible, but had to make some inferences. I myself have never used anything that added `Bioconductor`, `Python`, or a few other fields. These should be checked.


R/lockfile-validate.R
```diff
@@ -0,0 +1,75 @@
+
+#' Validate the renv lockfile against a schema
```
kevinushey, line 2:
```suggestion
#' Validate an renv lockfile against a schema
```

R/lockfile-validate.R
```diff
@@ -0,0 +1,75 @@
...
+
+  project <- renv_project_resolve(project)
+  lockfile <- lockfile %||% renv_lockfile_path(project = project)
+  schema <- schema %||% system.file("schema",
+                                    "draft-07.renv.lock.schema.json",
+                                    package = "renv",
+                                    mustWork = TRUE)
+
+  # "ajv" engine required for schema specifications later than draft-04
+  jsonvalidate::json_validate(lockfile,
```
kevinushey, line 68: Could we use either add `jsonvalidate` to the Suggests file of the package DESCRIPTION file, or do something like https://github.com/rstudio/renv/blob/5d0d52c395e569f7f24df4288d949cef95efca4e/R/dependencies.R#L1664-L1683 and better handle the presence (or absence) of the `jsonvalidate` package?

kevinushey (approved):
Overall LGTM -- thank you!

R/lockfile-validate.R
jrdnbradford: Added to the `DESCRIPTION`'s `Suggests`.

jrdnbradford:
Updates have been pushed. 

Presuming this gets merged, feel free to @ me on issues, bugs, etc on this. Schemas and their validation is always a balancing act, and the R community at large may need different tradeoffs. 

Closes #1877.

kevinushey:
A huge thanks again for putting this together!

PR #1915, opened 2024-05-31
Refine auto-loading behaviour

jennybc:
* Tweak the use of RStudio's session init hook to deliver the intent of PR #1652.
* Otherwise, if RStudio and its session init hook aren't available, just emit a message if we discover the need for `renv::restore()` during auto-loading.

To do/discuss:

* [x] ~After "Installing cli ..." it's easy to think the session is busy or hung. But it's not. If you hit return, you get the normal R prompt. Why does that happen?~ *As I play around with renv, I run into this phenomenon elsewhere. It feels like one or more of the messaging helpers, such as `catf()` or `writef()` sometimes fail to get their trailing newline? In any case, I think it's unrelated to what I'm doing here and out of scope for this PR.*
* [x] The "Operation canceled." is unsightly and confusing. Seems like that should be suppressed or, at least in this context, hidden.
* [x] Try this on: Never try to interact with the user when auto-loading, since user interaction is UB during R startup. If we can defer loading using a hook (like `rstudio.sessionInit` below), do that. If we can't, emit a message telling the user what to do. ~If possible, style the `renv::restore()` part as a runnable link.~ *I had to give up on that part.*

### DIFF
--- a/R/load.R
+++ b/R/load.R
@@ -885,13 +885,7 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
     caution("- None of the packages recorded in the lockfile are currently installed.")
     autoloading <- getOption("renv.autoloader.running", default = FALSE)
     if (autoloading) {
-      if (requireNamespace("cli", quietly = TRUE)) {
-        cli::cli_bullets(c(
-          "*" = "Use {.run renv::restore()} to restore the project library."
-        ))
-      } else {
-        caution("- Use `renv::restore()` to restore the project library.")
-      }
+      caution("- Use `renv::restore()` to restore the project library.")
       return(FALSE)
     }
 
--- a/R/load.R
+++ b/R/load.R
@@ -146,11 +146,17 @@ renv_load_action <- function(project) {
   if (!interactive())
     return("load")
 
-  # if we're running within RStudio at this point, and we're running
-  # within the auto-loader, we need to defer execution here so that
-  # the console is able to properly receive user input and update
-  # https://github.com/rstudio/renv/issues/1650
   autoloading <- getOption("renv.autoloader.running", default = FALSE)
+  # if we're auto-loading, it's too early to interact with the user, which is
+  # often advisable, i.e. if we detect that the user needs to run renv::restore()
+  # https://github.com/rstudio/renv/issues/1650
+  #
+  # if the frontend is known to support a session init hook, defer loading until
+  # R is fully initialized, at which time it will be possible to interact with
+  # the user (currently this just applies to RStudio)
+  #
+  # otherwise, proceed with the knowledge that, if the user needs to run
+  # renv::restore(), a message to that effect will be emitted
   if (autoloading && renv_rstudio_available()) {
     setHook("rstudio.sessionInit", function(...) { renv::load(project) })
     return("cancel")
@@ -876,7 +882,8 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
   if (length(intersect(lockpkgs, libpkgs)) == 0 && length(lockpkgs) > 0L) {
 
     caution("- None of the packages recorded in the lockfile are currently installed.")
-    if (renv_rstudio_autoloading()) {
+    autoloading <- getOption("renv.autoloader.running", default = FALSE)
+    if (autoloading) {
       caution("- Use `renv::restore()` to restore the project library.")
       return(FALSE)
     }
--- a/R/load.R
+++ b/R/load.R
@@ -884,7 +884,13 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
     caution("- None of the packages recorded in the lockfile are currently installed.")
     autoloading <- getOption("renv.autoloader.running", default = FALSE)
     if (autoloading) {
-      caution("- Use `renv::restore()` to restore the project library.")
+      if (requireNamespace("cli", quietly = TRUE)) {
+        cli::cli_bullets(c(
+          "*" = "Use {.run renv::restore()} to restore the project library."
+        ))
+      } else {
+        caution("- Use `renv::restore()` to restore the project library.")
+      }
       return(FALSE)
     }
 
--- a/R/load.R
+++ b/R/load.R
@@ -70,7 +70,7 @@ load <- function(project = NULL, quiet = FALSE, profile = NULL, ...) {
 
   action <- renv_load_action(project)
   if (action[[1L]] == "cancel") {
-    cancel()
+    quietly(cancel())
   } else if (action[[1L]] == "init") {
     return(init(project))
   } else if (action[[1L]] == "alt") {
--- a/R/load.R
+++ b/R/load.R
@@ -70,7 +70,8 @@ load <- function(project = NULL, quiet = FALSE, profile = NULL, ...) {
 
   action <- renv_load_action(project)
   if (action[[1L]] == "cancel") {
-    quietly(cancel())
+    autoloading <- getOption("renv.autoloader.running", default = FALSE)
+    cancel(verbose = !autoloading)
   } else if (action[[1L]] == "init") {
     return(init(project))
   } else if (action[[1L]] == "alt") {
--- a/R/load.R
+++ b/R/load.R
@@ -881,7 +881,7 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
       return(FALSE)
     }
 
-    response <- ask("Would you like to restore the project library?", default = FALSE)
+    response <- ask("Would you like to run `renv::restore()` to restore the project library?", default = FALSE)
     if (!response)
       return(FALSE)
 
--- a/R/load.R
+++ b/R/load.R
@@ -146,12 +146,6 @@ renv_load_action <- function(project) {
   if (!interactive())
     return("load")
 
-  # if this project already contains an 'renv' folder, assume it's
-  # already been initialized and we can directly load it
-  renv <- renv_paths_renv(project = project, profile = FALSE)
-  if (dir.exists(renv))
-    return("load")
-
   # if we're running within RStudio at this point, and we're running
   # within the auto-loader, we need to defer execution here so that
   # the console is able to properly receive user input and update
@@ -161,6 +155,12 @@ renv_load_action <- function(project) {
     setHook("rstudio.sessionInit", function() { renv::load(project) })
     return("cancel")
   }
+
+  # if this project already contains an 'renv' folder, assume it's
+  # already been initialized and we can directly load it
+  renv <- renv_paths_renv(project = project, profile = FALSE)
+  if (dir.exists(renv))
+    return("load")
 
   # check and see if we're being called within a sub-directory
   path <- renv_file_find(dirname(project), function(parent) {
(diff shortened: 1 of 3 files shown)


jennybc:
Before:

*I'll fill this in later.*

After:

```
- Operation canceled.
- Project '~/rrr/renv-coldstart' loaded. [renv 1.0.7.9000]
- None of the packages recorded in the lockfile are currently installed.
Would you like to run `renv::restore()` to restore the project library? [y/N]: y

The following package(s) will be updated:

# CRAN -----------------------------------------------------------------------
- cli   [* -> 3.6.2]

- Installing cli ...                            OK [linked from cache]
```

R/load.R
```diff
@@ -146,22 +146,22 @@ renv_load_action <- function(project) {
   if (!interactive())
     return("load")
 
-  # if this project already contains an 'renv' folder, assume it's
-  # already been initialized and we can directly load it
-  renv <- renv_paths_renv(project = project, profile = FALSE)
-  if (dir.exists(renv))
-    return("load")
-
```
jennybc, line 154: I think this check should ?probably? come after the one that checks for "are we autoloading"?

I also wonder if the condition should gain another check: Is there a project-specific library? Because if you `git clone` someone else's renv project, the `renv/` directory will be there but you can't really say that the project has been initialized. It's possible I'm wrong about this, because I'm still a bit foggy on the different states and verbs.

I can also imagine that these two matters interact with each other. I.e. if the condition gets modified, does that change the need to reorder these?

R/load.R
```diff
@@ -881,7 +881,7 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
       return(FALSE)
     }
 
-    response <- ask("Would you like to restore the project library?", default = FALSE)
+    response <- ask("Would you like to run `renv::restore()` to restore the project library?", default = FALSE)
```
jennybc, line 884: It feels like this could ease decision-making for the user, i.e. it's easier to answer this question confidently if you know exactly what's being proposed. Also might have some value in terms of user education.

R/load.R
```diff
@@ -146,22 +146,22 @@ renv_load_action <- function(project) {
...
-    return("load")
-
   # if we're running within RStudio at this point, and we're running
   # within the auto-loader, we need to defer execution here so that
   # the console is able to properly receive user input and update
   # https://github.com/rstudio/renv/issues/1650
   autoloading <- getOption("renv.autoloader.running", default = FALSE)
   if (autoloading && renv_rstudio_available()) {
-    setHook("rstudio.sessionInit", function() { renv::load(project) })
+    setHook("rstudio.sessionInit", function(...) { renv::load(project) })
```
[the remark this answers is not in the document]
jennybc: I think this is the explanation:

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/modules/SessionRHooks.R#L26-L29

RStudio forwards arguments and the whole thing happens inside `tryCatch()`. So I guess if the function definition include no arguments, the attempt to run the hook fails. Silently.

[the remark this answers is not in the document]
jennybc: The whole chain is:

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/modules/SessionRHooks.hpp#L19

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/SessionMain.cpp#L857

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/modules/SessionRHooks.hpp#L36-L39/./

https://github.com/rstudio/rstudio/blob/8ebe5431ca2fb3c9719b70d4fbc45201d226c79c/src/cpp/session/modules/SessionRHooks.R#L16-L31

R/load.R
```diff
@@ -70,7 +70,7 @@ load <- function(project = NULL, quiet = FALSE, profile = NULL, ...) {
 
   action <- renv_load_action(project)
   if (action[[1L]] == "cancel") {
-    cancel()
+    quietly(cancel())
```
jennybc, line 73: Without this (but with the other changes in this PR), the user sees a potentially confusing message about cancelling an operation (what operation?) at startup:

```
- Operation canceled.
- Project '~/rrr/renv-coldstart' loaded. [renv 1.0.7.9000]
```

This is my best attempt to suppress this, using what I see here re: suppressing messages. I'd rather put the `quietly()` around `return("cancel")` below, but that doesn't actually suppress the message.

R/load.R
```diff
@@ -146,22 +146,22 @@ renv_load_action <- function(project) {
...
-
   # if we're running within RStudio at this point, and we're running
   # within the auto-loader, we need to defer execution here so that
   # the console is able to properly receive user input and update
   # https://github.com/rstudio/renv/issues/1650
   autoloading <- getOption("renv.autoloader.running", default = FALSE)
   if (autoloading && renv_rstudio_available()) {
-    setHook("rstudio.sessionInit", function() { renv::load(project) })
+    setHook("rstudio.sessionInit", function(...) { renv::load(project) })
     return("cancel")
```
jennybc, line 156: Morally, I want to do `quietly(return("cancel"))`, but that doesn't actually work.

R/load.R
kevinushey: I think this mostly comes down to whether the cancel was performed interactively (the user actually asked to cancel) versus us non-interactively cancelling.

Maybe we can detect if the autoloader is running at that point, and do something like `cancel(quiet = TRUE)`, and update the definition of `cancel` appropriately?

R/load.R
```diff
@@ -876,12 +882,19 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
...
-      caution("- Use `renv::restore()` to restore the project library.")
+    autoloading <- getOption("renv.autoloader.running", default = FALSE)
+    if (autoloading) {
+      if (requireNamespace("cli", quietly = TRUE)) {
+        cli::cli_bullets(c(
+          "*" = "Use {.run renv::restore()} to restore the project library."
+        ))
+      } else {
+        caution("- Use `renv::restore()` to restore the project library.")
+      }
```
jennybc, line 893: Main change: Switch condition from `renv_rstudio_autoloading()` to just checking if we're autoloading. Because I think we're all in agreement that we should only message in this case, regardless of the frontend.

The second change, which I think we'll have to abandon, is kind of interesting and worth discussing first: try to emit the advice to run `renv::restore()` as a clickable hyperlink.

It turns out you can never hit the `cli::cli_bullet()` code above. Why not? Because the check for cli happens with respect to the current libpaths, meaning with respect to the current renv project.

And if cli is there then it's impossible that `length(intersect(lockpkgs, libpkgs)) == 0`.

So, since there's no ability to use packages for the UI that aren't present in the project-specific library, this particular effort is pointless.

There are, no doubt, other places where we *could* actually emit runnable hyperlinks, but I think the cli-using smarts would probably be better baked into `caution()` and friends.

R/rstudio.R
```diff
@@ -10,10 +10,6 @@ renv_rstudio_available <- function() {
 
 }
 
-renv_rstudio_autoloading <- function() {
-  renv_rstudio_available() && getOption("renv.autoloader.running", default = FALSE)
-}
-
```
jennybc, line 16: I removed the sole usage of this function.

R/utils.R
```diff
@@ -126,13 +126,15 @@ ask <- function(question, default = FALSE) {
...
+  # code. Attempting to do so can crash the R process."
   # https://github.com/rstudio/renv/issues/1879
-  if (renv_rstudio_available()) {
-    autoloading <- getOption("renv.autoloader.running", default = FALSE)
-    if (autoloading)
-      return(default)
-  }
+  autoloading <- getOption("renv.autoloader.running", default = FALSE)
+  if (autoloading)
+    return(default)
```
jennybc, line 137: More simplification now that we've decided to never attempt interaction when autoloading, regardless of the frontend.

R/load.R
jennybc: OK good to know it's OK to propose a change in `cancel()`. I'll work on adding that to this PR.

R/load.R
```diff
@@ -876,12 +882,19 @@ renv_load_report_synchronized <- function(project = NULL, lockfile = NULL) {
   if (length(intersect(lockpkgs, libpkgs)) == 0 && length(lockpkgs) > 0L) {
 
     caution("- None of the packages recorded in the lockfile are currently installed.")
-    if (renv_rstudio_autoloading()) {
-      caution("- Use `renv::restore()` to restore the project library.")
+    autoloading <- getOption("renv.autoloader.running", default = FALSE)
+    if (autoloading) {
+      if (requireNamespace("cli", quietly = TRUE)) {
+        cli::cli_bullets(c(
```
kevinushey, line 888: `renv` intentionally doesn't try to load any packages (or use any custom packages), as we don't want to cause issues for users on Windows who might want to install or update these packages.

Can we use own our `renv` shim here? Could this just be `caution()`? How hard would it be for us to implement support for hyperlinks ourselves in `renv`?

R/load.R
```diff
@@ -70,7 +70,8 @@ load <- function(project = NULL, quiet = FALSE, profile = NULL, ...) {
 
   action <- renv_load_action(project)
   if (action[[1L]] == "cancel") {
-    cancel()
+    autoloading <- getOption("renv.autoloader.running", default = FALSE)
+    cancel(verbose = !autoloading)
```
jennybc, line 74: Allows us to see the message after overt cancelation by the user, but suppress it when deferring the project load to the session init hook. As discussed elsewhere.

R/load.R
jennybc: Handles by adding a `verbose` argument to `cancel()`.

R/load.R
jennybc: I'm going to remove this whole attempt to provide a clickable (as in, runnable) hyperlink. There's the point you make but also the fact that it's logically impossible to hit this code (explained elsewhere).

This effort made me appreciate that you really can't use other packages to create the UI for renv itself.

R/load.R
jennybc: Relocated.

jennybc:
This is ready now.

I took a look at adding a test but TBH it doesn't look very approachabel. I think this behaviour is legitimately hard to test (interactive usage, with and without RStudio), but I'm also not familiar with all of renv's test helpers, which are extensive. So either this can go forward without an explicit test or else I could use some guidance on how to approach that.

jennybc:
If you're interested in getting clickable hyperlinks w/o taking a true cli dependency (I understand why that's a nonstarter), this would be one way:

https://github.com/r-lib/rlang/blob/main/R/standalone-cli.R

I could open a separate issue for that.

kevinushey (approved):
LGTM!

kevinushey:
Do we need a NEWS update? (I'm going to merge this now for testing but we can merge a separate NEWS update after if appropriate)

kevinushey:
> If you're interested in getting clickable hyperlinks w/o taking a true cli dependency (I understand why that's a nonstarter), this would be one way:
> 
> https://github.com/r-lib/rlang/blob/main/R/standalone-cli.R
> 
> I could open a separate issue for that.

I'd definitely be on board for this.

PR #1950, opened 2024-07-25
Tweak `lockfile_from_manifest()` defaults/docs

hadley:


R/manifest-convert.R
```diff
@@ -20,18 +21,11 @@
 #'   object; otherwise, the lockfile will be written to the path specified by
 #'   `lockfile`.
 #'
-#' @details
```
hadley, line 23: This appeared to say the same thing as the parameter documentation, so I deleted it.

I do wonder if `renv.lock` should be the default, so you can then call `renv::restore()` to get the packages you need.

R/manifest-convert.R
```diff
@@ -20,18 +21,11 @@
...
-#' that can be used to create a new `renv.lock` file. If `lockfile` is set to a
-#' character string, a new file will be created with that path -- e.g.
-#' `renv.lock` -- and the lockfile object will be returned.
-#'
 #' @return
 #' An renv lockfile.
 #'
 #' @keywords internal
-renv_lockfile_from_manifest <- function(manifest,
+renv_lockfile_from_manifest <- function(manifest = "manifest.json",
```
hadley, line 28: IMO this is going to be 90%+ of use cases, so might as well provide it as the default.

Should I also add an explicit check that this file exists in this function?

kevinushey (approved):
LGTM!

PR #1980, opened 2024-09-05
when attempting to load a lockfile and finding only a manifest, automatically generate a lockfile and load that

toph-allen:
This pull request changes the behavior of `renv_lockfile_load()`, which attempts to load a lockfile from a directory. When no lockfile is found, but a `manifest.json` is found, the manifest is used to automatically generate a lockfile, which is then loaded. The behavior only occurs when the function is called with `strict = FALSE`, the default. If no manifest is present, the function `renv_lockfile_init()` is called, as before. A message is printed when this happens.

I added a supporting function in `paths.R` to return a manifest path, just cribbing from `renv_paths_lockfile()`. However, just doing `file.path(project, "manifest.json")` might be just as good. I'm not sure what the semantics are around path loading, so was just trying to follow existing patterns.

The logic in `renv_manifest_path()` is similar to `renv_paths_lockfile()`, or at least, the default case. I'm unsure about needing to add an extra layer of abstraction similar to `renv_lockfile_path()` -> `renv_paths_lockfile()`.

When testing, I saw a duplicated log line:

```
No lockfile found; creating from `manifest.json`.
- Lockfile written to "/private/var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gp/T/RtmpsmRCq6/file62e9cace1/renv.lock".
- Lockfile written to "/private/var/folders/q7/1b7h_nx931q6mc70r3w8k_2c0000gp/T/RtmpsmRCq6/file62e9cace1/renv.lock".
```

This came from explicit logging in `renv_lockfile_from_manifest()`, which duplicated a log message within `renv_lockfile_write()`, so I removed the duplicated outer logging statement.

### DIFF
--- a/R/lockfile.R
+++ b/R/lockfile.R
@@ -122,16 +122,22 @@ renv_lockfile_save <- function(lockfile, project) {
 }
 
 renv_lockfile_load <- function(project, strict = FALSE) {
-
-  path <- renv_lockfile_path(project)
-  if (file.exists(path))
-    return(renv_lockfile_read(path))
+  lockfile_path <- renv_lockfile_path(project)
+  if (file.exists(lockfile_path))
+    return(renv_lockfile_read(lockfile_path))
 
   if (strict) {
     abort(c(
       "This project does not contain a lockfile.",
       i = "Have you called `snapshot()` yet?"
     ))
+  }
+
+  manifest_path <- renv_manifest_path(project)
+  if (file.exists(manifest_path)) {
+    caution("No lockfile found; creating from `manifest.json`.")
+    renv_lockfile_from_manifest(manifest_path, lockfile_path)
+    return(renv_lockfile_read(lockfile_path))
   }
 
   renv_lockfile_init(project = project)
--- a/R/lockfile.R
+++ b/R/lockfile.R
@@ -116,15 +116,6 @@ renv_lockfile_path <- function(project) {
   renv_paths_lockfile(project = project)
 }
 
-renv_manifest_path <- function(project = NULL) {
-
-  # similar to default case of `renv_paths_lockfile()`
-  project <- renv_project_resolve(project)
-  renv <- renv_paths_renv(project = project)
-  file.path(dirname(renv), "manifest.json")
-
-}
-
 renv_lockfile_save <- function(lockfile, project) {
   file <- renv_lockfile_path(project)
   renv_lockfile_write(lockfile, file = file)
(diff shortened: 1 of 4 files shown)


R/lockfile.R
```diff
@@ -116,16 +116,24 @@ renv_lockfile_path <- function(project) {
   renv_paths_lockfile(project = project)
 }
 
+renv_manifest_path <- function(project = NULL) {
+
+  # similar to default case of `renv_paths_lockfile()`
+  project <- renv_project_resolve(project)
+  renv <- renv_paths_renv(project = project)
+  file.path(dirname(renv), "manifest.json")
```
toph-allen, line 124: Was unsure of the exact semantics here; I just followed the `renv_paths_lockfile()` function.

kevinushey: I believe you just want `file.path(project, "manifest.json")` here? If that's the case, you can just remove this function too.

R/lockfile.R
```diff
@@ -116,16 +116,24 @@ renv_lockfile_path <- function(project) {
...
 }
 
 renv_lockfile_load <- function(project, strict = FALSE) {
-
-  path <- renv_lockfile_path(project)
-  if (file.exists(path))
-    return(renv_lockfile_read(path))
+  lockfile_path <- renv_lockfile_path(project)
+  if (file.exists(lockfile_path))
+    return(renv_lockfile_read(lockfile_path))
```
kevinushey, line 136:
```suggestion
  path <- renv_lockfile_path(project)
  if (file.exists(path))
    return(renv_lockfile_read(path))
```

R/lockfile.R
```diff
@@ -134,6 +142,13 @@ renv_lockfile_load <- function(project, strict = FALSE) {
     ))
   }
 
+  manifest_path <- renv_manifest_path(project)
+  if (file.exists(manifest_path)) {
+    caution("No lockfile found; creating from `manifest.json`.")
+    renv_lockfile_from_manifest(manifest_path, lockfile_path)
+    return(renv_lockfile_read(lockfile_path))
+  }
+
```
kevinushey, line 151:
```suggestion
  manifest <- renv_manifest_path(project)
  if (file.exists(manifest)) {
    caution("No lockfile found; creating from `manifest.json`.")
    renv_lockfile_from_manifest(manifest, path)
    return(renv_lockfile_read(path))
  }

```

kevinushey (commented):
Overall LGTM; main nits are around style + suggesting we don't need the extra function for finding the path to manifest.json

R/lockfile.R
toph-allen: Yeah, that really is all I want. :)

toph-allen: Yeah, that really is all I want. :) I wasn't sure if `renv_resolve_project()` would account for anything where the actual dir needed for `project` was different from the string passed in (I assumed that was why it was called in the codepath that resolves the lockfile).

I'll remove the function :)

kevinushey (approved):
LGTM!

kevinushey:
I'll take care of it, thanks!

PR #2024, opened 2024-10-24
[CI] align GH and GL paths

fh-mthomson:
`RENV_PATHS_ROOT` is already used by GitHub; I've been using this config accordingly without issue.


kevinushey:
Thanks!

PR #2032, opened 2024-11-08
update embed docs to include qmd

StatsRhian:
`embed()` supports `.qmd` files as of https://github.com/rstudio/renv/commit/884569687ed75efa22a4f5b9fd5ab6f75aa9a7c3

However, `.qmd` format is not mentioned in the docs. 

I've not bumped / updated NEWS as you might want to combine with larger changes.


kevinushey (approved):
Thanks!

PR #2042, opened 2024-11-22
Pass project to renv_bootstrap_run reloading with the correct project

mtheiss:
If renv was bootsraped with a differen RENV_PROJECT environment, the reload would ignore the setting and create renv/ directory in the working directory instead of the project directory.


kevinushey (approved):
LGTM; thanks!

PR #2064, opened 2024-12-18
Fix `strict` schema validation test

jrdnbradford:
This test was passing for the wrong reason: the lockfile was invalid. It now tests that setting `strict = TRUE` fails correctly and `strict = FALSE` succeeds correctly given the custom `schema`.


kevinushey:
Thanks!

PR #2065, opened 2024-12-18
Fix R-CMD-check badge

jrdnbradford:
This fixes the `R-CMD-check.yaml` badge. The `README` currently has an outdated link and is reporting as failed even though the check is passing.


kevinushey (approved):
Thanks!

PR #2066, opened 2024-12-23
Add `lockfile_validate` to reference

### LINKED ISSUE #2048: `lockfile_validate` `pkgdown` site docs
I was looking over the [renv site](https://rstudio.github.io/renv/index.html) and realized `lockfile_validate` doesn't show up explicitly in the Reference.

The function has a decent amount of documentation associated with, so if we just add `#' @rdname lockfiles` to it, it gets merged with the general lockfiles docs in some unexpected ways. 

Here are two ways of handling this:

## Update pkgdown Config
Add its own reference in the pkgdown config:
```yaml
- title: Lockfile Management
  contents:
  - lockfiles
  - lockfile_validate
```
This will maintain its own documentation.

## Move Docs to `lockfiles.R`
We can add a Validation docs header to the docs in `lockfiles.R` and move the `lockfile_validate` docs under that, add `#' @rdname lockfiles` to the function. This way everything related to lockfile management is in one place. But `lockfile_validate` has `@details`, `@examples`, etc that would not be rendered in the same way there.

Any other ways this can be done? I'm happy to fix and open a PR for whatever you think is best here.

jrdnbradford:
Closes #2048.

This seems like the best way to add this function to the reference since the docs need to be filled out a bit more than functions under `@rdname lockfiles`. Happy to change it up if you disagree.

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -45,6 +45,7 @@ reference:
 - title: Lockfile Management
   contents:
   - lockfiles
+  - lockfile_validate
   - record
   - remote
   - modify


kevinushey:
Thanks!

PR #2079, opened 2025-01-24
Update reference to renamed vignette

fkohrt:
The vignette `cellar` has been renamed to `package-sources` as part of ba4f0c0. This commit updates the reference to the vignette.


kevinushey:
Thanks for catching that!

PR #2059, opened 2024-12-12
include all required fields for hashing in lockfile

kevinushey:
Closes https://github.com/rstudio/renv/issues/2057.

Should we still include a Hash field in the lockfile, or should we recompute hashes on the fly? This mainly comes to mind as the hash would be invalid if the lockfile's fields were changed by a user.

I also wonder if there are any DESCRIPTION fields that we should exclude from the lockfile -- for example, fields like Packaged, Built, and Date/Publication, since those will differ depending on where and how the package was built. (A common request for users was to make sure that packages installed from CRAN, and packages installed from P3M, produced otherwise identical lockfile records.)

Should we include the exact, raw, contents of the DESCRIPTION fields in the lockfile (including newlines)? Or is it okay to normalize that whitespace? (In any case, it does not affect the hash.)

PR #2132, opened 2025-04-27
Update the `renv.lock` JSON schema

jrdnbradford:
This updates the `renv.lock` schema to fit with the changes made in #2059. 

The addition of these new fields would not have caused validation failures for `renv::lockfile_validate()` since the schema allows for other, non-defined fields. But now that they are defined in the schema we will get the extra validation (and if users add the schema to VS Code, Positron, etc, they will get the nice hover help for the new fields).

I also went through and clarified some of the `descriptions`, prettified the JSON, and updated the tests.


kevinushey:
Thanks!

PR #2145, opened 2025-06-23
s/RStudio Connect/Posit Connect/

nealrichardson:
Saw this vignette quoted in https://github.com/posit-dev/publisher/issues/2667 and figured it was time to update the language here. I did not search the rest of the repo, just did find/replace in this file.


kevinushey:
Thanks!

PR #2173, opened 2025-09-03
Update actions/checkout to v5

remlapmot:
I noticed the version mentioned in the ci.Rmd vignette is a bit out of date, I bumped it there and in the workflows.


kevinushey:
Many thanks!

PR #2183, opened 2025-10-16
Find extra dependencies for plumber2 projects

marcosnav:
Plumber projects may have a `_server.y(a)ml` file which contains the `engine` as a dependency that is not directly imported as `library(...)` but must be considered by `renv`.

### DIFF
--- a/R/dependencies.R
+++ b/R/dependencies.R
@@ -717,7 +717,7 @@ renv_dependencies_discover_plumber_server <- function(path) {
 
   # check if engine field exists and has a value
   engine <- contents$engine
-  if (is.null(engine) || !nzchar(engine))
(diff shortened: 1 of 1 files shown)


R/dependencies.R
```diff
@@ -703,6 +705,25 @@ renv_dependencies_discover_quarto <- function(path) {
   renv_dependencies_list_empty()
 }
 
+renv_dependencies_discover_plumber_server <- function(path) {
+  # require yaml package for parsing YAML
+  if (!renv_dependencies_require("yaml", "Server"))
```
kevinushey:
```suggestion
  if (!renv_dependencies_require("yaml", basename(path)))
```

Issue #2196, opened 2025-12-16
renv::lockfile not exported from namespace renv


alex-wenzel:
I was attempting to follow [this example](https://rstudio.github.io/renv/reference/lockfile-api.html), but `lockfile()` doesn't appear to be exported from `renv`. I checked `NAMESPACE` in the current `main` branch and it only exports functions like `lockfile_create()`, `lockfile_modify()`, etc. 

Unless I'm missing something, I believe either `lockfile()` is missing from `NAMESPACE` or that the documentation example is incorrect or outdated. I confirmed I am using `renv` version 1.1.5, which is the same version that appears in the header of the documentation example above.

kevinushey:
This function is marked as internal here:

https://github.com/rstudio/renv/blob/f53f84aedab2b33ce856df0a4e915d4564e04f90/R/lockfile-api.R#L113

so even though it's documented, it's not exported as I wasn't certain this was the API I wanted to expose for manipulating lockfiles in `renv`.

alex-wenzel:
Makes sense. In that case I suppose [this doc page](https://rstudio.github.io/renv/reference/lockfile-api.html) should probably be removed, or maybe have a note added that the function was switched to internal and link to [this one](https://rstudio.github.io/renv/reference/lockfiles.html) instead.

PR #2198, opened 2025-12-18
link lockfile-api to current lockfile API docs

alex-wenzel:
Addresses #2196. Not sure if `internal` or `deprecated` is the right word, or if there's a better one to use...


kevinushey:
Thanks!

PR #2216, opened 2026-02-10
per-package interactive recovery in restore()

### LINKED ISSUE #1893: Allow renv::restore to help restoring
After updating to R 4.4 I just spent 20 minutes repeating the following procedure to restore an old project:

* renv::restore()
* renv::update("packagethatcausedproblems")
* renv::record("packagethatcausedproblems")
* rince and repeat

Could we have `restore(and_try_updating_to_make_it_work=TRUE)` handle this itself? (With a better argument name obviously)

There is no need for a human to be at the helm to do the above routine repetatively for a period of time. Of course there are perfectly valid reasons why more or less blindly updating arbitrary packages isn't a good idea, but I imagine, for probably a large majority of projects, package versions that are no longer installable can be safely updated as the prefered solution instead of troubleshooting installation of old versions.

Halfway through I tried `renv::update()` to just update all packages, (and risking rebuilding packages not in cache), but this didnt fix things, still had to go back to update individual packages and repeat. This is also not an ideal solution as you now update everything and not just the ones that are causing a problem.

kevinushey:
## Summary

- When a package fails during an interactive `restore()` (`prompt = TRUE`), the user is now prompted with a menu offering three choices: install the latest version, skip the package, or cancel the restore.
- Adds `continue` parameter to `restore()` for non-interactive error accumulation (reports all failures at the end).
- Handler selection: `prompt = TRUE` → interactive recovery menu; `continue = TRUE` → accumulate errors; neither → immediate error (existing default).

Closes #1893

## Test plan

- [x] New test: "restore prompts for recovery on failure when prompt = TRUE" — verifies skip path via `renv.menu.choice = 2L`
- [x] New test: "restore recovery can install latest version" — verifies install-latest path via `renv.menu.choice = 1L`
- [x] New test: "restore recovery can cancel the restore" — verifies cancel path via `renv.menu.choice = 3L`
- [ ] Manual test: run `renv::restore()` interactively after upgrading R with a lockfile containing an old package version

### DIFF
--- a/R/restore.R
+++ b/R/restore.R
@@ -326,31 +326,39 @@ renv_restore_recover <- function(package, action, project, recovered) {
     cancel  = "Cancel the restore"
   )
 
-  choice <- menu(
-    choices = choices,
-    title   = sprintf("Failed to restore package '%s'. How would you like to proceed?", package),
-    default = 3L
-  )
+  repeat {
 
-  if (identical(choice, "cancel"))
-    stop(error)
+    choice <- menu(
+      choices = choices,
+      title   = sprintf("Failed to restore package '%s'. How would you like to proceed?", package),
+      default = 3L
+    )
 
-  if (identical(choice, "skip"))
-    return()
+    remote <- NULL
+    if (identical(choice, "latest")) {
+      remote <- package
+    } else if (identical(choice, "version")) {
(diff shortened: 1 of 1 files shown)


kevinushey:
This PR won't be compatible with the new parallel installation code; we'll need to rework this.

Issue #1735, opened 2023-10-19
Renv installation fails with dependency only available in Archive


bentilley:
Hello! Thanks for the great package 🎉 

## Description of Issue

I'm looking for some guidance on how `renv` handles dependencies stored in the archive of a repo. Specifically we are using Artifactory, and have a package `ABC` that isn't available at `/src/contrib` on the repo, it's only available at `/src/contrib/Archive`. This file doesn't appear in the `PACKAGES` index for the repo.

From checking through the `renv` source, it seems that `renv` uses this `PACKAGES` index exclusively to determine what is available on the repo. This means that `renv::install("ABC", repos="artifactory.example.com/repo")` fails with `Error: package 'ABC' is not available`.

If I do `renv::install("ABC@0.1.2", repos="artifactory.example.com/repo")` (i.e. requesting the specific version that is in the archive. Then this succeeds. From checking the source, this seems to be because specific version installations elicit an additional `renv` heuristic check in the `/Archive` location.

So that's good, that we can explicitly fetch the package. However, what I don't know how to handle is when I try to install a package `MainPackage` that depends on the archived package `ABC`. In this situation, when I run `renv::install("MainPackage", repos="artifactory.example.com/repo")`, `renv` goes to fetch the package dependencies and in the end errors with `Error: package 'ABC' is not available`.

It looks to me that during the dependency fetching, `renv` is only consulting the `PACKAGES` index, and therefore doesn't know that the `ABC` package is available in the archive.

## Questions

1. Is this a bug with our repo? i.e. Should the archived package appear in the `PACKAGES` index somehow? (I should note that the same package is available on public CRAN, and on CRAN it also doesn't appear in the `PACKAGES` index, which makes me think that this is not the case. However, I'm relatively new to the R eco system, so def wouldn't put it past myself simply having completely the wrong end of the stick about how these things _should_ be working!)
2. Should `renv` be looking further afield to determine the available packages on a repo? I've noticed while exploring this issue that cranlike repos provide an `archive.rds` file available at `/src/contrib/Meta/archive.rds`, should `renv` also consult this and add the data to its list of index DB's when determining what packages are available?
3. If the answer to 2 is "no", then should `renv` also be performing its own archive heuristic check (i.e. looking in `/Archive`) for packages that don't appear in it's index DB's?

Thank you for any insight you can shed on this 🙏 . `renv` is great!

kevinushey:
[...]

PR #2220, opened 2026-02-11
Select correct source when crandb and CRAN return same version.

aramirezreyes:
This works relates to #1735

Disclaimer: I got help from cursor AI to help me understand the flow of some of these functions and critique my solutions.

I think I have a solution for the problems at the end of that discussion, but I may need some handholding and reassurance I am not breaking other cases. Could someone take a look at this?

**Problem 1**: when trying to install a package, if `options(renv.config.crandb.enabled = TRUE)`,  the version returned by crandb is the live version on CRAN. Renv will try to build a URL to get the file from the archive. But the archive does not have that file as it is the live version.

```r
Warning: error downloading 'https://cran.r-project.org/src/contrib/Archive/RColorBrewer/RColorBrewer_1.1.3.tar.gz' [error code 22]
```

**Solution 1**: when both repo and crandb return the same version, select the one with url (repo)

**Problem 2**: The version is converted to its numeric value, so when trying to build a URL, the version is incorrect: (1.1-3 is converted to 1.1.3, leading to an incorrect url).

**Solution 2**: leave the value untouched for building the url.

I added a test, which seems to work but I am getting a few failed tests locally. I probably do not the correct setup to run the full test suite.

### DIFF
--- a/tests/testthat/test-available-packages.R
+++ b/tests/testthat/test-available-packages.R
@@ -198,6 +198,7 @@ test_that("available_packages() tolerates missing repositories", {
 test_that("crandb query returns R-compatible versions", {
   skip_on_cran()
   skip_if_offline()
+  skip_if_offline(host = "r-pkg.org")
   renv_scope_options(renv.config.crandb.enabled = TRUE)
 
   # Query crandb for a package with known R version requirements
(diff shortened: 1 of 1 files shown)


kevinushey:
Looks like tests are failing -- are you able to investigate?

aramirezreyes:
I tried in two different R versions and I cannot reproduce the failed tests. Would you mind re-triggering the tests? I wonder if we were unlucky and crandb was unreachable. I also added a `skip_if_offline(host = r-pkg.org)`.

kevinushey:
Looks like CI is happy now. Thanks!

PR #2234, opened 2026-03-04
fix(): ensure lib path is created

blyedev:
Currently the renv bootstrapping from cache on a fresh machine with the cache dir restored fails with a warning that the dir does not exist.


kevinushey:
LGTM -- thanks!

blyedev:
Thanks for the lightning fast review! Quick question, whats the release cycle? I'd love to upgrade our internal systems to a tagged release. No rush ofc, just want to make a mental note on when to check in on this.

PR #2246, opened 2026-03-24
Rename and export `lockfile_from_manifest`

### LINKED ISSUE #2245: Export `renv_lockfile_from_manifest()`?
I found that I needed to create an `renv.lock` from a manifest file, and was surprised that `renv_lockfile_from_manifest()` wasn't yet exported. It looks like it's never been (though it is used in some auto-init code, the tools I'm using don't quite follow that path, so I need to call it manually.

jonkeane:
Resolves #2245


kevinushey:
Thanks, but I intentionally don't want to export this, because it's more-or-less a utility function that is only relevant for rsconnect. (correct me if I'm wrong)

Does this still need to live in renv? Would it be possible to move / reimplement this in rsconnect?

jonkeane:
FWIW, I was using it to publish with the Publisher something I already had a manifest for, and just needed to make sure I used the same packages. So no rsconnect involved. 

In the publisher, we will want to use something like this to shortcut there (the current path that published uses with renv didn't use this, hence me needing to do it manually).

Happy to close this if you really don't want to export this, but in my experience being able to make a lockfile from a manifest helps users get into more modern lockfile experiences more easily. If we don't export this, we might need to add it to other pathways to ensure that publisher does hit it.

kevinushey:
How hard would it be to move this code into publisher? Would it be made easier if we exposed some slightly more generic tools from `renv`?

jonkeane:
I do think renv is the right home for this honestly. There is value for someone who has a manifest from rsconnect and wants to start using renv but isn't using publisher.

But if you're concerned about exposing this to renv, you we could add it to publisher. The r calling code is a bit in flux right now, but I'm happy to help point out where you should PR this to.

jrosell:
I also think these functions should be exported in the renv package because one can generate a lockfile locally from any exported project from posit connect.

kevinushey:
Decided to put this into a more generic `lockfile(from = <...>, to = <...>)` interface.

Issue #1936, opened 2024-06-28
xml2 not being installed


sjenkins20:
We're using testthat and in particular the test reporter. The reporter requires the xml2 package.
https://github.com/r-lib/testthat/blob/HEAD/R/reporter-junit.R

The testthat package checks to see if xml2 is installed. 
https://github.com/r-lib/testthat/blob/HEAD/R/reporter-junit.R

This dependency isn't being detected by renv and the package is not being installed when inferring dependencies. We have tried running the command `renv::install('xml2')` both with and without dependency inferrence and xml2 does not appear in the final renv.lock file, so we have had to manually edit renv.lock to add it which is not best for performing easy package updates.

I have a simple test case, requiring the packages testthat and Matrix to demonstrate that dependency inference is occurring. Run requirements.R to see the issue. We've tried both in the state below and with `renv::init(bare=TRUE)` and xml2 doesn't remain in the lock file.

script.R
`library('Matrix')

result <- function(text) {

	print(text)

	return(TRUE)
}`

testthat.R
`library('testthat')

working_dir <- getwd()`

requirements.R
`renv::init()

# Install xml2
renv::install('xml2', repos='https://cran.ma.imperial.ac.uk/')`

During execution of renv::install for xml2, we get the following console message
`The following package(s) will be installed:
- xml2 [1.3.6]
These packages will be installed into "./renv/library/windows/R-4.4/x86_64-w64-mingw32".

# Installing packages --------------------------------------------------------
- Installing xml2 ...                           OK [linked from cache]
Successfully installed 1 package in 38 milliseconds.`

sjenkins20:
On our codebase where we noticed the problem, we initially run `renv::init()` and then run `renv::install('xml2')`. The following message is produced
```
Loading required package: xml2
[1] "INFO: Installing xml2"
[1] "Using repo"
The following package(s) will be installed:
- xml2 [1.3.6]
These packages will be installed into "./renv/library/windows/R-4.4/x86_64-w64-mingw32".

# Installing packages --------------------------------------------------------
- Installing xml2 ...                           OK [linked from cache]
Successfully installed 1 package in 46 milliseconds.
[1] "SUCCESS: xml2 successfully installed"
```
At the end of the process, the message says
```
Warning message:
In library(package, lib.loc = lib.loc, character.only = TRUE, logical.return = TRUE,  :
  there is no package called 'xml2'
```

kevinushey:
What version of `renv` are you using? Can you also run installation with:

[...]

PR #2333, opened 2026-07-13
detect rlang::check_installed() and testthat optional dependencies

kevinushey:
Addresses parts of #1936.

## What this does

- Adds a new dependency-discovery handler for `rlang::check_installed()` and
  `rlang::is_installed()`, mirroring the existing `requireNamespace()` handler.
  It handles bare and namespaced calls, named or positional `pkg` arguments,
  vectors like `c("shiny", "diffviewer")`, and version-qualified entries like
  `"openxlsx (>= 4.0.0)"` (the version qualifier is stripped).

- Extends the testthat handler to infer optional dependencies implied by
  certain function usages:
  - `expect_known_hash()` -> digest
  - `expect_s7_class()` -> S7
  - `expect_vector()` -> vctrs
  - `run_cpp_tests()` -> xml2
  - `skip_if_offline()` -> curl
  - `snapshot_review()` -> shiny, diffviewer

- Junit reporter detection (xml2) now also matches `test_check()` and
  `test_local()`, in addition to `test_package()` / `test_dir()` /
  `test_file()`. `test_check()` is the usual entry point in
  `tests/testthat.R`, which is likely what the original report was running.

`skip_if_not_installed()` is deliberately not detected, since its purpose is
to declare a package optional.

## What this doesn't do

Static analysis still can't see `check_installed()` calls made inside
third-party package code (e.g. testthat's own `JunitReporter` check, or the
`openxlsx` example in the issue comments); for those cases the
`_dependencies.R` escape hatch or `snapshot(type = "all")` remains the
recommendation.

Note: the regenerated `man/dependencies.Rd` also picks up a paragraph about
`renv.dependencies.implied` that existed in the roxygen source but had not
yet been committed to the Rd.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,15 @@
 
 # renv (development version)
 
+* `renv::dependencies()` now detects packages referenced via
+  `rlang::check_installed()` and `rlang::is_installed()`. (#1936)
+
+* `renv::dependencies()` now infers the optional packages required by certain
+  testthat functions; for example, `skip_if_offline()` implies a dependency on
+  curl, and `snapshot_review()` implies dependencies on shiny and diffviewer.
+  In addition, usage of the Junit reporter is now also detected in calls to
+  `test_check()` and `test_local()`. (#1936)
+
 * `renv::install()` with pak enabled no longer upgrades already-installed
   dependencies that are only pulled in transitively -- most visibly recommended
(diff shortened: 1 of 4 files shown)


DanChaltiel:
Hi Kevin,
Sorry, I thought it was on topic, but I should have made a separate issue.
I installed from GitHub (1.2.3.9000) and gave it a try, and it works even better than expected.
Thank you for taking the time!

PR #2334, opened 2026-07-13
fix length-1 check in requireNamespace dependency discovery

sims1253:
Found during my big https://github.com/sims1253/ry audit


kevinushey:
Thanks!

PR #2343, opened 2026-08-03
report the underlying cause when package downloads fail

### LINKED ISSUE #2340: `renv::install` for a GitHub package fails to download
```
> renv::install("rspatial/terra@b94822ca32fcad64117026865dd93f10d4e5e6bf", verbose = T)
The following package(s) will be installed:
- terra [rspatial/terra@b94822ca]
These packages will be installed into "~/XIS/renv/library/linux-ubuntu-resolute/R-4.6/x86_64-pc-linux-gnu".

Do you want to proceed? [Y/n]:
# Downloading packages -------------------------------------------------------
✖ terra 1.9-41

The following package(s) were not installed successfully:
- [terra]: failed to download
You may need to manually download and install these packages.
```

I can't figure out how to get a detailed error message. I have a `GITHUB_PAT` set. I'm on `renv` 1.2.3, R 4.6.0, and Ubuntu 26.04.

kevinushey:
Closes #2340.

## Problem

When a package fails to download, the graph installer reports only a generic message:

```
The following package(s) were not installed successfully:
- [terra]: failed to download
```

The sequential download fallback catches the real error -- for example, renv's own "destination directory is not writable" check -- and discards its message, leaving the user with no clue about the cause (see #2340, where the culprit was a shared renv root with misconfigured permissions).

## Approach

- Capture the condition message from failed fallback retrievals, and surface it in the install error summary:

  ```
  The following package(s) were not installed successfully:
  - [bread]: destination directory "/path/to/renv/source/repository/bread" is not writable; cannot proceed
  ```

- Refine the unwritable-destination error in `renv_download_impl()` to name the containing directory rather than the (not-yet-existing) temporary download file, and avoid double-quoting the path.

Existing snapshots for a missing package now report `package 'missing' is not available` instead of the generic message.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,12 @@
 
 # renv (development version)
 
+* When a package fails to download, renv now reports the reason for the
+  failure, rather than a generic "failed to download" message. For example,
+  when the download destination within the renv root is not writable (as can
+  happen with a misconfigured shared cache), the error now says so directly.
+  (#2340)
+
 * `renv::use()` with pak enabled now honours the requested remotes, rather than
   installing the latest version of each package from the active repositories.
   Previously, a call like `renv::use("generics@0.1.3")` would install the
--- a/R/download.R
+++ b/R/download.R
@@ -132,10 +132,11 @@ renv_download_impl <- function(url, destfile, type = NULL, request = "GET", head
   url      <- chartr("\\", "/", url)
   destfile <- chartr("\\", "/", destfile)
 
-  # check that the destination file is writable
+  # check that the destination file is writable; report the containing
+  # directory, since the destination file itself need not exist yet
   if (!renv_file_writable(destfile)) {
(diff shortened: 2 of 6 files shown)


Kodiologist:
There may be a [time-of-check to time-of-use](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue here: the destination can cease to be writable after it's been checked but before the write starts.

kevinushey:
True -- the check is inherently point-in-time. It's intended as a best-effort diagnostic for the common case of persistently misconfigured permissions (as in #2340), so that the failure produces an actionable message. If the directory becomes unwritable after the check, the download itself still fails, and that failure is caught and reported through the normal error paths -- so the race can only make the error message less specific; it shouldn't affect correctness.

That said, if you have a scenario where this leads to concrete misbehavior, please feel free to file a new issue with the details.

Issue #1901, opened 2024-05-13
`ppm.enabled` in settings vs. config


bzkrouse:
The documentation implies that the config and settings `ppm.enabled` options operate similarly. However a recent use case brought up some questions about the difference between these two options. Here is the scenario:

- Repo is set to a specific, older CRAN snapshot
- Package version from the specified CRAN snapshot is *earlier* than the default package version from P3M (according to the version of R being used in the session)
- Windows OS
	
For example, from R 4.3.2, the following code:
```
renv::install("dplyr", repos = "https://packagemanager.posit.co/cran/2023-03-01")
```

Results in 
```
# Downloading packages -------------------------------------------------------
- Downloading dplyr from P3M ...                OK [1.5 Mb in 2.1s]
Successfully downloaded 1 package in 2.3 seconds.
The following package(s) will be installed:
- dplyr [1.1.4]
```

As opposed to downloading dplyr 1.1.0 from the 2023-03-01 snapshot. 
I have found that if I set config ppm.enabled = FALSE and settings ppm.enabled = TRUE, I *think* I am able to avoid {renv} from reaching out to P3M, while keeping other benefits of PPM being enabled. 

Is this the expected behavior/the recommended configuration for sticking to just the specified repos? If so, is there a way to update the documentation so these 2 options can be more easily distinguished?

Thanks very much in advance for your help!

bzkrouse:
@kevinushey thanks for your reply. I should mention I only have this issue on Windows. Here are some diagnostics:

```
 Diagnostics Report [renv 1.0.7]
===============================

# Session Info ---------------------------------------------------------------
R version 4.3.2 (2023-10-31 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19045)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.utf8 
[2] LC_CTYPE=English_United States.utf8   
[3] LC_MONETARY=English_United States.utf8
[4] LC_NUMERIC=C                          
[5] LC_TIME=English_United States.utf8    

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     
loaded via a namespace (and not attached):
[1] compiler_4.3.2 tools_4.3.2    renv_1.0.7    
# Project --------------------------------------------------------------------
Project path: "~/renv_ppm"

# Status ---------------------------------------------------------------------
This project does not contain a lockfile.
[...]

PR #2356, opened 2026-08-20
Fix available package fallback ordering

LeonidasZhak:
## Summary

Restore P3M as a fallback when configured repositories have no candidate, while keeping configured repository records preferred.

## Thanks to maintainers

Thanks for the recent available-package resolver work in #2348 and for the diagnosis documented in #1901.

## Issue or motivation

When a package is unavailable from the configured repositories, the available-package lookup should continue to later fallback sources. This is relevant to the repository/snapshot behavior described in #1901.

## Root cause

The resolver builds an ordered list of repository, crandb, archive, and P3M candidates, but the final return expression only considered the first two entries after crandb was added in #2348. A P3M candidate could therefore be found and then discarded.

## Change

Return the first non-null candidate in the existing method order and add regression tests for both P3M fallback and configured-repository precedence.

## Tests

- Focused mocked regression tests: 2 tests, 2 successes
- `R CMD build .`: passed
- `R CMD check --as-cran renv_1.2.4.9000.tar.gz`: passed with 3 existing NOTES (development version, groundhogr.com HTTP 406, and local HTML Tidy version)

## Scope

This changes candidate selection only. It does not change repository querying, P3M database contents, version compatibility rules, or package installation behavior.

### DIFF
--- a/R/available-packages.R
+++ b/R/available-packages.R
@@ -422,28 +422,18 @@ renv_available_packages_latest <- function(package,
 
   })
 
-  # if both entries are null, error
-  if (all(map_lgl(entries, is.null))) {
-    map(errors$data(), warning)
-    entry <- the$rejected_packages[[package]]
-    if (!is.null(entry))
-      stopf("package '%s' is not available\n- %s", package, entry$reason)
-    stopf("package '%s' is not available", package)
-  }
+  # prefer configured repositories, using crandb, archives, and P3M as
+  # fallbacks when earlier methods have no candidate.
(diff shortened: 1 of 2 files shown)


kevinushey:
Thanks -- I made an alternate fix in https://github.com/rstudio/renv/pull/2357. Can you let me know if this is sufficient?

LeonidasZhak:
Yes, #2357 is sufficient from my perspective. Thanks for the alternate fix and for preserving the intended repository precedence while avoiding a premature re-enablement of P3M. The archive handling and the additional tests look good as well.

PR #2366, opened 2026-09-15
safely handle invalid git tarball when extracting sha1 in bootstrap

jkylearmstrong:
## Problem

When `renv_bootstrap_git_extract_sha1_tar()` attempts to inspect a Git archive bundle to extract the commit SHA-1 from the pax header, an invalid, corrupt, or non-existent file causes an uncaught error (e.g. from `file()`, `gzcon()`, `readBin()`, or `rawToChar()`). This can cause the bootstrap process to abort prematurely instead of falling back to alternative resolution methods.

## Solution

- Wrap `renv_bootstrap_git_extract_sha1_tar()` in a `tryCatch(..., error = function(e) NULL)` so invalid or unreadable archives fail closed and return `NULL`.
- Keep `R/bootstrap.R` and `inst/resources/activate.R` synchronized.
- Add a unit test in `tests/testthat/test-bootstrap.R` verifying that non-existent files and invalid binary payloads fail closed and return `NULL`.

## Verification

Ran `tests/testthat/test-bootstrap.R`:
- Unit test `"git tarball sha extraction fails closed on invalid input"` passes with 2/2 successes.


jkylearmstrong:
@copilot

kevinushey:
Thanks!

PR #2367, opened 2026-09-16
`renv_renvignore_exists`

jkylearmstrong:
added `renv_renvignore_exists` to address all instances of 

# Why the User Experience Was Problematic
When users configure a `.renvignore` file, they expect the tool to be actively honoring it. When renv encountered a large file count or a slow scan and printed: 
```
Consider using .renvignore to ignore irrelevant files.
```
it created immediate doubt:

"Did `renv` fail to read my .renvignore?"
"Is the syntax in my file broken?"
"Is `renv `ignoring my ignore rules entirely?"
By changing the warning when `.renvignore` is detected to:

```
A large number of files (%i in total) have been discovered using current .renvignore specifications.
It may take renv a long time to scan these files for dependencies.
Consider modifying .renvignore to ignore irrelevant files.
```
the message directly reassures the user:

Confirmation: `renv` acknowledges that `.renvignore` was detected and applied.

Context: The reported count reflects the files remaining after applying those rules.
Actionable guidance: It tells them to modify their existing file to exclude further directories/files, rather than prompting them to create one.

Architectural Benefits of `renv_renvignore_exists()` 

Centralizing this into `renv_renvignore_exists()` and `renv_renvignore_path()` in `R/renvignore.R` makes the behavior consistent and extensible across the codebase:

Consistent Project Resolution: By using:
```
project <- project %||%
  renv_dependencies_state(key = "root") %||%
  renv_project_resolve()
```

both `R/dependencies.R` (which calls renv_renvignore_exists() without arguments) and `R/snapshot.R` (which passes project) resolve the exact same project root path cleanly.

Unified Messaging Tone: Both `dependencies()` and `snapshot()` now use the identical condition and matching language ("using current `.renvignore` specifications" and "Consider modifying `.renvignore`").
Single Point of Maintenance: If `renv` later expands how it checks for ignore configurations (such as checking `.gitignore` or detecting ignore files in subdirectories), you only need to update the logic inside `renv_renvignore_exists()` and all call sites in `dependencies.R`, `snapshot.R`, or any future diagnostic functions will immediately inherit the updated detection behavior.


kevinushey:
Thanks, much appreciated!

